PR #5, opened 2020-08-08
Qw123wh description play store italian

qw123wh:
check if there are any problems. Maybe the accented letters give problems I don't know

### DIFF
--- a/fastlane/metadata/android/en-US/full_description.txt
+++ b/fastlane/metadata/android/en-US/full_description.txt
@@ -4,7 +4,7 @@ Note: This project is currently in alpha stage. If you want to see a feature bei
 
 Currently implemented and fully working features:
 * Latin keyboard layouts
- * QWERTY, QWERTZ, AZERTY, Spanish and Swiss. More coming in future releases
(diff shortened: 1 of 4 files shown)


qw123wh:
the second file is the important one (add file via upload) . The former is useless

patrickgold:
Thank you for your submission. The accented letters indeed are only a question mark (= invalid character code). Did you use UTF-8 as the file encoding in your text editor? If that doesn't work you can also use HTML entities (see [https://www.freeformatter.com/html-entities.html](https://www.freeformatter.com/html-entities.html) section ISO-8859-1 Characters).
For instance: `abcdéf` would be written as `abcd&eacute;f`. The PlayStore then correctly renders the letter with its accent.

qw123wh:
I changed the accented letters see if that's okay.  However, it would be much easier for me to send you the translated files via email.

patrickgold:
Thank you for your submission, this time everything is ok.

Regarding sending the translation files via E-Mail: I intend to keep all changes on GitHub and manage them via pull requests, but I can do an exception for you in the future, if it is easier for you this way.

qw123wh:
OK thanks.  We will talk about this later when there is more to translate

PR #6, opened 2020-08-08
Add support for persian keyboard layout

PHELAT:
First of all, thank you for developing this awesome project. This might be the one open-source keyboard that actually works fine. In this pull-request I'm adding the support for Persian layout and this also comes with the ability to have half-space key in any layout that needs one. One important infrastructure change in this pull-request is the ability to load a custom layout modifier for each keyboard layout.

### DIFF
--- a/app/src/main/assets/ime/config.json
+++ b/app/src/main/assets/ime/config.json
@@ -11,7 +11,8 @@
     "icelandic": "Icelandic (QWERTY)",
     "swiss_german": "Swiss German (QWERTZ)",
     "swiss_french": "Swiss French (QWERTZ)",
-    "swiss_italian": "Swiss Italian (QWERTZ)"
+    "swiss_italian": "Swiss Italian (QWERTZ)",
+    "persian": "Persian"
   },
   "defaultSubtypes": [
     {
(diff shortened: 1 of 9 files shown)


patrickgold:
Thank you for your great addition to FlorisBoard, I really appreciate it. I've looked at the changes in the browser and everything checks out for now. I will later this evening or tomorrow morning check the changes locally and check them and if everything works I will happily merge your changes.

patrickgold:
I have checked the changes locally and everything but one tiny little thing checks out. Is there a specific reason why you included the delete key in the character persian layout instead of the modifier layout?

patrickgold:
Ah after trying around a bit I see that my LayoutManager implementation isn't the smartest and adds the first and last modifier character regardless of the key code. Then you have an empty key on the screen, which you tried to avoid by moving the delete key to the character layout. Good thinking on your side! Would it be possible that you still move the delete key back to the modifier layout and I will correct the merging behavior of LayoutManager?

The first row of the persian modifier layout should look like this: (a key with code 0 is a placeholder for the last row of the character layout)
```
    [
      { "code":    0 },
      { "code":   -5, "label": "delete", "type": "enter_editing" }
    ]
```

PHELAT:
@patrickgold Sure, it's now fixed in 235224aefd5f3a56268e79457fbf59775623319e.

patrickgold:
Perfect, thank you! I've merged your changes into the master branch and will soon commit a fix for the LayoutManager behavior.

patrickgold:
LayoutManager behavior has been fixed in 236f6826226fe5481bc533d18b298837bc8c9b15.

PR #8, opened 2020-08-10
Add Disable Checkbox for "Preview Popup" feature

hamedsj:
First of all, thank you for this clean and beautiful code, In this PR I added just some keys that are missing in the Persian keyboard layout that Added in the last PR, And besides that, I added a disable checkbox for a "preview popup" feature.


patrickgold:
Thank you for your contribution, I really appreciate it! I've reviewed your changes locally and everything checks out.

Note that I currently try some different Settings structures locally as the current implementation with the BottomNavigation is not very flexible in terms of adding whole new pref groups. This means I might move around your newly added preference in the Settings UI, just so you know.

Issue #9, opened 2020-08-10
Show Dashboard on the top of keyboard when Suggestion in disabled


hamedsj:
It will be awsome if We can add a dashboard including "select all", "copy", "cut", "paste", "left arrow", "right arrow" on the top of the keyboard while Suggestion feature is disabled.

like [this Keyboard](https://i.ibb.co/5j7RGbD/photo5877334691135206355.jpg)

patrickgold:
Your proposed feature is in general a mix of the planned clipboard support and the also planned cursor movement support. I will definitely look into it when I'm implementing these two features. Thank you for the idea!

patrickgold:
I've implemented cursor/clipboard related tools in #13, if you have any feedback to the newly added feature, please post it either here or in the PR. Thank you!

rancidfrog:
They clipboard tools are great.
But is there a way to have both suggestions and tools?
Or a quick way to switch from one to another, like  right arrow since there is a left arrow which takes you to the tab with emojis settings...
As both are used often

patrickgold:
@rancidfrog I see suggestions like this quite a lot (but more commonly with the number row). I am unsure how to handle it yet and what could be useful, what not. The Smartbar will get a huge overhaul in the course of the next week (due to experimenting with adding proper suggestions). I will see what I'll come up with then.

Issue #3, opened 2020-08-06
Show row of numbers above the keyboard


qw123wh:
look at this keyboard there are interesting things you could add such as the number row and other features

https://github.com/rkkr/simple-keyboard

patrickgold:
I will definitely look into adding this in the near feature, thanks for the request!

patrickgold:
An option to show a number row above the keyboard has been added with #13, if you have any feedback to the newly added features please write it either here or in the PR. Thank you!

PR #13, opened 2020-08-16
Add clipboard cursor related tools

patrickgold:
This PR adds a clipboard/cursor keyboard layout (accessible via Smartbar quick actions) and the ability to show a number row or clipboard cursor toolbar above the keyboard if candidate suggestions are disabled.

Editing keyboard layout:
image

Smartbar clipboard/cursor toolbar (#9):
image

Number row (#3):
image

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -34,6 +34,7 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation 'androidx.core:core-ktx:1.3.0'
     implementation 'androidx.preference:preference:1.1.1'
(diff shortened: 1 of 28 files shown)


bertin0:
Could you make the number row not a part of the smartbar? I find the smartbar useful and would like to use both.

patrickgold:
If I understood your proposal correctly you want to display both the clipboard cursor row _and_ the number row above the character layout? That is indeed not possible at the moment. As a fallback you can use the integrated number row in character layout (added in v0.2.2) and set Smartbar to display the clipboard cursor row.

PR #66, opened 2020-12-03
Added an Arabic keyboard and mod, and changed persian ID to 801

HeiWiper:
I've added an Arabic keyboard based on AOSP Arabic layout by following PR #6 
There are two other things I wanted to add but didn't figure out how to or if it is allowed :

1. Arabic keyboards usually come with Arabic-Indic digits which I have added to the top row as a popup but I would've preferred if they were lower in priority than the default digits like AOSP keyboard, here's how it is right now in both Persian and Arabic keyboards:
Screenshot_20201204-001818_TinyBit_launcher (1)

2. I wanted to edit the digits and symbols layout to change the digits as well as some symbols to the Arabic ones just like here in the screenshot below taken from AOSP keyboard.
Screenshot_20201204-000519_TinyBit_launcher (1)

### DIFF
--- a/app/src/main/assets/ime/config.json
+++ b/app/src/main/assets/ime/config.json
@@ -12,7 +12,8 @@
     "swiss_german": "Swiss German (QWERTZ)",
     "swiss_french": "Swiss French (QWERTZ)",
     "swiss_italian": "Swiss Italian (QWERTZ)",
-    "persian": "Persian"
+    "persian": "Persian",
+    "arabic": "Arabic"
   },
   "defaultSubtypes": [
     {
@@ -177,9 +178,16 @@
       "isEmojiCapable": true
     },
     {
-      "id": 800,
+      "id": 801,
       "languageTag": "fa-FA",
       "preferredLayout": "persian",
+      "isAsciiCapable": true,
+      "isEmojiCapable": true
+    },
+    {
+      "id": 901,
+      "languageTag": "ar",
+      "preferredLayout": "arabic",
       "isAsciiCapable": true,
       "isEmojiCapable": true
     }
--- a/app/src/main/assets/ime/text/characters/arabic.json
+++ b/app/src/main/assets/ime/text/characters/arabic.json
@@ -1,0 +1,46 @@
+{
+  "type": "characters",
+  "name": "arabic",
+  "direction": "rtl",
+  "modifier": "arabic",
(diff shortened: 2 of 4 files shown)


patrickgold:
Thanks for your great PR, I really appreciate it!

On a first glance while reviewing your changes on my phone, everything looks good. I will test your changes tomorrow in Android Studio and if everything checks out (which I'm sure it will) then I'll merge your changes. 

Best wishes
Patrick

patrickgold:
I've reviewed your PR and it looks pretty good!

Minor things:
- In the top-most row, you've hard coded the Arabic-Indic digits. I will make them a separate unit (with popups for the normal Arabic digits), so it can also be used in the Smartbar row and in the symbol layout too.
- A question to your second point: Are these symbols valid for both Persian and Arabic or only Arabic?

Another question unrelated to this PR:
Would it make sense to also make an Arabic/Persian number layout (for input fields which accepts only numbers)?
This is what it looks like when set to Arabic/Persian:
image

I will merge your PR now and based on this resolve your two bug reports and the hard-coded number row.

Thanks again for your great efforts, I really appreciate it!

HeiWiper:
First, I think a separate unit would be better but there might be a need to make another subtype for users who don't use these digits.
For the second point, I have noticed that these digits aren't the same as Persian ones, some of them look similar and some don't but they all have different codes.

Concerning the last question I don't think it would be necessary to make  another digits keyboard as the digits aren't compatible with number fields and even AOSP keyboard doesn't have such keyboard for Arabic/Persian.

I'm really glad to help and I will do try to do the French translation very soon (maybe before 0.3 milestone aka tonight or tomorrow)

patrickgold:
Thanks for your response! For now, I think I'll leave the numbers as they are and think about them at a later stage.

Good to know that other keyboards also use the normal digits in the numeric keyboard, I'll leave it as it is then.

I'm really grateful for your help. I've also seen that you've already translated FlorisBoard into French, thank you! :)

KingLucius:
Arabic speaker here and I can help with this commit, Swift KB uses Two Arabic layouts one with English Digits and another one with Arabic digits.

Arabic-Digits:
Arabic-Digits

Arabic-Digits1

English-Digits:
English-Digits

English-Digits1

using Arabic or English digits in a personal preference.

PR #75, opened 2020-12-08
Add Hungarian layout

zoli111:
I added Hungarian layout to this great keyboard.

### DIFF
--- a/app/src/main/assets/ime/config.json
+++ b/app/src/main/assets/ime/config.json
@@ -12,6 +12,7 @@
     "swiss_german": "Swiss German (QWERTZ)",
     "swiss_french": "Swiss French (QWERTZ)",
     "swiss_italian": "Swiss Italian (QWERTZ)",
+    "hungarian": "Hungarian (QWERTZ)",
     "persian": "Persian",
     "arabic": "Arabic"
   },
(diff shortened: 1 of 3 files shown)


patrickgold:
Thanks for your first-time contribution, zoli111!

I've reviewed your proposed keyboard addition and found an issue:
- The keyboard layout itself you are adding is redundant, as one could just use QWERTZ and get the same result. Your layout should add some special characters specific for your language (e.g. é, á, ...) as separate keys. Gboard has a pretty good example of how one could add separate special character keys for Hungarian. This layout could then be named 'Hungarian (QWERTZ)', as it is a layout specific for this language based on QWERTZ.

The extended popups are fine as they are but would need extra popups for the proposed special keys (as some in Gboard seem to have some). The config.json looks good. Note that regardless of the currently selected layout, FlorisBoard does automatically pull the extended popups for the current subtype local, so one could even get Hungarian extended popups while using AZERTY for instance.

If you incorporate these changes, I'll happily merge your PR. If you have any questions or want to discuss these proposed improvements, don't hesitate to ask me!

patrickgold:
For reference, this is what I meant with how Gboard has designed the Hungarian (QWERTZ) keyboard layout:

image

zoli111:
I added the ö é á and ü keys so it can be merged now.

By the way I haven't seen this layout before because the normal QWERTZ is the default at Gboard. Good to know there's like this.

patrickgold:
Thanks for editing the layout! I've reviewed your changes and will now merge these into FlorisBoard!

Issue #89, opened 2020-12-19
Help needed with adding esperanto keyboard layout


jeremiah-miller:
Hello, I'm learning esperanto and I'd like to be able to use florisboard instead of the aosp keyboard or [anysoftkeyboard](https://play.google.com/store/apps/details?id=com.anysoftkeyboard.languagepack.esperanto&hl=en_US&gl=US). I know you want to hold off on adding layouts till it's more stable and it's totally ok if you decide to stay firm on that, but if you are ok with adding it early I did my best to make a layout json file however gradle wasn't letting me install the debug app on my phone (`> java.util.concurrent.ExecutionException: com.android.builder.testing.api.DeviceException: com.android.ddmlib.InstallException: INSTALL_FAILED_UPDATE_INCOMPATIBLE: Package dev.patrickgold.florisboard signatures do not match previously installed version; ignoring!`) so I don't know if I did it right. I based the code values on the furthest right number in the table at https://en.wikipedia.org/wiki/Esperanto_orthography#Unicode

```
{
  "type": "characters",
  "name": "Esperanto",
  "direction": "ltr",
  "arrangement": [
    [
      { "code":  349, "label": "ŝ" },
      { "code":  365, "label": "ŭ" },
      { "code":  101, "label": "e" },
      { "code":  114, "label": "r" },
      { "code":  116, "label": "t" },
      { "code":  285, "label": "ĝ" },
      { "code":  117, "label": "u" },
      { "code":  105, "label": "i" },
      { "code":  111, "label": "o" },
      { "code":  112, "label": "p" }
    ], [
    ], [
      { "code":   97, "label": "a" },
      { "code":  115, "label": "s" },
      { "code":  100, "label": "d" },
      { "code":  102, "label": "f" },
      { "code":  103, "label": "g" },
      { "code":  104, "label": "h" },
      { "code":  106, "label": "j" },
      { "code":  107, "label": "k" },
      { "code":  108, "label": "l" }
      { "code":  309, "label": "ĵ" },
    ], [
      { "code":  122, "label": "z" },
      { "code":  265, "label": "ĉ" },
      { "code":   99, "label": "c" },
      { "code":  118, "label": "v" },
      { "code":   98, "label": "b" },
      { "code":  110, "label": "n" },
      { "code":  109, "label": "m" }
      { "code":  293, "label": "ĥ" }
    ]
  ]
}
```

patrickgold:
[...]

PR #90, opened 2020-12-19
Added Esperanto keyboard layout

jeremiah-miller:
Added Esperanto keyboard layout as discussed in #89


patrickgold:
Thanks for the PR, I really appreciate it!

Two things before I can merge your PR:
1. In your PR, you didn't add a file for the extended popups for the Esperanto language. Is this intended? If FlorisBoard does not find an eo.json file in the extended_popups directory, then there won't be any extended popups available if the user selects the Esperanto language.
2. I just noticed that you have an empty key row in esperanto.json (line 17 + 18).

If these two issues are resolved, I'll happily merge your PR!

jeremiah-miller:
Done, thanks for letting me know of those issues!

patrickgold:
Thanks again for your efforts!

There's still one thing with the expended popups: When I compare them with other keyboards like OpenBoard or GBoard, they has a lot of extended characters for some keys, which are not included with your eo.json file. As I don't speak Esperanto, I don't know if some of these characters are important or if the majority of the missing popups can be left out...

jeremiah-miller:
Is the goal with the extended popups to be to have all similar/related characters show in the popup, even if they're not relevant to that language?

patrickgold:
No, the goal is to have all extra characters relevant to the language you are defining, in your case Esperanto. FlorisBoard will then fetch the extended popups based on the selected locale. If a user for instance selects Locale:German/Layout:Esperanto, esperanto.json with extended_popups/de.json will be selected. If a user selects Locale:Esperanto/Layout:Esperanto, esperanto.json with extended_popups/eo.json will be selected. What I meant to say is that OpenBoard and GBoard have quite some extra popup characters defined for the Esperanto language, which are missing in eo.json.

jeremiah-miller:
oh interesting. So then why does en.json have so many extended popups? Few of them seem relevant to English

patrickgold:
You are right, the popups of the first languages I've added, especially English, German, french etc. have a basic subset of extended popups, which aren't necessary for the language. Later as more languages were added to FlorisBoard, the extended popups shifted to be built for the language instead.

The question I am trying to ask you is if the keys in `esperanto.json` and `eo.json` cover the most important keys for Esperanto. If so, I have no problem merging them as they are.

jeremiah-miller:
Ah I see. Yes, all the esperanto keys are already covered by the layout in `esperanto.json` and I just copied `en.json` to make `eo.json` however based on that understanding of how the extended popups work there are a few convenience keys which might be handy to add to `en.json` (such as q popping out of ŝ and a couple others like that), and I need to add and remove some stuff from `eo.json`. Or if you'd prefer, I can leave `en.json` untouched, either way I need to update `eo.json` though

patrickgold:
I'd prefer if you left `en.json` as it is and define all the extended popups in `eo.json`. Sorry if I bother you too much with this topic, but i just want to make sure that I check and discuss changes that are unclear before merging them. :)

jeremiah-miller:
No worries at all, I'd rather get it done right :) So if the `eo.json` file is meant to make all the characters relevant to Esperanto accessible, should I invert all the ones in `en.json`? So If I did that, instead of having 
```
  "a": [
    { "code":  224, "label": "à" },
    { "code":  230, "label": "æ" },
    { "code":  227, "label": "ã" },
    { "code":  229, "label": "å" },
    { "code":  257, "label": "ā" },
    { "code":  225, "label": "á" },
    { "code":  226, "label": "â" },
    { "code":  228, "label": "ä" }
  ],
```
I would have 
```
  "à": [
    { "code":   97, "label": "a" }
  ],
  "æ": [
    { "code":   97, "label": "a" },
    { "code":  101, "label": "e" }
  ],
  "ã": [
    { "code":   97, "label": "a" }
  ],
  "å": [
    { "code":   97, "label": "a" }
  ],
  "ā": [
    { "code":   97, "label": "a" }
  ],
  "á": [
    { "code":   97, "label": "a" }
  ],
  "â": [
    { "code":   97, "label": "a" }
  ],
  "ä": [
    { "code":   97, "label": "a" }
  ],
```
which would ensure that no matter which variation of 'a' is on a different layout, as long as the layout has some form of 'a' on it, the esperanto 'a' would be typable. Or should I target it towards a qwerty layout, so I don't have to worry about weird variations of 'a', just making sure that esperanto would be type-able on a qwerty layout (or ones with all the same keys) and not worrying about the other layouts?

patrickgold:
No, the way the `extended_popups` was designed is that you define all variants of `a` as a popup, based on the base layout for that language (here `esperanto.json`). FlorisBoard then automatically assigns the variants to the key `a`, regardless of where the key is in the keyboard layout. This enables me for instance to display all variants of `a` relevant for Esperanto, even if the base layout is not `esperanto.json`. So:

`base key -> array of all keys relevant to that key`, and that for every key in `esperanto.json`, except if there are no variants (like `f`, `j`, `b`, `m`, `ĵ`, ...).

In your case, 
```
  "ŝ": [
    { "code":  113, "label": "q" }
  ],
  "ĝ": [
    { "code":  119, "label": "w" },
    { "code":  373, "label": "ŵ" }
  ],
  "e": [
    ( all keys relevant to e)
  ],
and so on...
```

I don't know for sure if all key variants OpenBoard and Gboard have are relevant for Esperanto, but it can't hurt imo to include more variants than needed.

Hope that I could clarify on this.

jeremiah-miller:
Thanks! I think it really comes down to whether the extended popups in florisboard should only add popups for the language of the current locale, or for other potential locales/languages as well. It seems like openboard and others include as many as they can, even if they don't apply to that language. Personally I prefer only including the relevant ones and that's what I'll do, but I can see the benefits of including as many as possible. 

Also, since eo.json is only ever used when esperanto is the system locale it will never be accessible unless Florisboard allows for selecting a locale that's different from the system one, since esperanto isn't included as a locale for android. Not a super big deal imo, but it's relevant to note

jeremiah-miller:
Ok, I finally got it. I know you said I shouldn't edit `en.json`, however I added a few things to ensure that all the english characters can be accessed even if the selected language is esperanto

patrickgold:
Thanks for your efforts in fixing the extended popup issue, I really appreciate it. I've retested your changes and didn't find any more issues! 👍 

Regarding the modification of `en.json`: it's ok that you edited it, because I realized that the Esperanto layout is missing the q and w keys, which would be annoying for English users. Still I have to rethink how language independent popup keys interact with the extended popups (as they override each other if defined both in the extended popup file and as a direct popup). I have such an issue in several layouts now, (looking at Dvorak, which is more difficult to implement due to the weird layout). But this is another story and should not hinder passing this PR!

Your changes will be released later today together with the v0.3.1 release. Thanks again for your contribution and have a nice day / evening! :)

jeremiah-miller:
Awesome, thank you! Yeah that's an interesting problem, I might dig through the code later to see what it's doing and see if I can think of some potential solutions

I'm glad I can help, you have a nice day too!

PR #98, opened 2020-12-25
Fix key delete crash on holding down

patrickgold:
Fix that the delete key crashes FlorisBoard when holding down on this key. Related to #97.

PR #97, opened 2020-12-25
Delete Key gesture Improvents

yashx:
### Before

https://user-images.githubusercontent.com/20506602/103137817-7f7f2100-46f2-11eb-9703-caa2eeef7a66.mp4

### Now

https://user-images.githubusercontent.com/20506602/103137820-86a62f00-46f2-11eb-961f-64920cd0a1d4.mp4

## Adds "Delete words precisely"

https://user-images.githubusercontent.com/20506602/103137824-8c037980-46f2-11eb-89d8-7cd450c6122f.mp4

closes #25

PR #99, opened 2020-12-25
Fix left and right arrow in clipboard cursor row

yashx:
## Before

https://user-images.githubusercontent.com/20506602/103139783-c6c2dd00-4705-11eb-8db4-4cf5dc87c851.mp4

## Now

https://user-images.githubusercontent.com/20506602/103139786-d2ae9f00-4705-11eb-92b6-462e8a41cad7.mp4

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt
@@ -233,15 +233,21 @@ class KeyView(
                 isKeyPressed = true
                 florisboard?.keyPressVibrate()
                 florisboard?.keyPressSound(data)
-                if (data.code == KeyCode.DELETE && data.type == KeyType.ENTER_EDITING) {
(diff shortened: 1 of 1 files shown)


RickyM7:
The same crash reported in #97 occurs here but it has already been fixed in #98.

patrickgold:
@RickyM7 yes this will crash because it uses the old TimerTask, with the #98 solution it will work flawlessly. thanks again for trying it out!

@yashx Thanks for adding this! I just have to think of how to now merge together these changes, as there will definitely be merge conflicts now

yashx:
You can merge #98 first. I will make the appropriate changes in this PR so there are no conflicts.

patrickgold:
@yashx thanks! I will add the fix for EditingKeyView, then merge it into master, so you can adjust this change.

yashx:
@patrickgold I think it can be merged now.

patrickgold:
@yashx I've tested your changes and everything checks out. Thanks again for your contributions!

PR #102, opened 2020-12-26
Create canadian_french.json

The-Quantum-Alpha:
Canadian Multilingual (french) --> qwerty, but with éàè

Not *fully* complete, but many of us multilingual canadian would enjoy this a lot!

PR #103, opened 2020-12-26
Canadian french layout

The-Quantum-Alpha:
Added the Canadian omelette du fromage layout

### DIFF
--- a/app/src/main/assets/ime/config.json
+++ b/app/src/main/assets/ime/config.json
@@ -4,7 +4,7 @@
     "qwerty": "QWERTY",
     "qwertz": "QWERTZ",
     "azerty": "AZERTY",
-    "spanish": "Spanish (QWERTY)",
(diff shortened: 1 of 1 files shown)


patrickgold:
You can modify an existing open PR by just pushing another commit to the same patch branch, this way there's only one PR per feature added. GitHub automatically adds the commit to the PR and even recognizes force-pushes to an open PR, if done so. So there's no need to open another PR just to add the entry, you can just do it directly in #102 :)

The-Quantum-Alpha:
Hm, to used to our protocols... 

We are not allowed to modify an existing PR, due to some AI sorting that we have...

Habits to break!

Thanks for the reminder :P



Sincerely,
-TQA
Quantum computer engineer and scientist
D-W. Inc., ON


Dec. 25, 2020, 21:33 by notifications@github.com:

PR #107, opened 2020-12-26
Added Undo Redo Buttons to Quick Actions in Smart Bar

### LINKED ISSUE #70: Allow undo for pasting
Having an undo button when pasting would be great, for when pasting in wrong position.
Hacker's keyboard allows this through ctrl+z.
Having an undo icon in toolbar would be enough.

yashx:
On Whatsapp, telegram and chrome, It works without any issues.
On Keep notes, it works but Quick Action Bar automatically closes after single-use (This shouldn't happen and I can't figure out why it does so) 

## Demo

https://user-images.githubusercontent.com/20506602/103149604-e4895400-4790-11eb-84c6-53d69df6ad30.mp4

closes #70

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/core/EditorInstance.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/core/EditorInstance.kt
@@ -450,19 +450,21 @@ class EditorInstance private constructor(private val ims: InputMethodService?) {
      * @return True on success, false if an error occurred or the input connection is invalid.
      */
     fun performClipboardPaste(): Boolean {
-        val clipData: ClipData? = clipboardManager?.primaryClip
-        val item: ClipData.Item? = clipData?.getItemAt(0)
-        return when {
-            item?.text != null -> {
-                commitText(item.text.toString())
-            }
-            item?.uri != null -> {
-                commitContent(item.uri, clipData.description)
-            }
-            else -> {
-                false
-            }
-        }
+//        val clipData: ClipData? = clipboardManager?.primaryClip
+//        val item: ClipData.Item? = clipData?.getItemAt(0)
+//        return when {
+//            item?.text != null -> {
+//                commitText(item.text.toString())
+//            }
+//            item?.uri != null -> {
+//                commitContent(item.uri, clipData.description)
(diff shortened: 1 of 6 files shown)


patrickgold:
Thanks again for your contribution!

I've tested out your changes on my personal device and found out, that it undos the paste action, but also the input prior to the paste action. Is this intended? Note that this behavior is happening in nearly every input field, regardless of the app. The screen recording below was from the OnePlus notes app.

My device: `OnePlus 7T / Android 10`

### Demo

https://user-images.githubusercontent.com/19412843/103150363-c3fed100-4773-11eb-8c8e-82f15b2fe02d.mp4

RickyM7:
Working perfectly here.

yashx:
@patrickgold The Undo key sends "ctrl+z" and redo sends "ctrl+shift+z" so they have no control over what text is removed or not.

If I am not wrong, the reason this is happening is because of the way text is pasted by the keyboard. The keyboard here is basically reading the data from the clipboard and [writing it for you](https://github.com/florisboard/florisboard/blob/1e690018d7a0819ffb2c4055df8ad48044c68ab3/app/src/main/java/dev/patrickgold/florisboard/ime/core/EditorInstance.kt#L401) so instead of "Paste" being considered as a separate action, It is considered a part of the previous action and "Undo" key removes everything.

yashx:
Okay so I just tried this, sending "Ctrl+V" to paste instead of [doing this](https://github.com/florisboard/florisboard/blob/1e690018d7a0819ffb2c4055df8ad48044c68ab3/app/src/main/java/dev/patrickgold/florisboard/ime/core/EditorInstance.kt#L397-L408) fixes the issue.

patrickgold:
@yashx Now that you say it it makes sense, it is best to just emulate a hardware keyboard in this context too. If you want to, you can make the CTRL-V change, but leave the old paste code as a comment for now. Thanks!

yashx:
@patrickgold I have changed it and left the old code as comments.

patrickgold:
@yashx Thanks for making the change, I really appreciate it! I've tested it again on my device and it works flawlessly now. I will merge it into master! :)

Issue #105, opened 2020-12-26
Typing is quite "sticky"


abh-oss01:
1) The keyboard, and typing using it in general feels very sticky, wobbly, and not so smooth and suitable for typing longer texts.

2) The input has a micro delay of just short of a second, I mean, the time after pressing the letter or a symbol on the keyboard, till the letter/symbol appears on the screen, is just short of a second (Little lag).

**Environment information**
- FlorisBoard Version: 0.3.1
- Install Source: F-Droid 
- Device: rmx 2061
- Android version, ROM: 10, Stock

**Steps to reproduce**
1. Type using the keyboard
2. See/feel the listed problems

patrickgold:
Hmm yeah I will have to investigate this, as normally the reaction time of key presses is really fast. I will add it to my todo list!

abh-oss01:
It's not quite comfortable for typing longer texts.
My device is quite a recent one, plays 4K videos without a sweat, so rendering a small keyboard shouldn't be a problem for it.
No idea why it happens though.
Thanks for the a2a.

RickyM7:
@patrickgold  I also feel it on my phone but I think it's because of the popups that sometimes has a micro delay. Maybe making popups disappear faster can solve that.

yashx:
The likely reason @abh-oss01 feel this is because the text is not written when the key is pressed but when it is [released](https://github.com/florisboard/florisboard/blob/1e690018d7a0819ffb2c4055df8ad48044c68ab3/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt#L309)

patrickgold:
Hmm I am currently trying to improve the logic to decrease any input lag, will push via a PR as soon as I have any noticeable results.

RickyM7:
@patrickgold Switching between hints of a key also has a slight delay. You can see by gif but it is much more perceptive when using the app.
![test](https://user-images.githubusercontent.com/24703825/103179068-1a990780-4867-11eb-926a-eb60606cd0a2.gif)

patrickgold:
@RickyM7 Yeah that's because there is quite a lot of calculation to do in examinating which popup character should be selected and if the touch pointer is out of bounds, etc. Also PopupWindow isn't exactly the single most responsive component in Android, but it does a lot of work for actually drawing over the keyboard, so I am still using it. I can look into improving this bit, maybe try to avoid list operations as often as possible, and other similar things.

abh-oss01:
[...]

PR #108, opened 2020-12-26
Improve input UX and performance

patrickgold:
This PR tries to improve input performance by avoiding unnecessary object allocation. Related to #105

@abh-oss01 @RickyM7 as both of you stated that there's some delay in the input, does the new debug artifact change the input UX, and if so is it now better or worse than before? Thanks in advance!

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/media/MediaInputManager.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/media/MediaInputManager.kt
@@ -17,6 +17,7 @@
 package dev.patrickgold.florisboard.ime.media
 
 import android.annotation.SuppressLint
+import android.os.Handler
 import android.util.Log
 import android.view.MotionEvent
 import android.view.View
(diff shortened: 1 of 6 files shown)


RickyM7:
I think it's improved a little. The biggest problem for me is that sometimes popups appear with a certain delay (sometimes they don't even show up). If you want to see this, simply choose two keys and click quickly and alternately on them.

patrickgold:
Hmm yeah it is not possible that more than one popup is displaying at once, because I am reusing the same PopupWindow and just move it around. This is way cheaper in terms of performance that creating a new PopupWindow for each tap, which FlorisBoard used at the very beginning. I will look into maybe preallocating ~5 PopupWindows, which could be used to provide a better input UX while staying performant.

patrickgold:
Ok so I've messed around with Kotlin coroutines, TimerTasks, replacing PopupWindows with view overlays, but sadly I couldn't gain any more performance. I've now just improved the elevation of the PopupWindow and made the Editing and Media input use the new Handler as well. I will merge this PR now, as I'm not going anywhere with it (but all the changes that are here at least a little bit of improvement). Issue #105 will stay open though until I've found a solution which fixes the input lag on some devices.

PR #110, opened 2020-12-26
Add private mode (aka incognito mode) base

### LINKED ISSUE #106: Incognito mode
Please add an option to force use the "incognito mode" everywhere.

As of now, (v0.3.1), there is no such incognito mode to use.

patrickgold:
This PR adds a basic implementation of a private mode, which prevents features from temporarily saving the input data. This affects the clipboard history (NYI), clipboard suggestions and the next word algorithm adjustments (NYI). The private mode activates either if an app requests it or if the "Force private mode" flag is set in the Advanced Settings. In its current state also rather useless, but it will get important as soon as the composing suggestions are up and running. Below is a screenshot how the indication for the private mode looks like (purple icon in the Smartbar).

Screenshot_20201226-232625-1

Closes #106

### DIFF
--- a/app/src/main/assets/ime/theme/floris_day.json
+++ b/app/src/main/assets/ime/theme/floris_day.json
@@ -47,6 +47,10 @@
     "oneHandedButton": {
       "fgColor": "#424242"
     },
+    "privateMode": {
+      "bgColor": "#A000FF",
+      "fgColor": "#FFFFFF"
+    },
     "smartbar": {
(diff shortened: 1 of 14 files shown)


abh-oss01:
Looks great, thank you.

patrickgold:
@abh-oss01 Thanks for the feedback! I will now merge this PR, the private mode will be released later today together with v0.3.2!

PR #111, opened 2020-12-27
Switch to Timber for Logging

yashx:
It takes care of Tags and only prints Log statement in debug version.


patrickgold:
Thanks for your contribution! I didn't even know Timber existed, but it seems like a better implementation of the default Android Log utility. I've tested it locally and it works like a charm. I'll now merge your changes.

PR #114, opened 2020-12-27
Add hint priority mode setting

### LINKED ISSUE #39: Long-pressing does not select the key as shown on the keyboard
By default, long pressing _should_ insert the symbol shown on the keyboard. For example, the "A" key shows an "@" symbol on it so long-pressing it should open the popup with the "@" symbol selected. This applies to some other keys like "u" and "e" - maybe more.

**Current**
Long-pressing the "A" key opens the long-press popup and selects the "á" symbol by default. 
image

This is unexpected as the keyboard clearly shows that long-pressing the "a" key can be used to insert the "@" symbol.
image

**Desired**
Long-pressing the "A" key opens the long-press popup and selects the "@" symbol. Both the AOSP and AnySoftKeyboard keyboards function this way. I believe Gboard also works this way.

AOSP keyboard
image

AnySoftKeyboard
image

**Environment information**
- FlorisBoard Version: 0.2.3
- Install Source: GitHub
- Device: OnePlus 6T
- Android version, ROM: Android 10 (LineageOS 17.1)

**Reproduce**
1. Long-press and release "a" key without selecting anything from the popup.
2. Notice that "á" is inserted.

patrickgold:
This PR is finally a solution for the discussion, whether a hinted symbol or the accent character should take priority in the extended popup. The preference toggles for the hinted number row and the hinted symbols have been replaced by a hint priority setting for both, which can separately be set to `Disabled`, `Enabled (Hint takes priority)` and `Enabled (Accent takes priority)`.

In the future it is also planned to add an `Enabled (Smart prioritization)` mode, which decides based on the current subtype, which of the two possibilities should be selected.

Closes #39

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/core/PrefHelper.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/core/PrefHelper.kt
@@ -24,6 +24,7 @@ import dev.patrickgold.florisboard.R
 import dev.patrickgold.florisboard.ime.text.gestures.DistanceThreshold
 import dev.patrickgold.florisboard.ime.text.gestures.SwipeAction
 import dev.patrickgold.florisboard.ime.text.gestures.VelocityThreshold
+import dev.patrickgold.florisboard.ime.text.key.KeyHintMode
 import dev.patrickgold.florisboard.util.VersionName
 import kotlin.collections.HashMap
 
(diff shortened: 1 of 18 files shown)


RickyM7:
I tested it and everything worked perfectly here.

-Device: Redmi Note 4 (MIUI 11 - Android 7.0)

patrickgold:
> I tested it and everything worked perfectly here.

That's great to hear, thanks for the feedback!

PR #115, opened 2020-12-27
InputView Code Cleanup

yashx:
Minor changes for readability


RickyM7:
I tested it and everything seems to work perfectly here. One thing I realized is that switching from another keyboard to FlorisBoard seems to occur slightly faster now and this is great.

patrickgold:
@yashx Thanks for your great Pr in the last few days, I really appreciate your efforts and help in making FlorisBoard better! I've reviewed your changes and will momentarily merge your changes, which will be shipped soon with v0.3.2 today.

@RickyM7 I think this is be due to the fact that yashx gets a reference to all the one-handed elements once instead of every single time it is needed.

PR #119, opened 2020-12-31
Improve layout definition structure

patrickgold:
This PR is a rather huge one as it has changes in all (!) layout json files. This PR adds a lot of features, which are listed here:

- Each json file in `assets/ime/` (except themes for now and `config.json`) must include at least the fields `type`, `name` and `authors`. This meta data is important for v0.4.0/Module C, as the modularization into Assets to be used as extensions requires meta data to show in the Settings UI and in the resource cache organization.
- Popup definitions have completely changed:
  - Popups for a single key are now `PopupSet`s, which have a `hint`, `main` and `relevant` field. This change makes it easier for the hint mode setting to do its job. This also rather accidentally adds support for the key hint mode "Smart prioritization", as it relies on the popup set definition, if the `hint` or the `mode` should be chosen. Smart prioritization thus only works if the popups are correctly defined in the json file. I've done my best to do this for some locales I'm familiar with, but there are definitely subtypes which require some adjustments to be perfect in this sense.
  - Popups for a subtype are now defined in a `PopupMapping`, which maps each defined base key to a popup set, categorized into the existing key variations.
- Keys can now be grouped together with the help of group ids. Group 1 and 2 is used to identify the left and right "important" modifier key, where special popups for the settings, emoji, punctuation, etc. are shown. Group 3 is the enter group and also gets unique popups assigned based on the group.
- The code for merging in the popup and hint keys into the layout has been cleaned up.
- Note: this PR _may_ have some issues with performance of the extended popup rendering, but I will resolve this in the next PR, which will be done before the next release comes up.


RickyM7:
Everything looks perfect here.

PR #122, opened 2021-01-04
Rework popup UI implementation

patrickgold:
This PR completely reworks the popup UI implementation and gets rid of the rather slow PopupWindows to a more faster approach via layered views. Additionally, the logic of creating the UI has improved too, which should increase performance of the extended popups. With the rewrite of the popup UI also a lot of unnecessary layout code has been removed. Relevant to issue #105.


RickyM7:
D-dhdhdhdhdhdhdhdhddhdhdhdhdh

RickyM7:
D-dhdhdhdhdhdhdhdhddhdhdhdhdh

RickyM7:
D-dhdhdhdhdhdhdhdhddhdhdhdhdh

RickyM7:
Sorry, there was a weird bug with my GitHub app.
About this PR, it has really greatly improved the speed of the popups, of course if you move too fast it will still have a micro delay but it is almost imperceptible. Typing too fast also slows popups, but that doesn't seem to be a problem of the popups themselves.

patrickgold:
@RickyM7 thanks for your feedback! The problem with typing to fast is that the single popup can either be at key a or b, as it is only 1 view. What I could still try to do is have multiple popup views at once (just in case), so any performance issues when showing/hiding the popup is less noticeable. I know I said this earlier, but kinda forgot about it since I rebuilt the popup UI. I will look into it tomorrow before merging this PR.

Also I think there are other non-popup related issues like the small input delay between finger release and the input actually appearing in the target field, which is more hard to fix but will also definitely addressed by me.

patrickgold:
Above commit fixes a huge performance hit FlorisBoard had and I have just now discovered accidentally: All keys in the keyboard as well as in Smartbar suffered from severe overdraw, which caused a high strain on the CPU. Now the colors are only set once, which makes the idle active CPU usage go down from 25-50% (which was crazily high for a keyboard) to near 0%.

Note though that this temporarily breaks on-the-fly theme switching/adjusting, as the theme is now in some parts only loads in the Service startup and will not update until you switch keyboard and then back again to FlorisBoard. This especially affects the keyboard preview in the Settings. I will fix this with the internal theme engine rework.

@tsiflimagas As you stated in #105 that the idle active CPU usage was high, would it be possible that you install the [debug artifact](https://github.com/florisboard/florisboard/runs/1647466673) `app_debug.apk` and test it out on your device if my changes resolved your reported issue? The debug artifact will install as `FlorisBoard Debug` and will not touch your existing FlorisBoard installation. This would be very helpful, thanks in advance!

tsiflimagas:
@patrickgold ah, works like a charm, thanks for that! Load is back to normal. One thing about the debug build is that when switching to it, it opens in emoji view. Other than this, it feels great, fluidity somehow feels more than normal, even recursive deletion of characters appears to be faster than usual (could be placebo (?))!

patrickgold:
@tsiflimagas Thanks for the feedback, I really appreciate it! I will look into the emoji-on-startup bug (I have personally gotten it rarely today). Recursive deletion speed depends on the current CPU load, so yeah it is definitely possible that this increased the speed.

PR #133, opened 2021-01-08
Create hr.json

HeDidNothingWrong:
Croatian additional symbols (lowercase) - also applicable for Serbia, Slovenia, Bosnia. Casually combined with qwertz

PR #134, opened 2021-01-08
Update config.json to include hr keyboard extension

HeDidNothingWrong:
Update config for croatian extended popup

### DIFF
--- a/app/src/main/assets/ime/config.json
+++ b/app/src/main/assets/ime/config.json
@@ -156,6 +156,11 @@
       "id": 1101,
       "languageTag": "eo",
       "preferredLayout": "esperanto"
+    },
+    {
+      "id": 1201,
+      "languageTag": "hr",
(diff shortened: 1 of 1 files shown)


patrickgold:
Everything also checks out here on first look, will merge it like in #133 after testing it out manually.

Note that you can push to existing PRs to add certain things related to the same component, so opening an additional PR is not necessary normally in this case. As it is now this way, I will handle both separately, but it is just more clear if it is one PR.

HeDidNothingWrong:
Much appreciated as I am new to githib. Have assumed such option exists but haven't found it. Also the code is entitely untested and I don't know whether I need to define "hr" keyboard elsewhere or make an entire layout (which is redundant and seldom used) to make it recognizable as croatian keyboard/extension.

If all works well I will submit serb latin, slovene and montenegrin extensions (last one not being identical).

PR #137, opened 2021-01-08
Add Russian layout; Document adding new languages

### LINKED ISSUE #83: Cyrillic layout
I'd like to have cyrillic (russian) keyboard layout (and russian language)

williamtheaker:
This adds a [JCUKEN keyboard layout](https://en.wikipedia.org/wiki/JCUKEN) for Russian and extends `CONTRIBUTING.md` to document adding new languages.
Closes #83


patrickgold:
Thanks for your PR, I really appreciate it!

All added files look fine. Though does the Russian layout not have any extended popups? In your current configuration, you are missing a `ru.json` file in `assets/ime/text/characters/extended_popups/`. This will cause that FlorisBoard cannot provide any related characters when long-pressing a key.

Also I just noticed that both you and #133 used the id "1201" as subtype id. It would be best if you changed this id to "1301" to avoid duplicate id crashes. Thanks!

Also thanks for adding documentation on how to add new layouts to `CONTRIBUTING.md`!

williamtheaker:
I've changed the subtype id and added documentation addressing that and config for extended popups.

Russian doesn't really use diacritics, but I've added one character and `.ru` to the URI TLD section.

patrickgold:
Thanks for adding this, I really appreciate this! :) Merging your PR as soon as I've finished the theme rework, so it will be included in the next release v0.3.3!

williamtheaker:
Closes #83

bikeroleg (changes requested):
a silly typo

app/src/main/assets/ime/text/characters/jcuken_russian.json
```diff
@@ -0,0 +1,46 @@
...
+      { "code":  1091, "label": "у" },
+      { "code":  1082, "label": "к" },
+      { "code":  1077, "label": "е" },
+      { "code":  1085, "label": "н" },
+      { "code":  1075, "label": "г" },
+      { "code":  1096, "label": "ш" },
+      { "code":  1097, "label": "щ" },
+      { "code":  1079, "label": "з" },
+      { "code":  1093, "label": "х" },
+      { "code":  1066, "label": "Ъ" }
```
bikeroleg, line 19: here you use Ъ (capital hard sign) when it should be ъ (small soft sign); this looks odd on the keyboard and is definitely a mistake which should be fixed before merging.
```suggestion
      { "code":  1066, "label": "ъ" }
```

bikeroleg:
`Russian doesn't really use diacritics`
That's true, they're used _not so often_, but still good Russian keyboards (the absolutely best one is Yandex Keyboard) do implement stressed vowels like а́ я́ ю́ as a non-optional essential feature. I use it occasionally (for example, to remove the ambiguity with homonyms (бóльшая часть vs. большáя часть), for definitions etc). This should be a default or as a GUI-accessible option, without config surgery. I guess I'll open a proper feature-requesting issue after the merge or implement this myself.

app/src/main/assets/ime/text/characters/jcuken_russian.json
williamtheaker: Good catch thank you!

williamtheaker:
It looks like acute Cyrillic characters are two unicode codepoints. I'm not sure how to represent that in the current config.

patrickgold:
You are right, the current config does not allow for multiple codepoints per keys. An idea would be to make an additional `codePoints` field which is an Array<Int>, which when used allows for the `code` field to be omitted. This is a rather breaking change though, as many many lines of if statements rely on the `code` field to be always accessible and initialized. I will though definitely have to find a solution for this (in v0.5.0 milestone), where I also plan to add other non-latin layouts and languages, which may also have multi-code point characters.

patrickgold:
Thanks again for your contribution! I've checked your changes, resolved a merge conflict and will now merge your changes into master. They will be released later this evening with v0.3.3!

slashdevslashurandom:
Thanks for the addition! I should note that this layout is different from that used by the AOSP keyboard and ones derived from it — in these keyboards, to make the top row less cramped, the Ъ key is accessed by long-tapping the Ь key. Adding that change as an option would be really nice.

PR #187, opened 2021-01-14
Trying to add Brazilian Portuguese (pt-BR) Layout

RickyM7:
Objectives of this PR:
1. Order the extended popups characters according to the order seen on the `Gboard`.
2. Add the `.br` extension.

- Note 1: I'm not a programmer so this might have some errors.
- Note 2: I did not change the `config.json` because the `pt-BR` language is already there.

### DIFF
--- a/app/src/main/assets/ime/text/characters/extended_popups/pt-BR.json
+++ b/app/src/main/assets/ime/text/characters/extended_popups/pt-BR.json
@@ -1,0 +1,109 @@
+{
+  "type": "characters/extended_popups",
+  "name": "pt-BR",
+  "authors": [ "rickym7" ],
+  "mapping": {
+    "all": {
+      "a": {
+        "relevant": [
+          { "code":  228, "label": "ä" },
+          { "code":  229, "label": "å" },
+          { "code":  230, "label": "æ" },
+          { "code":  170, "label": "ª" },
+          { "code":  225, "label": "á" },
+          { "code":  227, "label": "ã" },
+          { "code":  224, "label": "à" },
+          { "code":  226, "label": "â" }
+        ]
+      },
+      "c": {
+        "relevant": [
+          { "code":  269, "label": "č" },
+          { "code":  263, "label": "ć" },
+          { "code":  231, "label": "ç" }
+        ]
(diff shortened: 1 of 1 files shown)


RickyM7:
@patrickgold I tested it and it didn't work. Can you help me with that later?

patrickgold:
Thanks for your PR! So if I understood correctly, you just added extended popups for pt-BR and it does not show up in the keyboard when using it with a pt-BR subtype?

RickyM7:
@patrickgold Exactly!

patrickgold:
Ok I found out the problem (see #188 ). It is a bug in FlorisBoard, you did everything correctly :)

I will push a fix after the theme rework, so your changes can start working.

RickyM7:
@patrickgold Got it, thank you!
Btw, I just realized that all default settings for extended popups have an additional quote (") in the [line 81](https://github.com/florisboard/florisboard/blob/master/app/src/main/assets/ime/text/characters/extended_popups/pt.json#L81). Is that a mistake?

patrickgold:
@RickyM7 No, as I want there to be a double-quote character and I have to escape this symbol as it normally means that a string is ending. So `\"` is really a `"`, but requires the backslash so the JSON parser correctly reads this string.

RickyM7:
@patrickgold Got it! Thank you for the explanation.

patrickgold:
@RickyM7 I'm including your changes in the next release v0.3.4 (in 1 week) if that is ok for you, as I have to resolve #188 first and also figure out another extended popup bug, which is too close to today's release to properly search for and resolve.

RickyM7:
@patrickgold Sorry for the delay. No problem, take the time needed.

patrickgold:
@RickyM7 Commit 75c59366a9953bd823510e3c038b38cbabbda8f3 fixes the bug that country-specific overrides were ignored, allowing me to merge your changes. Thanks again for your contribution! This change will be included in tomorrow's release v0.3.4.

RickyM7:
@patrickgold Perfect! I just tested it and everything worked perfectly. Thank you!

PR #212, opened 2021-01-19
Add new swipe actions / Improve gesture pref structure

### LINKED ISSUE #88: Keyboard switching function interferes with gesture
When I try to use the spacebar gesture to navigate between letters of a word most of the times I end up triggering the switching keyboards function, and sometimes even switching keyboards. It would be nice if the function could be disabled, or if it could take more time so that it doesn't trigger while doing this, or maybe something else could work, too.

### LINKED ISSUE #141: Allow for deletion of words and adding of spaces via gesture
Being able to swipe left to delete words and swipe right to add spaces is a very intuitive way of typing on the keyboard Fleksy. After getting used to this way of typing it's very hard to change back and would be great to have with this keyboard.

patrickgold:
This PR adds a few new swipe actions to choose from, improves the grouping of the gestures preferences a bit and allows to freely configure the space bar long press action.

Added new swipe actions (available to all keys except delete): 
- Insert space
- Move cursor to start of line
- Move cursor to end of line
- Show input method picker

These swipe actions are now available for all keys:
- Delete word

Closes #141

As the space bar space bar long press action can now also be freely configured (and set to `No action`, thus disabling the input picker), the space bar input picker interference issue with the cursor left/right is now resolved.

Closes #88


RickyM7:
Everything working perfectly!

patrickgold:
Thanks for the feedback Ricky!

tsiflimagas:
The new implementation of gestures is great! Nevertheless, long press on space bar could be improved (in terms of interfering with the swipe-to-move-cursor gesture) if a small delay could be applied to it. I don't know whether this is possible though.

patrickgold:
@tsiflimagas This is possible. Currently the long press time is dependent on the "Long key press delay" pref in `Settings > Keyboard`. Though increasing this pref's value also increases the time needed for the extended popups to show up, so I don't think it is the best idea to change this just for the space bar gesture.

tsiflimagas:
@patrickgold uh, damn, I hoped it would be possible to add delay only for one key. Yeah, I just increased the long press delay and it worked as expected indeed, but totally doesn't worth it only for one key. Anyway, nothing serious, I'll just disable long press and use the swipe up gesture for the space bar, this one can't interfere with swiping!

tsiflimagas:
Ps I tried different keyboards (gboard, simple keyboard, openboard, aosp keyboard) and by observing, I found out that gboard has hardcoded longpress delay for spacebar bar and delete key, and the rest of them only for delete key.

PR #218, opened 2021-01-21
Fix precise character deletion behaviour

### LINKED ISSUE #124: Deleting words process is kinda slow
The app is amazing but it's seems slow to delete something 
Florisboard

https://user-images.githubusercontent.com/52888633/103749708-25416280-5006-11eb-9556-c97b1734106c.mp4

Open board

https://user-images.githubusercontent.com/52888633/103749930-6df91b80-5006-11eb-858c-74e4a4eddc1d.mp4

### LINKED ISSUE #154: Make delete-swipe more similar to the one in OpenBoard
1. In OpenBoard, delete-swipe works a little differently. If you swipe left, then swipe all the way to the right, no text is selected.
In FlorisBoard, you can't completely reverse your deletion
-selection... at least sometimes you can't.
I think the OpenBoard-behavior makes more sense because... well... the user returned to his original position with his thumb. Thus nothing should be selected&deleted.

2. (Openboard's delete-swipe deletes more / is faster. In my opinion, that is better.
Of course spacebar-swipe and delete-swipe should have the same speed though)

patrickgold:
This PR is an attempt to fix the precise character deletion behavior for the delete key. Until now, the amount of characters the swipe selected was dependent on the speed of the swipe as well as on the current CPU load. Now, the amount of characters is calculated based on the distance you swiped and thus the behavior is way more consistent in all situations, independent of the speed and CPU load.

Closes #154 

The delete swipe is now also generally more fast and accurate, similar to OpenBoard or Gboard. Please note that the amount of charcters is dependent on the `Swipe distance threshold` value in the Gesture preferences.

Closes #124 

### Demo (old v0.3.3 vs new v0.3.4)

https://user-images.githubusercontent.com/19412843/105382125-0a7b0a80-5c10-11eb-9ef2-7780258a4fcd.mp4

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/gestures/SwipeGesture.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/gestures/SwipeGesture.kt
@@ -26,6 +26,36 @@ import kotlin.math.*
  * Wrapper class which holds all enums, interfaces and classes for detecting a swipe gesture.
  */
 abstract class SwipeGesture {
+    companion object {
+        /**
+         * Returns a numeric value for a given [DistanceThreshold], based on the values defined in
+         * the resources dimens.xml file.
+         */
+        fun numericValue(context: Context, of: DistanceThreshold): Double {
+            return when (of) {
+                DistanceThreshold.VERY_SHORT -> context.resources.getDimension(R.dimen.gesture_distance_threshold_very_short)
+                DistanceThreshold.SHORT -> context.resources.getDimension(R.dimen.gesture_distance_threshold_short)
+                DistanceThreshold.NORMAL -> context.resources.getDimension(R.dimen.gesture_distance_threshold_normal)
+                DistanceThreshold.LONG -> context.resources.getDimension(R.dimen.gesture_distance_threshold_long)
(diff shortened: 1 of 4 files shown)


Glitchy-Tozier:
This looks amazing!

RickyM7:
That's perfect now.

patrickgold:
Thanks for the feedback! I will merge the PR now, this change will be released later today with v0.3.4.

tsiflimagas:
Great work! Could the same be applied to swipe-to-move-cursor gesture?

Glitchy-Tozier:
Yessss, works great!!!

As @tsiflimagas suggested, this would alo be a great improvement to the spacebar-swipe!

patrickgold:
@tsiflimagas Yes, this shouldn't be difficult to port the new swipe logic to the space bar too.

@Glitchy-Tozier Just curious: did Google Play already deliver v0.3.4 to you? If so this means releasing updates in the middle of the week is faster that doing so on Sunday evening.

Glitchy-Tozier:
I think this swipe-speed should be controlled by the same variable as delete-speed.
This makes sure, the have the same sensitivity.

> Just curious: did Google Play already deliver v0.3.4 to you? If so this means releasing updates in the middle of the week is faster that doing so on Sunday evening.

Yup.
I thought it was yesterday, but apparently it actually hit the play store two days ago.

PR #232, opened 2021-01-23
Fix space bar long press

### LINKED ISSUE #228: Long-press improvement
### Issue
**(Assuming the space-bar long-press-action is set to "none / no action")**

When long-pressing on spacebar, nothing happens. Which, in my opinion, is wrong.

**It shouldn't be**

_long press is triggered → do nothing_.

**Instead it should be**

_Long press NEVER EVEN GETS ACTIVATED → space-bar behaves as if it was short-pressed_.

Meaning, that if there was no swiping going on, when the finger releases the space-bar, input `space`.

I assume, the "insert space" option was intended to do this, but this option is no replacement for a real "long-press disabled" option.
Why? Because "insert space" doesn't insert `space` when releasing the space-bar. It triggers like a regular long-press. Meaning that with this, we'd return to the https://github.com/florisboard/florisboard/issues/88 -problem.

### Request:
Add a switch that lets the user turn off long-presses for the space-bar.

This makes it only input space once released. Just like this button works:

patrickgold:
This PR fixes the space bar long press action in several ways:

- The long press delay is now exactly twice as long as the time set in the preferences. This ensures that it is more difficult to accidentally trigger the long-press action while using the space swipe left/right. The long-press time of any other key is not affected by this change.
- `No action` as well as `Insert space` now both insert a space exactly when the space bar is released, regardless how long it has been pressed. This enables the space bar to behave like a space bar when set to these cases. Previously, `No action` did nothing and `Insert space` did insert the space after exactly the `delay time` set in the preferences, which is not how this should behave. Both actions do not block the swipe left/right action. Closes #228.

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt
@@ -245,13 +245,22 @@ class KeyView(
                             }
                         }
                     }
-                    longKeyPressHandler.postDelayed(delayMillis) {
-                        if (data.popup.isNotEmpty()) {
-                            keyboardView.popupManager.extend(this, keyHintMode)
+                    if (data.code == KeyCode.SPACE) {
+                        longKeyPressHandler.postDelayed((delayMillis * 2.5f).toLong()) {
+                            when (prefs.gestures.spaceBarLongPress) {
+                                SwipeAction.NO_ACTION,
+                                SwipeAction.INSERT_SPACE -> {}
+                                else -> {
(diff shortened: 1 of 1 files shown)


tsiflimagas:
That's really great, thank you for taking care of it! While it should be fine for most users, there could still be some interference when setting long press delay very low (I use 100-120ms). Do you think that making the delay time for space bar three times longer would be ok? Then, the minimum for space bar would be 300ms, which is quite safe.

patrickgold:
@tsiflimagas Thanks for the feedback! IMO a x3 multiplier is too much, but a x2.5 multiplier (instead of x2) is ok for me, which I have changed in the commit. Will merge these changes now, they will be released in v0.3.6 today evening.

PR #234, opened 2021-01-24
Improve theme editor UI and UX

### LINKED ISSUE #222: Improving the usability of Theme Editor for non-advanced users
Thats about it :)

patrickgold:
This PR is an attempt at improving the theme editor UI and UX. In detail, this is what changed:

- The theme editor now has a help icon in the top right corner which leads to [this](https://github.com/florisboard/florisboard/wiki/Using-the-new-Theme-Editor) wiki entry. The referenced wiki entry is a guide to better understand some of the more complex things you can do with the theme editor, which may not be known upfront.
- Group name and attribute name inputs now also check if the given name already exists. While this didn't lead to crashes, duplicate attributes and groups overrode each other and created strange results. This is now not possible anymore and eliminates a possible error source.
- The group and attribute names now display localized strings instead of the raw English names. This is especially helpful for non-advanced users on non-English devices, where the feeling is more consistent with the rest of the app. Custom attributes are labeled "Custom" and display the raw English name besides it.
- The default color for a new attribute you create is now solid black (`#FF000000`) instead of transparent (`#00000000`). This improves the process of creating a new color attribute.
- Clean up of the theme string resources.
- Document the `ThemeEditorActivity` class

Closes #222

### Comparison old vs new (on an English device)

Old | New
----|------
theme_editor_old | theme_editor_new

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/theme/Theme.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/theme/Theme.kt
@@ -16,7 +16,9 @@
 
 package dev.patrickgold.florisboard.ime.theme
 
+import android.content.Context
 import android.graphics.Color
+import dev.patrickgold.florisboard.R
 import dev.patrickgold.florisboard.ime.extension.Asset
 
 /**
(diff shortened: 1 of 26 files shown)


Triplesweet:
@patrickgold thanks alot for doing this and one little suggestion. How about changing the word of "forground" with something simpler like "key symbol" or "key icon".
It is from someone who had too google it (eng isnt my first language) but other things looks much easier to understand.

patrickgold:
@Triplesweet hmm it is difficult because I cannot know what exactly will be in the "foreground". Sometimes it is a letter, so `text color` would be appropriate, sometimes it is an icon, so `icon color` would fit. Also, the key foreground colors both letters and icons at once, so I don't know what to write here. `Text and icon color` doesn't seem to good for me. I've also tried to find some alternative word for "foreground" but the synonyms make things worse IMO.

What I have to say though is if the theme strings are translated into your language (what they aren't as the strings are new atm), it will be easier for you to understand what is meant I think. So I will let "foreground" in for now, maybe in some future version I will find a better word for it.

PR #240, opened 2021-01-25
Improve adaptive theme / Fix color dialog cache bug in theme editor

### LINKED ISSUE #226: Improvement in theming: adapt colour to app
When adapt colour to app opt-in is turned on
There is  lag sometimes when opening an app like signal or WhatsApp
But this doesn't happen when the option is turned off
Also sometimes it switches to other keyboards
 
Screenshot_20210123-102218284

There is also some problems adapting keyboard in vlc app while using adapt colour to app option

Screenshot_20210123-102205791

I hope my suggestion is understandable for you 
Thank you very much for developing Florisboard
Version-0.3.4
Android 10
 resurrection Remix os vanilla

### LINKED ISSUE #237: Previous chosen theme-color resets upon reopening color-picker.
I feel like you need to long- press a color to actually activate it (?)

Basically, the color always reyets to the previous color after clicking on 'ok'.

**Environment information**
- FlorisBoard Version: 0.3.4
- Device: Samsung Galaxy S9
- Android version, ROM: 10

**Steps to reproduce**
Just try to change the main color.

Are there any quick video-sharing platforms? I can't add the video to a comment for some reason.

patrickgold:
This PR improves adaptive themeing and additionally fixes small bugs in the theme engine and editor. In detail, this is what changed:

- The adaptive theme process now is more consistent and checks for the `targetActivity` attribute. Also, if the launch activity has no theme set, the application is checked if a theme is set there. This makes the colors more consistent and for most apps colors are found.
- All parsed colors are forced to be non-transparent to fix issues with see-through keyboards.
- The key `foregroundPressed` attribute is now applied when the key is pressed (was previously ignored in some circumstances).
- The Smartbar `foregroundAlt` attribute now also adapts to the target app theme.
- The navigation bar foreground now also adapts the the background color parsed from the target app.
- Popups now take the `colorAccent` (or for MaterialComponents themes `colorSecondary`) color. This styles the popups (instead of being always grey) and creates more unique styles on a per-app basis.

In general, the adaptive theme process gives good results since this overhaul. Still, there are some apps where the colors are all white/black, etc. and the adaptive theme may not look that good.

Closes #226

Fix that the selection color in the theme editor did not override the cache if the color dialog was reopened before hitting "OK". Additionally, the color edit button's background color now matches the currently cached color. 

Closes #237

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt
@@ -829,7 +829,11 @@ class KeyView(
                     }
                 }
             }
-            labelPaint.color = themeValueCache.keyForeground.toSolidColor().color
+            labelPaint.color = if (isKeyPressed && isEnabled) {
+                themeValueCache.keyForegroundPressed.toSolidColor().color
+            } else {
+                themeValueCache.keyForeground.toSolidColor().color
+            }
             labelPaint.alpha = if (keyboardView.computedLayout?.mode == KeyboardMode.CHARACTERS &&
                 data.code == KeyCode.SPACE) { 120 } else { 255 }
             val centerX = measuredWidth / 2.0f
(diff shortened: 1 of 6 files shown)


tsiflimagas:
Great improvement!

PR #248, opened 2021-01-27
Fix symbols size

### LINKED ISSUE #203: Ability to change the font size of the symbols keyboard and the alphabet keyboard separately
If I set key font size to 100%, when I switch to symbols it goes up about 15%.
It would be nice if I could lower it to match each other as it's annoying that it is not uniform in size.

p0lbang:
This PR should fix #203

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/core/InputView.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/core/InputView.kt
@@ -113,9 +113,11 @@ class InputView : LinearLayout {
         val shouldGiveAdditionalSpace = prefs.keyboard.numberRow &&
                 !(tim.getActiveKeyboardMode() == KeyboardMode.NUMERIC ||
                 tim.getActiveKeyboardMode() == KeyboardMode.PHONE ||
-                tim.getActiveKeyboardMode() == KeyboardMode.PHONE2)
+                tim.getActiveKeyboardMode() == KeyboardMode.PHONE2 ||
+                tim.getActiveKeyboardMode() == KeyboardMode.SYMBOLS ||
+                tim.getActiveKeyboardMode() == KeyboardMode.SYMBOLS2)
         if (shouldGiveAdditionalSpace) {
-            val additionalHeight = desiredTextKeyboardViewHeight * 0.18f
+            val additionalHeight = resources.getDimension(R.dimen.textKeyboardView_baseHeight) * 0.18f
             baseHeight += additionalHeight
             baseTextInputHeight += additionalHeight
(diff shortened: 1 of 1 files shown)


patrickgold:
Thanks for your contribution!

I agree that this fixes #203, though it introduces the height change problem again, as the keyboard view height changes while being shown, causing some not-so-nice content flickering and jumps. This is the reason why the keys are currently so big. I think it would be better to change the actual symbol size to match the one of the character layout.

app/src/main/java/dev/patrickgold/florisboard/ime/core/InputView.kt
```diff
@@ -113,9 +113,11 @@ class InputView : LinearLayout {
         val shouldGiveAdditionalSpace = prefs.keyboard.numberRow &&
                 !(tim.getActiveKeyboardMode() == KeyboardMode.NUMERIC ||
                 tim.getActiveKeyboardMode() == KeyboardMode.PHONE ||
-                tim.getActiveKeyboardMode() == KeyboardMode.PHONE2)
+                tim.getActiveKeyboardMode() == KeyboardMode.PHONE2 ||
+                tim.getActiveKeyboardMode() == KeyboardMode.SYMBOLS ||
+                tim.getActiveKeyboardMode() == KeyboardMode.SYMBOLS2)
         if (shouldGiveAdditionalSpace) {
-            val additionalHeight = desiredTextKeyboardViewHeight * 0.18f
+            val additionalHeight = resources.getDimension(R.dimen.textKeyboardView_baseHeight) * 0.18f
```
p0lbang, line 120: when i changed it to this
val additionalHeight = baseTextInputHeight / 4
this should fix the flickering size changes. im pretty much new in using git so I dont know ho to do this properly hope this somewhat helps in figuring this bug.

patrickgold: The problem is still there: The keyboard suddenly changes height. Even if wen can eliminate the flickering within FlorisBoard, not every app's content and behavior is optimized for sudden height changes. This is why I want to completely avoid changing the height while the user does something in FlorisBoard.

p0lbang: Ohh okay. Sorry for the inconvenience 😅

sirsleon:
The issue is still present, any updates to this issue?

Screenshot_20210202-212850
Screenshot_20210202-212842

patrickgold:
@sirsleon No, at the moment not. A fix for this will be pushed once I am having the time to do so.

PR #275, opened 2021-01-29
Serbian layouts

GrbavaCigla:
### Things I changed:
- Serbian locale
- Serbian latin layout
- Serbian cyrillic layout
- Python script that updates `code` values for each `label`. Might be useful for someone adding new layouts

Tested on android 9 (API 28), on emulator.


patrickgold:
Thanks for your contribution!

I've had a quick overlook over your additions and the layouts itself seem to be fine. In `config.json`, the ID should be changed to `2101`, as `2001` has already been taken by the previous PR.

Regarding the Python script, I personally don't know how useful it is for everyone (as quite a few contributors build layouts on devices which can't even run Python), but it definitely can come in handy. Could you just move it into a new top-level directory (`utils`), so it doesn't lay around in the project root?

If those issues are resolved, I will solve the merge issues caused by Git (not your fault) and merge your changes into master.

GrbavaCigla:
Is this now good?

patrickgold:
Thanks for fixing my mentioned issues, I really appreciate it!

I have tried your layout out on my testing device and have two issues:
a. In the Serbia (QWERTZ) layout, you have 10 character keys in the third row, which looks kinda odd, as the keys are so squashed together. Wouldn't it look better to add `đ` as a popup to `d` and `ž` as a popup to `z`? The other accent characters of c and s are perfectly fine and look really good in row 1 & 2.
b. Again in Serbia (QWERTZ), it would be good to add extended popups in addition to all base characters which have accents. At minimum, even though they have a base key in the keyboard already, I would recommend the following:
- `c` -> `č`, `ć`
- `d` -> `đ`
- `s` -> `š`
- `z` -> `ž`

If there are any other popups which are rarely used but could be relevant, they should also be added.

The Serbian cyrillic layout looks perfectly fine, I have no issues for this. Also thanks for solving the merge conflict!

If these issues can be fixed (these are the last issues I promise :) ), I will happily merge your PR!

GrbavaCigla:
Actually only now I found that I have duplicate letter in latin serbian in third row...

GrbavaCigla:
How do I add extended popup only to latin and not to cyrillic? I fixed duplicate letter in latin. Also since croatian and serbian are the same languages (like English and American), extended popup for croatian is the same for serbian (latin).

patrickgold:
> How do I add extended popup only to latin and not to cyrillic?

You can't directly, because both the Latin and Cyrillic version have the same code (`sr-RS`). Though as you only define popups for Latin characters, this will not cause a problem for the Cyrillic variant. As I have just seen that the `hr.json` file indeed contains all suggested popups, I think it would be best to copy the `hr.json` file to `sr.json`, change the meta data on line 3-4 (leave 'hedidnothingwrong' as the first author and add your name as the second value as you are copying his file) and adjust the TLDs in "~right" (at the very end of the file). Then I think we should be good to go with merging this into master!

> just to clarify, third row was squashed because of the duplicate, now it is not that squashed.

Ah that was the reason why it was so squashed, I didn't even see the duplicate letter. Good catch! This being the case, you can let the `đ` and `ž` key in the third row, it looks pretty good then.

GrbavaCigla:
Done :D

patrickgold:
Perfect, thank you again for your efforts! Merging your changes in now.

PR #288, opened 2021-01-31
Improve moving gestures detection and UX

### LINKED ISSUE #125: Cursor leaving the input field
It is possible that you can make the cursor move out of the input field by using the keyboard's navigation keys, or the space bar swiping gesture. You can then probably see that some other, nearby element has been selected, as if the cursor moved into the close environment and triggered it, e.g. the app's enter button's pressing effect (the shade/ripple effect) gets triggered. I hope my description of the situation is clear enough 😅

### LINKED ISSUE #281: Spacebar swipe doesn't work like delete-swipe.
The sensitivity varies depending on how fast you swipe.Which _would_ be fine, but it's implemented exactly the wrong way.

Please use the same mechanism for spacebar-swipe as you do for delete-swipe.

Delete-swipe-pr: https://github.com/florisboard/florisboard/pull/218

**Environment information**
- FlorisBoard Version: 0.3.5
- Device: Samsung Galaxy S9
- Android version, ROM: 10

patrickgold:
This PR is an attempt to improve the swipe detection mechanism for moving swipe actions (space bar and precise character deletion). I detail, this is what changed:

- The `SwipeGesture` detection utility got an overhaul improvement in the detection of gestures. It should be slightly faster (as it ditches dynamic sized arrays) and the code base has been cleaned up.
- Moving swipes are detected way more precisely and thus will provide a faster feedback to movements.
- Space bar swipe actions now use the new moving detection algorithm, which fixes the inconsistent behavior of the amount of characters selected depending on the speed and CPU load. (Closes #281)
- Space bar swipe left/right can now also be used to move the cursor in the raw input of apps like Termux or SSH clients, as long as the cursor movement is controllable via `ARROW_LEFT` and `ARROW_RIGHT`. The new detection algorithm works seamlessly with these apps and the amount of characters moved is as consistent as in normal textboxes. Note, that the key delete is not supported in raw input editors, as there's no standardized way to select text.
- As the new space left/right code uses the `InputConnection` methods (except for raw input editors like Termux), the arrow can now not move outside the textbox when using the space bar swipe. (Closes #125)

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/gestures/SwipeGesture.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/gestures/SwipeGesture.kt
@@ -63,36 +63,59 @@ abstract class SwipeGesture {
      * @property listener The listener to report detected swipes to.
      */
     class Detector(private val context: Context, private val listener: Listener) {
-        private val eventList: MutableList<MotionEvent> = mutableListOf()
-        private var indexFirst: Int = 0
-        private var indexLastMoveRecognized: Int = 0
+        private var firstMotionEvent: MotionEvent? = null
+        private var lastMotionEvent: MotionEvent? = null
+        private var absUnitCountX: Int = 0
+        private var absUnitCountY: Int = 0
+        private var unitWidth: Double = numericValue(context, DistanceThreshold.NORMAL) / 4.0
 
         var distanceThreshold: DistanceThreshold = DistanceThreshold.NORMAL
+            set(value) {
+                field = value
+                unitWidth = numericValue(context, value) / 4.0
+            }
         var velocityThreshold: VelocityThreshold = VelocityThreshold.NORMAL
 
-        fun onTouchEvent(event: MotionEvent): Boolean {
+        /**
+         * Method which evaluates if a given [event] is a gesture.
+         *
+         * @param event The MotionEvent which should be checked for a gesture.
+         * @param alwaysTriggerOnMove Set to true if the moving detection algorithm should always
+         *  trigger, regardless of the distance from the previous event. Defaults to false.
+         * @return True if the given [event] is a gesture, false otherwise.
+         */
+        fun onTouchEvent(event: MotionEvent, alwaysTriggerOnMove: Boolean = false): Boolean {
             try {
                 when (event.actionMasked) {
                     MotionEvent.ACTION_DOWN,
                     MotionEvent.ACTION_POINTER_DOWN -> {
-                        clearEventList()
-                        eventList.add(MotionEvent.obtainNoHistory(event))
+                        resetState()
+                        firstMotionEvent = MotionEvent.obtainNoHistory(event)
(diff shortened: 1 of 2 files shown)


RickyM7:
Everything perfect about moving the cursor, however on the raw imput the gesture works but does not dynamically show the movement of the cursor.

FlorisBoard v0.3.5 vs Florisboard Debug (Tested in the "Acode" app (I believe it will serve to test raw imput))

https://user-images.githubusercontent.com/24703825/106379056-1db47580-6388-11eb-9aeb-57579d680fd6.mp4

patrickgold:
> Everything perfect about moving the cursor, however on the raw imput the gesture works but does not dynamically show the movement of the cursor.
FlorisBoard v0.3.5 vs Florisboard Debug (Tested in the "Acode" app (I believe it will serve to test raw imput))

I've had a quick glance on "Acode" and it seems that it is built via Cordova, which means the app is basically a web page (but specifically built to be an app for Android only). So this does not qualify as raw input, as it still allows for input connection methods to be used. A true raw input is a terminal app like Termux or any SSH client.

Still, this is a bug, but I am currently unsure if FlorisBoard or Acode is the problem. I have just installed the app and tested it out, and I have observed the following:

```
    imeOptions: action=IME_ACTION_NONE flags=IME_FLAG_NO_EXTRACT_UI|IME_FLAG_NO_FULLSCREEN
    initialCapsMode: flags=(none)
    initialSelStart: 5
    initialSelEnd: 5
    inputType: class=TYPE_CLASS_TEXT variation=TYPE_TEXT_VARIATION_WEB_EDIT_TEXT flags=TYPE_TEXT_FLAG_MULTI_LINE
    packageName: com.foxdebug.acodefree
```

- Based on the above debug output we can confirm that it is indeed a web input field. Web pages though work flawlessly in Firefox and Chromium-based browsers, so this shouldn't be the problem.
- When setting the selection, Acode correctly calls the `onUpdateSelection()` callback and also supplies the correct selection values, which indicates that FlorisBoard correctly evaluates and sets the selection.
- Gboard does work with the Acode input field, but none of the section updates show in the Acode UI, which seems to me that Acode may have a flawed input logic there.
- OpenBoard does work and even Acode shows updates in the UI, as it seems to send hardware emulated Arrows instead of using the input connection.

Based on the above points I do believe that the problem is more with Acode than FlorisBoard's logic. As other input fields work flawlessly in Acode (e.g. Console), I highly suspect a problem with the text editor main input logic, but only for the editor input field. May be worth to file an issue over at https://github.com/deadlyjack/code-editor to confirm on that.

RickyM7:
@patrickgold I just tested Termux and the gesture of moving the cursor worked perfectly, it really seems that what happened in Acode is in fact a bug in Acode itself.

patrickgold:
@RickyM7 Thanks for confirming on that! I will then merge this PR, should something still be adjusted at the movement gesture algorithm just open a new issue for this or comment to this PR.

Glitchy-Tozier:
Great job, it works really well!  :)

PR #319, opened 2021-02-05
Edit bépo layout to add "êà" keys

Salamandar:
Those keys have two uses : 
* Be present
* They move the existing keys so that they are correctly aligned. In standard bépo, H,F are right below N,M. That's currently not the case.


patrickgold:
Thanks for your contribution! I've reviewed your changes and everything checks out. Will merge them into master now.

Salamandar:
Thanks a lot !!

PR #326, opened 2021-02-08
Use haptic feedback by default if available

### LINKED ISSUE #255: Add support for haptic feedback on Pixel devices
Is it possible to add the haptic feedback on the devices that supports it (example pixel), instead of the regular vibration?

serebit:
Resolves #255.

My original plan was to check for the device's haptics support, and use that information to determine whether or not haptic feedback should be enabled by default. A separate setting was to be provided in FlorisBoard settings, labeled "Use haptic feedback for vibration", which would be enabled if the device supported haptic feedback and disabled otherwise. This didn't pan out, as there's no function that can query whether or not a device supports haptic feedback.

As a result, I've elected to modify the default vibration behavior when a user taps a key. If the vibration strength is set to "System default", the device will first attempt to perform haptic feedback; if it succeeds, the vibration function completes. If the attempt fails, the function will fallback to the old behavior of using a default vibration strength of 36. If a manual vibration strength is set in FlorisBoard preferences, haptic feedback will be disabled. This behavior is consistent with the default AOSP keyboard.

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/core/FlorisBoard.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/core/FlorisBoard.kt
@@ -210,6 +210,8 @@ class FlorisBoard : InputMethodService(), ClipboardManager.OnPrimaryClipChangedL
         baseContext.setTheme(currentThemeResId)
 
         inputWindowView = layoutInflater.inflate(R.layout.florisboard, null) as InputWindowView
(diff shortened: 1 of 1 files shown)


patrickgold:
Thanks a lot for your contribution! Looking at your changes, everything looks fine. As I don't have a device which supports haptic feedback where I could test this out, I am gonna assume that this works as intended. Merging into master now.

Issue #325, opened 2021-02-07
Next-word-predictions development discussion thread


patrickgold:
Hi all!

The time has finally come, I am working on dictionary suggestions! As I want the development of this feature to be as transparent and open as possible, I will share the roadmap for the next 2-3 weeks with you. This issue thread serves as a room for general discussion about the implementation and feedback given in this issue will directly influence the early development of suggestions. The development of word suggestions is rather big, so I split it up in four main phases. After finishing each phase, the already implemented part of the suggestions will be released, so all users on F-Droid and Google Play can try it out. Only if you want to, you can also install the debug artifact builds which get created for each commit in the PR, so you can try out new stuff even faster.

During development, all preferences regarding this topic will be marked as [EXPERIMENTAL] and all string resources regarding this feature will be non-translatable. I am doing this to prevent circumstances where I change a string like 5 times and so the translators get notified 5 times to translate the same string over and over again, even though often only a small thing has changed. Nearing the end of the last phase, I will unlock these strings for translation.

For each phase beginning from phase 1, a separate PR will be filed and linked here.

NOTE: below phases are currently crossed out because the time plan got completely overthrown by the OOM issues as well as the Flictionary implementation not working out as it should. I don't have a concrete plan for suggestions, but I think I will implement spellchecking (and suggesting spell-checked words) first, as this part is a bit easier to add than next-word predictions. The old Flictionary implementation will stay in for now until I've found a replacement.

### Phase 0
~This phase has technically already finished. It includes collecting ideas for which sources can be used (a big thanks to everyone who commented on #34!!), providing a way to have standardized dictionary formats in the APK (see https://github.com/florisboard/dictionary-tools) as well as reading and understanding how word suggestions should work. As this phase didn't change a thing in the application itself, no release for this phase is being made.~

### Phase 1 (v0.3.8) (#329)
~This is the "real" first phase, in which the input logic code gets an overhaul. This especially changes the `EditorInstance` logic to be more performant while providing better context to the prediction logic. Also, simple tests are conducted if the created binary dictionaries from the dictionary-tools are actually readable on Android and are built correctly. Then, the logic for providing words for the current input is created. Note, that the predictions in this phase are quite "dumb", as the next-word algorithm is missing.~

~Supported languages in this phase:~
- ~English~

### Phase 2 (v0.3.11)
[...]

PR #329, opened 2021-02-09
Suggestions implementation: Phase 1

patrickgold:
This PR is a status update for the suggestion implementation. For general information about the implementation of suggestions, see #325. In detail, this is what phase 1 changes:

- Complete overhaul of the EditorInstance core class. This was needed because the old implementation had some issues in its structure and its way of caching the current input text. The new implementation now automatically only caches a reasonable amount of characters (192+64 at max) to speed up performance and ease memory use. Additionally, the current word as well as words around the cursor are detected, which can be used by the suggestions to have a better context for generating new words to display.
- First simple implementation of a Dictionary and usage of the binary dictionaries created with the dictionary-tool and a simple algorithm to provide composing suggestions to the user.
- Preferences adaption to include experimental toggles to enable suggestions.
- In this phase, only `en_US` (English (United States)) is supported and suggestions are created regardless of the active subtype. Dictionary selection based on subtypes will come with phase 2.


sabzo:
@patrickgold How's the implementation going? Is there proof-of-concept I or others can look at?

patrickgold:
@sabzo I haven't committed in the last two days because I've been working on writing the encoder for Flictionaries in Python, where I noticed that my address/n-gram tree proposal isn't that memory efficient as I thought (the file size jumped from 1.7MB to about 4.4 MB which is unacceptable). So I had to roll back to a previous idea and then incorporated the new things like bi/trigrams etc. and shortcuts into the file spec. See https://github.com/florisboard/dictionary-tools (main branch) for the latest code. The dict-tool does not support all features of the Flictionary spec yet, but in theory the file spec covers everything needed to store a language model (https://github.com/florisboard/dictionary-tools/blob/main/flictionary.md). I've also limited the order of n-grams in the range 1..8 (both inclusive), which IMO is more than sufficient to save any language model. The problem is that I am storing the number of the n-gram order in a 3-bit value, which allows only 2^3=8 values.

On the Kotlin side, I've also made progress. I've rewritten the decoding method of Flictionaries, which supports all features the Python encoder supports thus far. Additionally I am able to create unigram predictions based on the tokens of the current word, which also suggests words with a max edit distance of 1 (if you uncomment the if statement for this). This sounds great at first, but unfortunately the performance of my search algorithm is more than catastrophic thus far, even on my fast OnePlus 7T. I have a few ideas though how I can approach this problem, which I will work on tomorrow.

If you want to do a code review, I would really appreciate this, but I think it makes more sense to let me clean up and improve the search logic first, because else the list of things which are incorrect/should be improved is quite long :)

sabzo:
@patrickgold definitely, I think cleaning up a bit and when you're ready may probably the most efficient or easier process! Let me know when ready.

patrickgold:
@sabzo Above commit reworks the Kotlin runtime representation of the n-grams (looks similar to how it is also saved to the disk within the Flictionary). This allows me to search for a word faster because I do not have to loop through all words until I have a sufficient amount of words. The responsiveness of the UI has vastly improved. While previously the worst case search times were about ~500ms, we are now at about ~80ms (best case at around 0.5ms). This is still much, but I am slowly but surely decreasing the processing time of the prediction algorithm. One thing that does not work with the new representation at the moment is the dit distance, which I will fix tomorrow.

I have another question about NLP which arose while I thought about other languages: Generally speaking, how should I handle capitalization in the language model? While in English this isn't that important, other languages like German require the first letter of nouns to be written in uppercase. Should I save both the capitalized and un-capitalized word in the n-gram tree or always the un-capitalized one?

sabzo:
> One thing that does not work with the new representation at the moment is the dit distance, which I will fix tomorrow.

Looking forward to the edit distance working!

> I have another question about NLP which arose while I thought about other languages: Generally speaking, how should I handle capitalization in the language model? While in English this isn't that important, other languages like German require the first letter of nouns to be written in uppercase. Should I save both the capitalized and un-capitalized word in the n-gram tree or always the un-capitalized one?

In this case you would have to add both the capitalized and uncapitalized words -- it's the only way for the language model to be aware of both cases. The language model predicts based on the tokens it was trained on.

Also on the current binary dictionary you put there are some improvements that could be made, but again I strongly believe that should be abstracted out for those with more grasp for NLP. For example for the english dictionary a spellcheck could've been used to first clean the data. I notice that profanity is being shown when you type `fu` the recommendations given are "`fuck`, `full`, `fucking`". I don't think you should spend too much time on the NLP, especially as you're learning, when really the core of this project, and your strength, is on building a keyboard. As long as the Language Model can easily be replaced and people can submit better implementations, that would be the most ideal.

patrickgold:
> In this case you would have to add both the capitalized and uncapitalized words -- it's the only way for the language model to be aware of both cases. The language model predicts based on the tokens it was trained on.

Ok that's what I thought too.

> For example for the english dictionary a spellcheck could've been used to first clean the data.

Currently, I validate English words with this regex: ```((\p{L}\p{M}*)|\'|\-)+```, which cleans a lot of garbage like numbers, emojis, words which have some other random special character in them which is definitely not a part of a word, etc. Adding a spellchecker to the Python combined-list-to-Flictionary converter shouldn't be an issue, but is currently not my main concern.

> I notice that profanity is being shown when you type fu the recommendations given are "fuck, full, fucking"

That's because the source I am using defines normal frequencies for profanity words, which the prediction algorithm then considers as a normal word. I could use one of the many profanity-lists out there to set the frequencies manually to 0 for these words before building the dictionary.

> As long as the Language Model can easily be replaced and people can submit better implementations, that would be the most ideal.

Yes, the LanguageModel data is freely interchangeable and the prediction code will work with that, so it is very easy to replace the downloaded dictionary in a later stage.

sabzo:
> Adding a spellchecker to the Python combined-list-to-Flictionary converter shouldn't be an issue, but is currently not my main concern.

Agreed. 

> Yes, the LanguageModel data is freely interchangeable and the prediction code will work with that, so it is very easy to replace the downloaded runtime dictionary in a later stage.

+1

> If you want to do a code review, I would really appreciate this, but I think it makes more sense to let me clean up and improve the search logic first, because else the list of things which are incorrect/should be improved is quite long :)

Is the code now good to look at?

patrickgold:
Above three commits further improve the performance and adds a distance measure to the prediction algorithm. Additionally, the codebase has been expanded to filter out profanity words (as long as they are correctly marked with f=0 in the Flictionary).

> Is the code now good to look at?

If you want, you can! I've implemented everything unigram-related and which is required to generate basic predictions. Bigrams etc. should work in theory but that is currently not really tested, as I need this only when implementing the next-word algorithm (currently only the current-word algorithm is implemented and working).

As for the accuracy of the suggestions, I am impressed that I made it that far but comparing it to other keyboards it lacks in quality. I think this is a mixture of both my edit distance implementation and the wordlist source I use. I will definitely experiment further in the next few days and see where I can further improve the code and maybe add the profanity filter to the Python library. If everything goes according to plan, I will complete phase 1 this Sunday and release v0.3.8 (where suggestions will be available as an Experimental+Alpha feature for everyone to test out).

sabzo:
@patrickgold Awesome! the speed/and suggestions for auto-complete are definitely better! So next word prediction is phase 2?

sabzo:
@patrickgold I checked out the branch and I noticed when you type really fast just some random characters there's a lag and the keyboard itself seems to freeze up. 

Also when you type a few words and try to backspace really fast/quickly by repeatedly tapping, or by pressing the backspace button the keyboard freezes/crashes (keyboard disappears) and then reopens a few seconds later.

Have you noticed this?

patrickgold:
@sabzo Yes, trying it out on an older device, this issue is more than notable. Above commit fixes a huge memory leak + improves memory management in the prediction algorithms, which has greatly improved the performance on my older device. DOes this improvement also occur on your device?

Issues I still notice are:
- The dictionary loading time when opening the keyboard for the first time is a bit slow on older devices
- Sometimes the Garbage Collector blocks the main thread and the keyboard freezes for a sec, but it happens a lot less and doesn't crash FlorisBoard anymore (at least on my testing device)

app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/Flictionary.kt
```diff
@@ -271,7 +271,7 @@ class Flictionary private constructor(
                     _freq = -1
                 ),
                 maxEditDistance = 3,
```
sabzo, line 275: not that it matters but most likely spellings are 2 edit distance -- maybe setting to 2 may reduce memory need?

patrickgold: Difficult to measure exactly, as the timer is not exact, but I just tested it out and it feels a tad smoother than with 3. Will conduct further experiments tomorrow, but personally I have no problem going with 2 instead of 3.

patrickgold:
I've just added a new version of the precompiled English (US) dictionary which has been checked against a very basic profanity wordlist and thus marks more than 4000 words as profanity which are only suggested if set so in the preferences.

I'll go off for now, so I will respond to further comments tomorrow!

patrickgold:
@sabzo I just wanted to let you know that I'll merge this branch in, as feature wise it is complete (for what I've planned in Phase 1) and I have to fix some other non-suggestion issues which occurred in the last two weeks. If any fixes for suggestions should come up, they can of course be merged in via a separate PR.

I aim to release v0.3.8 on Sunday evening, so users can try out the experimental suggestions stuff. Thanks again for your help thus far, I really appreciate it!

PR #331, opened 2021-02-09
Add exception for bat files to gitattributes

serebit:
This prevents `gradlew.bat` from being reset to LF line endings, which aren't guaranteed to be interpreted correctly by Windows CMD.


patrickgold:
Thanks for your PR! I haven't had a problem with this since I started the development, but knowing Windows and its ~love~ hate for LF newlines, I agree that this is a good exception for the .gitattributes file. Merging into master now!

PR #338, opened 2021-02-12
Add options for key spacing

### LINKED ISSUE #67: Customize keys/rows spacing
Having an option to customize (slightly to avoid errors) the spacing of both the keys and rows may be something interesting.
This option can be implemented in two ways (I think), literally pushing away the keys and rows a bit or reducing the size of the top and bottom edges of the keys.

### LINKED ISSUE #354: Normal Numpad Numbers height is miniscule
Split from https://github.com/florisboard/florisboard/issues/341 (4)

#### Short description
The numbers on numpad are way too small, and doesnt fit. Feels weird using it. And the dialer numpad (of florisboard) is completely different too, i thought they were same thing. (the bottom one - not to scale). I am mentioning dialer numpad in the hope that maybe it can suggest something on how can normal numpad and its layout be improved.
![1photo_2021-02-12_23-30-49]
![q6HAggbrop]

[1photo_2021-02-12_23-30-49]: https://user-images.githubusercontent.com/19423063/107851466-2d27ca00-6e30-11eb-898f-d6bc42ad4b5b.jpg
[q6HAggbrop]: https://user-images.githubusercontent.com/19423063/107852192-66af0400-6e35-11eb-98f5-421ff648fba6.png

dialer numpad = the numpad which shows when saving a contact, dont know if that's its proper name.

#### Steps to reproduce
-  open keyboard > symbols view <kbd>?123</kbd> > numpad view <kbd>1234</kbd>

#### Environment information
- FlorisBoard Version: 0.3.5
- Install Source: F-Droid
- Device: Jivi Prime P30
- Android version, ROM: 7.0 (seems AOSP)

X-yl:
Hello :) 

I wanted to implement proposal #67 because I kinda wanted the feature as well. So, I've added options to to adjust the key spacing like this:

I chose 0.5 dp as the unit because unfortunately a full dp is a bit big.

As an example,

Before (spacing 2dp, 5dp):  

After (spacing 4dp, 4dp):

(There's a bit more space between the columns, it looks a bit more even)

This is my first contribution here so let me know if I need to change something :)

---
Closes #67 
Closes #354

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
@@ -300,9 +300,11 @@ class TextInputManager private constructor() : CoroutineScope by MainScope(),
 
     override fun onSubtypeChanged(newSubtype: Subtype) {
         launch {
-            val keyboardView = keyboardViews[KeyboardMode.CHARACTERS]
-            keyboardView?.computedLayout = layoutManager.fetchComputedLayoutAsync(KeyboardMode.CHARACTERS, newSubtype, florisboard.prefs).await()
-            keyboardView?.updateVisibility()
+            // Seems like because the key spacing can change, all layouts need to be refreshed
+            for ((keyboardMode, keyboardView) in keyboardViews) {
+                keyboardView?.computedLayout = layoutManager.fetchComputedLayoutAsync(keyboardMode, newSubtype, florisboard.prefs).await()
+                keyboardView?.updateVisibility()
+            }
         }
     }
 
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
@@ -300,11 +300,9 @@ class TextInputManager private constructor() : CoroutineScope by MainScope(),
 
     override fun onSubtypeChanged(newSubtype: Subtype) {
         launch {
-            // Seems like because the key spacing can change, all layouts need to be refreshed
-            for ((keyboardMode, keyboardView) in keyboardViews) {
-                keyboardView?.computedLayout = layoutManager.fetchComputedLayoutAsync(keyboardMode, newSubtype, florisboard.prefs).await()
-                keyboardView?.updateVisibility()
-            }
+            val keyboardView = keyboardViews[KeyboardMode.CHARACTERS]
+            keyboardView?.computedLayout = layoutManager.fetchComputedLayoutAsync(KeyboardMode.CHARACTERS, newSubtype, florisboard.prefs).await()
+            keyboardView?.updateVisibility()
         }
     }
 
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/core/PrefHelper.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/core/PrefHelper.kt
@@ -316,6 +316,8 @@ class PrefHelper(
             const val HEIGHT_FACTOR_CUSTOM =            "keyboard__height_factor_custom"
             const val HINTED_NUMBER_ROW_MODE =          "keyboard__hinted_number_row_mode"
             const val HINTED_SYMBOLS_MODE =             "keyboard__hinted_symbols_mode"
+            const val KEY_SPACING_HORIZONTAL =          "keyboard__key_spacing_horizontal"
+            const val KEY_SPACING_VERTICAL =            "keyboard__key_spacing_vertical"
             const val LANDSCAPE_INPUT_UI_MODE =         "keyboard__landscape_input_ui_mode"
             const val LONG_PRESS_DELAY =                "keyboard__long_press_delay"
             const val NUMBER_ROW =                      "keyboard__number_row"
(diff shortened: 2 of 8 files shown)


RickyM7:
I just tested and find some problems:

1. In my opinion the vertical spacing should not be done by increasing/decreasing the size of the keys but rather increasing/decreasing keyboard size, so the keys would continue to be the standard size but with more space.

2. I think the change values are very exaggerated (you will have no problem with vertical spacing if the first point is fixed, but in horizontal spacing the value should not go more than 10 half dp because it would not cause some of the visual problems I saw (such as symbols not being visible, etc.)).

X-yl:
I believe the height can be adjusted independently so you can easily make the keys remain the same size by changing the keyboard height. I don't think it's a good idea to lock the height to the spacing because it reduces customisability for no reason.

And yes, 20 is definitely a very extreme maximum, I wasn't sure what would be a good one and I didn't want to make it too restricted. I don't mind changing it to something lower or otherwise the maintainer can change it when merging. Although I think limiting it to 10 is a bit low since that was already the default vertical spacing.

patrickgold:
Thanks a lot for your PR!

I've reviewed your changes at first glance everything looks fine, though there are some issues which arose while testing (See comment I make to the troubling lines of code).

Addressing the issues RickyM7 brought up:
1. I think it is best to leave keyboard height and vertical spacing to allow for greater customization.
2. The maximum values are indeed a bit big, but if you are a user on a tablet, they might be necessary. So I think the 0..20 half dp range should be kept in.

Another issue I see is that the spacing option shouldn't affect the Smartbar keyboards, it should just use the default spacing defined in the dimension files.

app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
```diff
@@ -300,9 +300,11 @@ class TextInputManager private constructor() : CoroutineScope by MainScope(),
 
     override fun onSubtypeChanged(newSubtype: Subtype) {
         launch {
-            val keyboardView = keyboardViews[KeyboardMode.CHARACTERS]
-            keyboardView?.computedLayout = layoutManager.fetchComputedLayoutAsync(KeyboardMode.CHARACTERS, newSubtype, florisboard.prefs).await()
-            keyboardView?.updateVisibility()
+            // Seems like because the key spacing can change, all layouts need to be refreshed
```
patrickgold, line 303: Refreshing all layouts here is very bad, because this method is called pretty often, but with the intention that ONLY the KeyboardMode.CHARACTERS layout is updated. Doing a refresh of all layouts results in laggy performance, so I think the margin change shouldn't be applied here.

As the current configuration of KeyView updates the margin only on initialize, I see why you chose to refresh all layouts here. Still, IMO the margin of the KeyView should be applied in KeyView.onMeasure(), thus avoiding the need to rebuild all layouts.

X-yl: Ah got it, I'll fix that.

X-yl:
I'm not sure I get what you mean by Smartbar keyboards, it seems to work with the Smartbar on, do you mean something else?

patrickgold:
What I mean is that the `KeyboardView` and `KeyView` class is also used also in the Smartbar to display the clipboard cursor row. There, a greater margin in the vertical mode looks horrible, while looking super cool in the "normal" `KeyboardView` (characters, symbols, etc.). You can check if your `KeyboardView` instance is a Smartbar one by reading the `isSmartbarKeyboardView` property.

X-yl:
Ahhh I get you now, sure no problem I'll fix that.

RickyM7:
> I believe the height can be adjusted independently so you can easily make the keys remain the same size by changing the keyboard height. I don't think it's a good idea to lock the height to the spacing because it reduces customisability for no reason.
>
> 1. I think it is best to leave keyboard height and vertical spacing to allow for greater customization.

@patrickgold @X-yl It makes total sense, even I tested increasing the vertical spacing along with an increase in keyboard height and everything was just perfect.

> 2. The maximum values are indeed a bit big, but if you are a user on a tablet, they might be necessary. So I think the 0..20 half dp range should be kept in.

Really, and independent of any value, the user who will be in control and will choose the most appropriate value for your device.

X-yl:
Hey I think I found a bug?

https://github.com/florisboard/florisboard/blob/3ea620a22e36626a87e949c5ea969b4662c1177d/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt#L875-L876

These lines should have `measuredWidth` and `measuredHeight` shouldn't they? With `desiredHeight` and `desiredWidth`, it sometimes passes a negative value to `setTextSizeFor` which gets it stuck in an infinite loop.

patrickgold:
> Hey I think I found a bug?

Hmm could quite possibly be, because previously the margins were static, and now you can change it, which can cause negative numbers. Does using `measuredWidth` and `measuredHeight` give you good sizing results? If so, then go for it, else we have to find some other measure to use in there.

> edit: also could you give me a screenshot of where exactly changing the key margins affects the smartbar? I can't seem to reproduce it, it always looks the same to me, see for example:

I mean the clipboard cursor row, not the quick actions you have open. The clipboard cursor row is visible if you have suggestions disabled and the quick actions are closed. I can give you a screenshot later if needed, currently I don't have your branch on my local setup and I am experiencing with suggestion algorithm stuff.

X-yl:
>Does using measuredWidth and measuredHeight give you good sizing results? 

Yes, measuredWidth and measuredHeight work correctly. I'll go ahead and make that change then.

>The clipboard cursor row is visible if you have suggestions disabled and the quick actions are closed.

Okay, I see what you mean now, no need for a screenshot.

X-yl:
Okay I made the changes as requested, let me know if something needs to be changed :)

patrickgold:
Thanks a lot for the changes. I will re-review your changes and should everything check out merge it :)

patrickgold:
@X-yl Sorry that it took me so long to re-review your changes, the suggestion implementation stuff distracted me a bit. I've tested out your changes on my local setup and everything seems to check out fine. Will merge it into master now. Will be released this Sunday with v0.3.8. Thanks again for your contribution!

PR #394, opened 2021-02-23
Add popup for character "ž"

ronanlefichant:


patrickgold:
Thanks for your PR! While your fix technically works, the Ž is now visible on all languages as a popup, rather than only Croatian etc. which isn't intended that way. There's currently a bug in the popup merging code which prevents the symbols to correctly merge with the Ž on Dvorak, which when fixed adds the Ž automatically to the extended popups if defined in the corresponding extended_popups file, so this change is unnecessary.

patrickgold:
Fixed by 69c52c00f64d74803235418ef21a3e5704ac26c3

Issue #349, opened 2021-02-14
Setting to have Select All, Copy, Left Arrow, Right Arrow, Cut & Paste w/ Suggestions


trymeouteh:
Please have a setting were you can have the select all button, copy button, left arrow button, right arrow button, cut button and paste button just like when the setting "Display suggestions while you type" is disable in the smartbar but to have this row of buttons when you enable the "Display suggestions while you type" setting. Here is a few ways I vision this to be achieved

1. The smartbar has a arrow button ">" in a circle on the left side. You could add another arrow button "<" on the right side that will show/hide the select, copy, left arrow, right arrow, cut and paste buttons.

2. The smartbar has a arrow button ">" in a circle on the left side. The select, copy, left arrow, right arrow, cut and paste buttons can be added to the buttons in the left ">" button.

3. Have two smartbars, one stack on the other. One smartbar is the select, copy, left arrow, right arrow, cut and paste buttons and the other smartbar is the word suggestions.

Great keyboard app BTW! Thanks for putting it on F-Droid!

patrickgold:
Thanks for your suggestion!

I have already discussed this with others and we had the idea to make a swipe left/right action in the Smartbar, which would allow you to quickly switch between different Smartbar contents. The problem is that I have absolutely no space for an extra arrow button, but a swipe action is definitely possible.

> Have two smartbars, one stack on the other. One smartbar is the select, copy, left arrow, right arrow, cut and paste buttons and the other smartbar is the word suggestions.

Stacking two Smartbars on top of each other doesn't look that good imo

trymeouteh:
I think the swipe action will be best and the arrow will be good if you have room.

trymeouteh:
Amazing work! Cant wait for word suggestions!

PR #454, opened 2021-03-13
Added support for private clipboard and clipboard history

### LINKED ISSUE #60: Add multi copy paste (Clipboard history)
Like AnySoftKeyboard which makes a list of everything you copied and you can choose to paste anyone

### LINKED ISSUE #113: Private Clipboard (Clipboard manager)
I know that android has protected it's clipboard a lot better than in the past, though I think it would be very nice to have florisboard with it's own clipboard separate from the system. 

Sadly, not sure of any current open source projects that have this feature. I currently have an app that does this, would be cool to just be built into the keyboard. Maybe with a history of it's own like other issues reference.

### LINKED ISSUE #435: (Persistent) Clipboard History
Related to #424, I'd like to suggest the Clipboard History feature similar to what SwiftKey does.

- The keyboard will remember everything copied to clipboard in the last n units of time (open for suggestions),
- items copied previously can be pasted from the Clipboard (History?) screen
- items can be pinned in order not to get automatically forgotten after the time expiration period
- items can be removed from history manually by swiping them away from the list

This would help sort out the paste suggestion issues some might have as it will keep multiple items in cache and it will be easy to paste desired item.

Screenshots [here](https://github.com/florisboard/florisboard/issues/424#issuecomment-790465673).

X-yl:
Closes #60 
Closes #113 
Closes #435 

:)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,6 +1,8 @@
+
 plugins {
     id("com.android.application") version "4.1.2"
     kotlin("android") version "1.4.30"
+    kotlin("kapt") version "1.4.30"
 }
 
 android {
@@ -54,6 +56,7 @@ android {
     }
 }
 
+
 dependencies {
     implementation("androidx.appcompat", "appcompat", "1.2.0")
     implementation("androidx.core", "core-ktx", "1.3.2")
@@ -69,6 +72,8 @@ dependencies {
     implementation("com.jakewharton.timber", "timber", "4.7.1")
     implementation("com.michael-bull.kotlin-result", "kotlin-result", "1.1.10")
     implementation("com.nambimobile.widgets", "expandable-fab", "1.0.2")
+    implementation("androidx.room", "room-runtime", "2.2.6")
+    kapt("androidx.room", "room-compiler","2.2.6")
 
     testImplementation("junit", "junit", "4.13.1")
     testImplementation("org.mockito", "mockito-inline", "3.7.7")
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -20,6 +20,11 @@
     <!-- Permission needed to vibrate if the user has key press vibration enabled -->
     <uses-permission android:name="android.permission.VIBRATE"/>
 
+    <!-- Permission to access the clipboard. Never granted to anything. -->
+    <permission
+        android:name="dev.patrickgold.florisboard.ACCESS_CLIPBOARD_PROVIDER"
+        android:protectionLevel="dangerous"/>
+
     <application
         android:name=".ime.core.FlorisApplication"
         android:allowBackup="false"
@@ -110,6 +115,15 @@
             android:icon="@mipmap/ic_launcher"
             android:label="@string/crash_dialog__title"
             android:theme="@style/CrashDialogTheme"/>
+
+        <provider
+            android:name="dev.patrickgold.florisboard.ime.clip.provider.FlorisContentProvider"
+            android:authorities="dev.patrickgold.florisboard.provider.clip"
+            android:grantUriPermissions="true"
+            android:permission="dev.patrickgold.florisboard.ACCESS_CLIPBOARD_PROVIDER"
+            android:exported="true">
+        </provider>
+
 
     </application>
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -20,11 +20,6 @@
     <!-- Permission needed to vibrate if the user has key press vibration enabled -->
     <uses-permission android:name="android.permission.VIBRATE"/>
 
-    <!-- Permission to access the clipboard. Never granted to anything. -->
-    <permission
-        android:name="dev.patrickgold.florisboard.ACCESS_CLIPBOARD_PROVIDER"
-        android:protectionLevel="dangerous"/>
-
     <application
         android:name=".ime.core.FlorisApplication"
         android:allowBackup="false"
@@ -120,8 +115,7 @@
             android:name="dev.patrickgold.florisboard.ime.clip.provider.FlorisContentProvider"
             android:authorities="dev.patrickgold.florisboard.provider.clip"
             android:grantUriPermissions="true"
-            android:permission="dev.patrickgold.florisboard.ACCESS_CLIPBOARD_PROVIDER"
-            android:exported="true">
+            android:exported="false">
         </provider>
 
 
--- a/app/src/main/assets/ime/text/characters/extended_popups/$default.json
+++ b/app/src/main/assets/ime/text/characters/extended_popups/$default.json
@@ -7,7 +7,8 @@
       "~enter": {
         "main": { "code": -213, "label": "switch_to_media_context", "type": "system_gui" },
         "relevant": [
-          { "code": -216, "label": "toggle_one_handed_mode_right", "type": "system_gui" }
+          { "code": -216, "label": "toggle_one_handed_mode_right", "type": "system_gui" },
+          { "code": -214, "label": "switch_to_clipboard_context", "type": "system_gui"}
         ]
       },
       "~left": {
--- a/app/src/main/assets/ime/text/extension/clipboard_cursor_row.json
+++ b/app/src/main/assets/ime/text/extension/clipboard_cursor_row.json
@@ -10,7 +10,8 @@
       { "code":  -20, "label": "arrow_left", "type": "navigation" },
       { "code":  -21, "label": "arrow_right", "type": "navigation" },
       { "code": -131, "label": "clipboard_cut", "type": "enter_editing" },
-      { "code": -132, "label": "clipboard_paste", "type": "enter_editing" }
+      { "code": -132, "label": "clipboard_paste", "type": "enter_editing" },
+      { "code": -214, "label": "switch_to_clipboard_context", "type":  "system_gui"}
     ]
   ]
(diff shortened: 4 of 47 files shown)


tsiflimagas:
Only problem I can see is that having both sync options enabled causes the keyboard to crash. Here is the log, if useful https://del.dog/mobognecre

X-yl:
> Only problem I can see is that having both sync options enabled causes the keyboard to crash. 
Thanks, that's a pretty big bug lol, I'll fix it ASAP.

X-yl:
I think I fixed it but honestly I really don't know how.

tsiflimagas:
Yes, it's fixed! If it works, it works I guess 😅 One tiny thing is the back arrow being kinda too small and decentered. Sorry for nitpicking😅

X-yl:
> Yes, it's fixed! If it works, it works I guess :sweat_smile: One tiny thing is the back arrow being kinda too small and decentered. Sorry for nitpicking :sweat_smile:

Alright, fixed! Nitpicking is always welcome :smile:

D3SOX:
Works really well, even pasting copied images into Discord. Thank you!

The only thing I noticed is that when you have suggestions enabled you can't access it but this is another unrelated issue. I think there should be a way to access those buttons
Screenshot_20210313-152021
with suggestions enabled.

tsiflimagas:
I catched another crash. Steps to reproduce: 1)create two notes, one being image, 2)tap on the non image note first (just tap), then tap on the image (also just tap). Log https://del.dog/buxexecait.txt
Another thing is that first note to be created can't be an image, you have to create a text one first (that's why second step in like that 😅).
Lastly, killing the keyboard process empties the clipboard content (idk if that's intended).
I'm not sure if these are to be discussed here or in issues after this is merged, so forgive me if it's the second case😅

X-yl:
> I catched another crash. Steps to reproduce: 1)create two notes, one being image, 2)tap on the non image note first (just tap), then tap on the image (also just tap). Log https://del.dog/buxexecait.txt

Yep, I can reproduce. I'll take a look and fix it. I honestly have no idea why it's throwing a SecurityException, since it definitely accesses the URI to display the image..

> Another thing is that first note to be created can't be an image, you have to create a text one first (that's why second step in like that 😅).

Hmm yeah, that's a bug too.

> Lastly, killing the keyboard process empties the clipboard content (idk if that's intended).

Hmm, I think that's fine for just the clipboard history, but yeah the pins should definitely stay even if the app is restarted. Will add.

> I'm not sure if these are to be discussed here or in issues after this is merged, so forgive me if it's the second case😅

Nope, this is the right place. So I can iron out all the bugs before it gets merged.

X-yl:
> I catched another crash. Steps to reproduce: 1)create two notes, one being image, 2)tap on the non image note first (just tap), then tap on the image (also just tap). Log https://del.dog/buxexecait.txt

Hmm alright, so I've been playing around and it seems that basically the permission to access the URI copied onto the clipboard is temporary, and it expires once the item is no longer on the system clipboard. That's why I didn't notice it in testing, I was using the default options which don't overwrite the system clipboard. Gboard gets around this by shipping their own ContentProvider which keeps the image after the URI expires.

This is quite a significant issue, so I think I'll close this PR for now, get it working properly and then reopen it later. Thanks for your patience.

X-yl:
Alright, I think I've made the necessary changes and fixes. There are still bound to be bugs, so let me know if you catch any.

tsiflimagas:
It seems that previously mentioned issues are fixed! Only thing I noticed is that pasting images in some apps doesn't work. For example, I could paste images in messenger just fine, but not in telegram (I get a pop-up that it's not supported, though I can with gboard). Maybe someone else has tried more apps.
Also a suggestion, if I may. Perhaps the little window you get when longpressing on an item could be divided into buttons, instead of having the buttons only on the areas of the icons of the options (pretty tiny detail again).

X-yl:
> It seems that previously mentioned issues are fixed! Only thing I noticed is that pasting images in some apps doesn't work. For example, I could paste images in messenger just fine, but not in telegram (I get a pop-up that it's not supported, though I can with gboard). Maybe someone else has tried more apps.

So good news and bad news. Good news is that this was simple mistake in my mimetype comparison, which I have fixed. Bad news is that now Telegram just ignores the `commitContent`. I have no idea how to fix this, since there's no error or anything. Ideas are welcome.

> Also a suggestion, if I may. Perhaps the little window you get when longpressing on an item could be divided into buttons, instead of having the buttons only on the areas of the icons of the options (pretty tiny detail again).

Ah I get what you mean. Can you try the latest commit and see if it's better?

tsiflimagas:
@X-yl yes, buttons work great, that's what I expected! As for image pasting, I don't know if there are any other apps which pose the same issue, maybe it happens only on telegram(?). Sadly I'm not a dev, otherwise I would be glad if I had any idea. Maybe Patrick has some thoughts on this.

X-yl:
Alright I grabbed the Telegram source, compiled it and had a look at the logs, it's mixed news again. Good news is that I figured out what the issue is. Bad news is that it's another permission issue (uh-oh). Luckily I think the fix should be just granting the permission somewhere. I'll investigate.

MisterY:
> Bad news is that now Telegram just ignores the `commitContent`. I have no idea how to fix this, since there's no error or anything. Ideas are welcome.

Would it be worth it to have an option to use the real clipboard as a medium for pasting? 
Just an idea, since I haven't tried this yet.

X-yl:
So, after much investigation, I've found the culprit. For some reason, instead of using the `ContentResolver.openInputStream` method, Telegram has decided to basically implement their own version (why would you do that??), which directly asks for the path to the image (using a column that only the internal `openFileHelper()` is supposed to use..) and tries to open it directly. Of course, Telegram is not permitted to access the file directly, and thus the permission error.  So anyway, I've stopped using `openFileHelper)_` (because that's what gboard does) and instead just implemented `openFile()` myself.

Alright, end of rant. I think it's fixed, I'm pushing the commits now.

> Would it be worth it to have an option to use the real clipboard as a medium for pasting?
> Just an idea, since I haven't tried this yet.

That would work in principle, however it's a bit hacky and also we want to try and isolate the floris clipboard if sync is off.

tsiflimagas:
It works great, sweet job!

patrickgold:
Hey @X-yl 

thanks a lot for your clipboard manager implementation, I really appreciate it! I've tried out the clipboard manager in several states during your development and must say it has improved a lot and looks very clean now! I've run the latest code on my testing device and it works really well. There are still bugs though I've encountered which I want to share with you:

- Pasting images works in many apps (especially in Signal, WhatsApp etc.) but does not work in Google Docs (there's no "This app does not support pasting this content" toast or whatsoever, just nothing happens). Haven't found any other apps which don't support pasting images as of now.
- When I have lets say 10 items in the clipboard history and then I set the max item count to 5 in the preferences, this heavily confuses the updating. Sometimes the clip items are empty or get updated delayed. This is a minor bug though and can easily be fixed by the user by clearing the history.

I will test out your clipboard manager implementation in practical use until tomorrow and see if I encounter anything else. Additionally I will review the code again, but as far as I've seen it is very clean and fully merge-able! When everything checks out, I will merge your implementation stuff if you are ready and your changes will be included in the v0.3.10-beta02 release.

Thanks again for your contribution and also a big thanks to all contributors in this PR which helped chase away bugs! :+1:

X-yl:
I'm glad to help, it's about time there was a proper, fully-featured open source alternative to gboard :)

Regarding apps that don't accept pasting images, I've only noticed gmail doesn't work. However, as far as I can tell, it's not a bug. The gmail app does not set the supported mimetypes, which indicates it doesn't support the Commit Content API. Gboard does actually let you paste images though, so I suppose gmail is still using some legacy method or has direct integration with gboard.

I've just pushed a fix for the second bug, I think it should work better now. Unfortunately as far as I can tell there's no way to have like an `onPrefChangedListener`, so the items only clear out after something else is copied. 
  
And yeah, I'm not usually an android dev so if you find any issues or anti-patterns in the code, please do let me know so I can clean it up.

patrickgold:
I've reviewed your changes and everything checks out. Merging into master now. Will be included in today's beta release v0.3.10-beta02. Thanks again for your contribution!

hbarsaiyan:
Thank you clipboard history was a much needed feature :)

hbarsaiyan:
Btw clipboard manager is not showing up for me on [NekoX](https://github.com/NekoX-Dev/NekoX) (Telegram Client). I read above it is fixed for telegram so thought i should point this out. Anyways the clipboard manager works very well on other apps.

X-yl:
Hey @hbarsaiyan can you elaborate on what's not working? I tried it with NekoX and it seems to be working fine. If you mean that the suggestions are hiding the entire clipboard cursor row, you can also access the clipboard history by long pressing on enter or by setting up a gesture. You can also make the clipboaed cursor row reappear by selecting some text.

hbarsaiyan:
@X-yl Ahh sorry i got what is not working. When the text field is empty the clipboard options do not appear but if i select some text it appears.
Screenshot_20210320-005542_Bromite
Screenshot_20210320-005746_Bromite

X-yl:
I believe you basically want #349, however it's not really related to this pull request and feature

Unrelated, but cool wallpaper. I used to have the same one a while ago lol

Kreuger:
As someone who has been using Florisboard for a while now, I just wanted to say thank you for this. I love the app and having someone else to help with the coding is a great benefit.

PR #455, opened 2021-03-15
Add theme import/export / Fix theme editor jumping to top

### LINKED ISSUE #93: Create and share Custom themes
As [json files](https://github.com/florisboard/florisboard/blob/master/app/src/main/assets/ime/theme/floris_day.json) are already used to load default themes, I think it should be possible to let people export and load them for custom themes.

It will allow people to create and share custom themes, and can result in creating a community-based theme store.

### LINKED ISSUE #379: Theme editor going back to top
Hello, and thanks for this highly customizable keyboard !

In the theme editor, after changing an attributes, the list goes back to top, then you have to scroll all the way down to next parameter.
Would it be possible to keep the position of the list ?

patrickgold:
This PR adds the ability to import/export json theme files from/to the local file system by using the Android `GET_CONTENT` and ~`OPEN_DOCUMENT`~ `CREATE_DOCUMENT` actions. Note that the file selection UI is device/system dependent and can not be influenced by FlorisBoard. If the theme import/export functionality works well on all devices I will adopt this also to other areas of FlorisBoard, especially though dictionaries and layout json files.

Closes #93 

---

The theme editor now does not jump to the top when opening an attribute dialog.

Closes #379 

---

Other non-user relevant changes:

- Switched from [michaelbull/kotlin-result](https://github.com/michaelbull/kotlin-result) to the in-built Kotlin Result type
- Add internal `FlorisActivity`, which provides several useful utilities used across different activities

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,6 @@ to get more information on this topic.
   [Jared Rummler](https://github.com/jaredrummler)
 * [Timber](https://github.com/JakeWharton/timber) by
   [JakeWharton](https://github.com/JakeWharton)
-* [kotlin-result](https://github.com/michaelbull/kotlin-result) by
-  [Michael Bull](https://github.com/michaelbull)
 * [expandable-fab](https://github.com/nambicompany/expandable-fab) by
   [Nambi](https://github.com/nambicompany)
 
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -55,8 +55,10 @@ android {
 }
 
 dependencies {
+    implementation("androidx.activity", "activity-ktx", "1.2.1")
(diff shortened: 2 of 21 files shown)


tsiflimagas:
When exporting a theme, after system's file manager is opened, I get no option to save the file. I'm on android 1, if this helps.

patrickgold:
Yup I just discovered this issue too on the emulators which have stock Android 10/11, I am currently investigating and fixing this issue.

patrickgold:
The Issue should now be fixed, at least on all my testing devices the export UI now offered me to save a file.

tsiflimagas:
Yes, everything works good! Nice feature!

githubgirly:
How do I find the theme? It's not in my file manager

PR #486, opened 2021-03-22
Create gboard_night.json

Netscaping:
Since there is a Gboard Day theme I added the night version.


patrickgold:
Thanks for your PR! I've just reviewed your theme and it looks just like Gboard, nice work! I will merge in your theme when the build check has finished.

D3SOX:
I don't think these white buttons fit that well. Maybe this can be improved

Screenshot_20210322-152851_crDroid_Home_1

patrickgold:
There's currently no option in the theme engine to separately style the toggle button and quick actions, so this would have to be addressed in a separate PR regarding theme engine improvements.

PR #491, opened 2021-03-24
Fix for #481

### LINKED ISSUE #481: Clipboard manager related question
Look the video and tell me why florisboard is not pasting copied data?

https://user-images.githubusercontent.com/75420364/111910267-b2814880-8a86-11eb-92c5-72934793a1c2.mp4

I copied that in bromite incognito mode but i dont think this should be cause of this happening

X-yl:
Essentially I forgot to change the mimetype to plaintext when copying html, which causes #481

Somewhat unrelated, but I also switched to using a static array for the plaintext mimetype array. This way, every time a plaintext item is added, it doesn't allocate a brand new array.

Fixes #481


patrickgold:
Looks good to me, thanks for the fix!

PR #512, opened 2021-03-27
Add the Workman keyboard layout

icyphox:
This patch adds the Workman keyboard layout. Ref: https://workmanlayout.org

### DIFF
--- a/app/src/main/assets/ime/config.json
+++ b/app/src/main/assets/ime/config.json
@@ -30,7 +30,8 @@
     "swiss_french": "Swiss French (QWERTZ)",
     "swiss_italian": "Swiss Italian (QWERTZ)",
     "turkish_q": "Turkish-Q",
-    "turkish_f": "Turkish-F"
+    "turkish_f": "Turkish-F",
+    "workman": "Workman"
   },
   "defaultSubtypes": [
     {
--- a/app/src/main/assets/ime/text/characters/workman.json
+++ b/app/src/main/assets/ime/text/characters/workman.json
@@ -6,14 +6,14 @@
   "arrangement": [
     [
       { "code":  113, "label": "q" },
-      { "code":  119, "label": "d" },
-      { "code":  102, "label": "r" },
-      { "code":  112, "label": "w" },
-      { "code":  103, "label": "b" },
+      { "code":  100, "label": "d" },
+      { "code":  114, "label": "r" },
+      { "code":  119, "label": "w" },
+      { "code":   98, "label": "b" },
       { "code":  106, "label": "j" },
-      { "code":  108, "label": "f" },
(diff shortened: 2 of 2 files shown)


patrickgold:
Thanks for your PR! Unfortunately there is an issue with your layout preventing me to merge it into master: It seems to me that you copied the Colemak json file but only updated the `label` fields. More importantly though the `code` fields must be correctly set in order for FlorisBoard to send the correct letter to the input connection. The code is the standarised UTF-8 code of each letter, you can use https://unicode-table.com to find the code for each letter (Use the HTML code number after the `&#`).

icyphox:
Oh silly me, of course! Let me go ahead and do that.

kj7rrv:
How can I install the debug build? Android says "app not installed" when I try to install it.

patrickgold:
@icyphox I tried it out and except for one key code everything looks good. I've corrected it for you and will now merge it into master. Thanks for your contribution!

> How can I install the debug build? Android says "app not installed" when I try to install it.

@kj7rrv Could be because the debug builds don't increment the version codes, thus the Android installer rejects the APK. Uninstalling and reinstalling the debug version of FlorisBoard should fix this problem.

kj7rrv:
@patrickgold I only have the stable and beta right now, so that isn't the issue.On Mar 29, 2021 06:53, Patrick Goldinger ***@***.***> wrote:
@icyphox I tried it out and except for one key code everything looks good. I've corrected it for you and will now merge it into master. Thanks for your contribution!

How can I install the debug build? Android says "app not installed" when I try to install it.

@kj7rrv Could be because the debug builds don't increment the version codes, thus the Android installer rejects the APK. Uninstalling and reinstalling the debug version of FlorisBoard should fix this problem.

patrickgold:
> I only have the stable and beta right now, so that isn't the issue

Then your phone blocks installations of APKs which are unsigned (meaning my debug builds don't have a signature attached to them because they are built on GitHub's servers). The best thing to do is to wait for the next beta release, which is signed properly with my signature keys and thus is always install-able. Alternatively you could try to enable the developer mode and see if this allows you to install the debug build.

kj7rrv:
Okay, no problem. I do have developer mode on already.Is it possible for GitHub to sign APKs using a test key or something?On Mar 29, 2021 07:03, Anirudh Oppiliappan ***@***.***> wrote:

Knew there was a typo in there somewhere. Thanks a bunch, @patrickgold — great keyboard app, now with my layout of choice!

patrickgold:
@kj7rrv I just wanted to let you know that the "App not installed" message appears on your device because of #535 (basically the content provider namespace clashes, thus preventing the installation). Once that is fixed you should be able to install beta and debug builds in parallel again.

PR #527, opened 2021-03-29
Neo2 layout

ostrya:
Initial attempt at Neo2 / Bone layout

* For now, only layers 1, 2 and 3 are supported.
* Layer 2 is reachable via caps, apart from number row, comma and full stop (which I think are easier to use if not affected by caps). Instead, the relevant characters are added as popups.
* Layer 3 is set up as main popup, apart from number row, where
  a) due to existing number_row layout the main popup is already set
  b) the characters apart from the Euro sign are not used very often

The overall layout is kept as much as possible, with the following
exceptions:
* The number row contains only numbers to support the built-in number_row layout. The minus sign is moved to the lowest character row, while circumflex and grave accents are not included.
* To not overcrowd the layout and have the same number of keys for first and second row, the acute accent is not included as separate key but can be reached as additional popup to sharp s.
* The minus sign is put between m and j (or z and k respectively) where comma and full stop usually would be, because the backspace takes up too much space for both keys.
* Also, having comma and full stop on the same height with the space  key makes the layout more consistent with the existing layouts and the special usage as ~left and ~right keys.

Please note that I added a special `de_DE-neobone` locale variant:

The default de locale already defines a lot of extended popups which do not match the Neo2 / Bone layout logic. Adding a locale variant allows overriding those defaults. In addition, it allows adding custom popups for the number row. This especially allows including circumflex and grave accents as popups for 1 and 0 respectively.

bone
neo2

### DIFF
--- a/app/src/main/assets/ime/text/characters/bone.json
+++ b/app/src/main/assets/ime/text/characters/bone.json
@@ -1,0 +1,50 @@
+{
+  "type": "characters",
+  "name": "bone",
+  "label": "Bone",
+  "authors": [ "ostrya" ],
+  "direction": "ltr",
+  "modifier": "neo2",
+  "arrangement": [
+    [
+      { "code":  106, "label": "j" },
+      { "code":  100, "label": "d" },
+      { "code":  117, "label": "u" },
+      { "code":   97, "label": "a" },
+      { "code":  120, "label": "x" },
+      { "code":  112, "label": "p" },
+      { "code":  104, "label": "h" },
+      { "code":  108, "label": "l" },
+      { "code":  109, "label": "m" },
+      { "code":  119, "label": "w" },
+      { "code":  223, "label": "ß", "shift": { "code": 7838, "label": "ẞ" }, "popup": {
+        "relevant": [
+          { "code":  180, "label": "´" }
+        ]
+      } }
+    ],
+    [
+      { "code":   99, "label": "c" },
+      { "code":  116, "label": "t" },
+      { "code":  105, "label": "i" },
+      { "code":  101, "label": "e" },
+      { "code":  111, "label": "o" },
+      { "code":   98, "label": "b" },
+      { "code":  110, "label": "n" },
+      { "code":  114, "label": "r" },
+      { "code":  115, "label": "s" },
+      { "code":  103, "label": "g" },
+      { "code":  113, "label": "q" }
+    ],
+    [
+      { "code":  102, "label": "f" },
+      { "code":  118, "label": "v" },
+      { "code":  252, "label": "ü" },
+      { "code":  228, "label": "ä" },
+      { "code":  246, "label": "ö" },
+      { "code":  121, "label": "y" },
+      { "code":  122, "label": "z" },
+      { "code":  107, "label": "k" }
+    ]
+  ]
+}
--- a/app/src/main/assets/ime/text/characters/bone.json
+++ b/app/src/main/assets/ime/text/characters/bone.json
@@ -7,44 +7,55 @@
   "modifier": "neo2",
   "arrangement": [
     [
-      { "code":  106, "label": "j" },
-      { "code":  100, "label": "d" },
-      { "code":  117, "label": "u" },
-      { "code":   97, "label": "a" },
-      { "code":  120, "label": "x" },
-      { "code":  112, "label": "p" },
-      { "code":  104, "label": "h" },
-      { "code":  108, "label": "l" },
-      { "code":  109, "label": "m" },
-      { "code":  119, "label": "w" },
-      { "code":  223, "label": "ß", "shift": { "code": 7838, "label": "ẞ" }, "popup": {
-        "relevant": [
-          { "code":  180, "label": "´" }
-        ]
-      } }
+      { "$": "auto_text_key", "code":  106, "label": "j" },
+      { "$": "auto_text_key", "code":  100, "label": "d" },
+      { "$": "auto_text_key", "code":  117, "label": "u" },
+      { "$": "auto_text_key", "code":   97, "label": "a" },
+      { "$": "auto_text_key", "code":  120, "label": "x" },
+      { "$": "auto_text_key", "code":  112, "label": "p" },
+      { "$": "auto_text_key", "code":  104, "label": "h" },
+      { "$": "auto_text_key", "code":  108, "label": "l" },
+      { "$": "auto_text_key", "code":  109, "label": "m" },
+      { "$": "auto_text_key", "code":  119, "label": "w" },
+      { "$": "case_selector",
+        "lower": {
+          "code":  223, "label": "ß", "popup": {
+            "relevant": [
+              { "code":  180, "label": "´" }
+            ]
+          }
+        },
+        "upper": {
+          "code": 7838, "label": "ẞ", "popup": {
+            "relevant": [
+              { "code":  180, "label": "´" }
+            ]
+          }
+        }
+      }
     ],
     [
-      { "code":   99, "label": "c" },
-      { "code":  116, "label": "t" },
-      { "code":  105, "label": "i" },
-      { "code":  101, "label": "e" },
-      { "code":  111, "label": "o" },
-      { "code":   98, "label": "b" },
-      { "code":  110, "label": "n" },
-      { "code":  114, "label": "r" },
-      { "code":  115, "label": "s" },
-      { "code":  103, "label": "g" },
-      { "code":  113, "label": "q" }
+      { "$": "auto_text_key", "code":   99, "label": "c" },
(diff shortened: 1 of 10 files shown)


patrickgold:
Thanks for your PR, I really appreciate it! Currently I am experimenting with subtype specific symbol layouts, with this also the meta data syntax for all layouts slightly and the ime config file syntax completely changes. This will definitely cause a merge conflict here, so I will handle your PR and adjust the syntax once the subtype specific symbol feature is merged. I think this will either be on Thursday or Friday, depending on how well things are working with the improved layout definitions.

Regarding the custom locale: I see what you did there, but is the term `neobone` just used by you to add custom popups for neo2 or is this a somewhat standard?

ostrya:
> Currently I am experimenting with subtype specific symbol layouts

That's great! I already noticed the layouts would work much better if I could put the layer 3 into the symbol layer, because then I would also get hinting.

> Regarding the custom locale: I see what you did there, but is the term neobone just used by you to add custom popups for neo2

Yes, well, the locale variant `neobone` really is only an invented term by me. According to the Javadoc on Locale, a valid variant matches `[0-9][0-9a-zA-Z]{3} | [0-9a-zA-Z]{5,8}`. The underlying issue was that I did not want to have the extended popups defined by the `de.json` base file, but those files are chosen based on the locale, and I could neither invent a new language nor a new country, so I went for a user-defined variant.

patrickgold:
I've merged in #561 and subtype specific layouts, so the master branch is ready again to accept layout contributions. As for your ne02 layouts, you said you have custom symbol layouts, which can now be added by placing the layout file in `assets/ime/text/symbols/<layoutname.json>`.

For defining the default attribute, the syntax has slightly changed and now looks like this: https://github.com/florisboard/florisboard/blob/7b25381850b5e5e3ca3c4c99f469b560ecddc423/app/src/main/assets/ime/config.json#L532-L539. You could add a "symbols" line to the preferred object, so by default the neo2 symbols layout gets pre-selected in the subtype add dialog.

Last but not least the layout labels: They are not stored in the `characterLayouts` array in the IME config anymore, but are stored directly in each layout file, directly below "name" as "label" : "Your Custom Label".

If the above mentioned issues are fixed, I will happily merge your PR!

ostrya:
I have had a look at the new configurable symbol layout, and it nicely fits with the way the neo layouts works. But I have one issue: currently, the hints from the symbols layer are applied on the "wrong" rows.

Basically, my symbols layout has the same amount of rows as the character layout, and I would like the hint on the character layout key to be taken from the corresponding row and key in the symbol layout. But the way numeric hinting is built does not let me do this:

When numeric row is enabled and numeric hints are enabled:
* **is**: first character row has numeric row as hints
* **should be**: in my opinion, this combination has little benefit, as you already have the numeric row. But for consistency, I would expect the numeric hints to be shown in the numeric row, so that the 1:1 mapping of the layout works correctly.

When numeric row is enabled and numeric hints are disabled:
* **is**: first character row has no hints at all
* **should be**: first character row should use first symbol row as hints

When numeric row is disabled and numeric hints are enabled:
* **is**: ok as is

When numeric row is disabled and numeric hints are disabled:
* **is**: first character row has no hints at all
* **should be**: first character row should use first symbol row as hints

The **should be** cases are written for the case when the symbol layout has the same number of rows as the character layout. In other cases (e.g. character = qwerty (4 rows) vs. symbol = western (3 rows)), the merge could be done in a way that the last rows are aligned rather than the first rows.

Do you think this is something you can change (or that I should create a proposal PR for), or is the current behavior intended to stay this way? In that case, I will probably need to adapt my changes accordingly and probably remove the separate symbol layer again.

ostrya:
I also took the liberty to add a custom modifier layout for the symbol layer to both better match key positions from the character layer and be able to use double and single quote instead of comma and full stop and complete the 3rd neo2 layer.

patrickgold:
> Do you think this is something you can change

Yes, definitely! Will make a PR with the change once v0.3.10 is released (seems to be tomorrow as no critical crashes occurred).

ostrya:
Since this has kept me thinking through the night, I have already added a proposal to this PR. If you prefer, I can also create a separate PR. Or if you don't like it at all, I can also remove it from the PR and wait for your implementation :-)

patrickgold:
If possible, could you extract this to a separate PR? I've just reviewed the change and while it partially works (there's some issues in it), I have to also think of some other feature requests around this behavior which should get incorporated.

As for the real topic of this PR: I've reviewed the layout & code and in general it looks very good. I also saw that you added support for custom symbols modifiers, which is great! (wanted to do this quite for a time now but didn't ever need it, so I was too lazy to modify the layout manager haha).

ostrya:
As you proposed, I have moved the new layout merge logic into the new PR #606, so I guess this PR is now ready for final review. Though I believe that merging the other PR first would make the user experience on this layout better, as the symbol characters are currently not where one would expect them to be :-)

app/src/main/assets/ime/text/symbols/neo2.json
```diff
@@ -0,0 +1,46 @@
...
+      { "code":   63, "label": "?" },
+      { "code":   40, "label": "(" },
+      { "code":   41, "label": ")" },
+      { "code":   45, "label": "-" },
+      { "code":   58, "label": ":" },
+      { "code":   64, "label": "@" }
+    ],
+    [
+      { "code":   35, "label": "#" },
+      { "code":   36, "label": "$" },
```
patrickgold, line 37:
```suggestion
      { "code": -801, "label": "currency_slot_1" },
```

The user has the ability to set the currency set for the symbols layout, thus neo2 should also respect this. The real currency symbol will then automatically be inserted in runtime.

app/src/main/assets/ime/text/characters/bone.json
```diff
@@ -0,0 +1,50 @@
+{
+  "type": "characters",
+  "name": "bone",
+  "label": "Bone",
+  "authors": [ "ostrya" ],
+  "direction": "ltr",
+  "modifier": "neo2",
+  "arrangement": [
+    [
+      { "code":  106, "label": "j" },
```
patrickgold, line 10: All character letters which need to be auto-capitalized need to be set to "auto_text_key". E.g.:

```
{ "$": "auto_text_key", "code":  113, "label": "q" },
```

Keys which are always the same regardless of the caps state do not need the class identifier upfront, as well as characters which are used within a `case_selector`.

app/src/main/assets/ime/text/characters/bone.json
```diff
@@ -0,0 +1,50 @@
...
+      { "code":  104, "label": "h" },
+      { "code":  108, "label": "l" },
+      { "code":  109, "label": "m" },
+      { "code":  119, "label": "w" },
+      { "code":  223, "label": "ß", "shift": { "code": 7838, "label": "ẞ" }, "popup": {
+        "relevant": [
+          { "code":  180, "label": "´" }
+        ]
+      } }
+    ],
```
patrickgold, line 25: The `shift` attribute is not respected anymore, a `case_selector` must be used here. Example (copied from Colemak):

```json
      { "$": "case_selector",
        "lower": { "code":   59, "label": ";", "popup": {
          "relevant": [
            { "code":   58, "label": ":" }
          ]
        } },
        "upper": { "code":   58, "label": ":", "popup": {
          "relevant": [
            { "code":   59, "label": ";" }
          ]
        } }
      }
```

app/src/main/assets/ime/text/characters/neo2.json
```diff
@@ -0,0 +1,50 @@
...
+      { "code":  104, "label": "h" },
+      { "code":  103, "label": "g" },
+      { "code":  102, "label": "f" },
+      { "code":  113, "label": "q" },
+      { "code":  223, "label": "ß", "shift": { "code": 7838, "label": "ẞ" }, "popup": {
+        "relevant": [
+          { "code":  180, "label": "´" }
+        ]
+      } }
+    ],
```
patrickgold, line 25: Same here for the `shift` attribute.

patrickgold:
Above review-comments are some issues I found with the neo2 layouts (needed partly because of the layout rework in #734).

app/src/main/assets/ime/text/symbols/neo2.json
ostrya: Makes sense to use this feature since it's there. But from the layout point of view, I feel the Euro sign below the numerical 7 would be more suited. The $ sigh here is I think more meant for programming purposes.

app/src/main/assets/ime/text/characters/bone.json
ostrya: I see, I'll fix this.

ostrya:
I think I fixed all issues that you noticed. Please have a look at the changes :-)

PR #537, opened 2021-03-30
Italian translation error

Mek101:


patrickgold:
Thanks for your PR. This is a localization-related error and translations are only managed on [Crowdin](https://crowdin.com/project/florisboard), as stated in the contributing guidelines. There, you can edit the string and in the next build cycle it will be pushed into the app source code.

PR #560, opened 2021-04-02
Add IPA keyboard layout

### LINKED ISSUE #558: Add IPA keyboard
[International Phonetic Alphabet](https://en.wikipedia.org/wiki/International_Phonetic_Alphabet) (IPA) is a way to represent sounds in languages. As I am often in linguistic discussions, I find it helpful to have a keyboard for it.

I suggest not following Gboard's example, whose layout spreads over four pages with all vowels on the second page and thus unpleasant to use. I have two suggestions for implementation:

- Use QWERTY layout, make the input based on [X-SAMPA](https://en.wikipedia.org/wiki/X-SAMPA) (sorry I don't have a visual for it)
  - the default key is like normal QWERTY
  - after pressing Shift, each key returns the equivalent IPA symbols encoded by X-SAMPA
    - e.g. `Shift` + `s` → x/S/ → ʃ
  - All keys' variants can be accessed by holding the key (e.g. hold s --> show x/S/ ʃ, x/s\/ ɕ, x/s`/ ʂ)
  - Symbols encoded by non-alphanumeric characters in X-SAMPA (which usually don't represent consonants nor vowels) are in the special character input.
- Use X11 keyboard layout, like in the below picture. In each key:
  - the bottom left is the default key
  - the top left is the key after pressing Shift
  - the other keys can be accessed by holding the key
  - characters that don't look like a letter (which usually don't represent consonants nor vowels) can be put in the special character input.

X11 layout

Huy-Ngo:
In this PR I want to resolve #558. I did not implement the scheme I proposed, because I realized neither are intuitive (for example, it is unintuitive to find ɱ at the extension of f rather than m).

Not all symbols are mapped yet, work in progress. Characters denoting length, tones, stress,... are not mapped.

### DIFF
--- a/app/src/main/assets/ime/config.json
+++ b/app/src/main/assets/ime/config.json
@@ -560,6 +560,16 @@
       "preferred": {
         "characters": "qwerty"
       }
+    },
+    {
+      "id": 2601,
+      "languageTag": "IPA-IPA",
+      "currencySet": "dollar",
+      "preferred": {
+        "characters": "ipa",
+        "symbols": "ipa",
+        "symbols2": "ipa"
+      }
     }
   ]
 }
--- a/app/src/main/assets/ime/text/characters/ipa.json
+++ b/app/src/main/assets/ime/text/characters/ipa.json
@@ -110,7 +110,7 @@
         "code": 112, "label": "p",
         "popup": {
           "relevant": [
-            { "code": 632, "label": "ɸ" }
+            { "code": 632, "label": "ɸ" },
             { "code": 651, "label": "ʋ" }
           ]
         },
--- a/app/src/main/assets/ime/text/characters/ipa.json
+++ b/app/src/main/assets/ime/text/characters/ipa.json
@@ -1,0 +1,301 @@
+{
+  "type": "characters",
+  "name": "ipa",
+  "authors": [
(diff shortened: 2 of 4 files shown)


Huy-Ngo:
@patrickgold Could you review this PR, please?

By the way, I noticed the labels on the keys don't match the labels defined in the config files. Is this intentional?

patrickgold:
I've reviewed and tested the layout and it looks really good! For me it is ready to be merged, let me know if this is also ok for you or if you still want to change something.

> By the way, I noticed the labels on the keys don't match the labels defined in the config files. Is this intentional?

Do you mean the labels of the keys which are basically modifiers to the previous letters and should have that blank dotted circle on the key? If so, yes, currently FlorisBoard is unable to display those beautifully and just displays the modifiers, leaving most of the key thus blank.

Huy-Ngo:
Thanks. Please merge it then.

patrickgold:
@Huy-Ngo above linked PR adds support for the combining diacritical marks, so your IPA key labels can be displayed properly. I did the same as many other websites and keyboards do, adding a dotted circle as the base for the label which should represent the base character to which it gets applied.

Huy-Ngo:
Thanks. That looks much better!

PR #580, opened 2021-04-07
Improve Persian popup (آ)

UltimateOmega:
With this rearrange, when Accent is prioritized in hinted symbols, آ is the first option in the popup.
آ is one of the most used alphabets in Persian language.


patrickgold:
Thanks for your contribution! Merging into master.

PR #600, opened 2021-04-09
Make version more generic so that it doesn't seem out of date.

Glitchy-Tozier:
I'm torn between a big `x` and a small one. Small is less in-your-face, which is nice. Big has the same size as numbers.

Not sure is this is the correct way to create pull requests so please shout at me angrily if something went wrong.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -25,7 +25,7 @@ Describe the bug in a short but concise way.
 4. See error
 
 #### Environment information
-- FlorisBoard Version: <!-- e.g. 0.3.6 -->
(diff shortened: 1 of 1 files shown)


patrickgold:
Thanks for your PR! This is absolutely the correct way to make PRs, you did everything right :)

Regarding the `x`/`X` decision: `X` has been used by me in many issues with generic version numbers (because the `v` in the front (which isn't in the bug report template) is already in lowercase). As the point of this change is to make the template version field more generic, I'd suggest `<!-- 0.X.X -->`, so we don't have to update it again for the v0.4.0 release.

Glitchy-Tozier:
There you go! It was quite a bit of work, but I'd say I did a pretty good job.  :)

patrickgold:
:smile:

PR #606, opened 2021-04-10
Merge hints more flexibly

### LINKED ISSUE #618: multiple symbols on long press
could you add multiple symbols to the long-press popup for certain keys? most consonants only have one key that pops up on the long press when there is plenty of space for more keys. heres what it might look like
florisboardproposal

ostrya:
As discussed, I extracted my proposal for adding hints to the character layout in a more general fashion into a separate PR.

To allow symbol layouts with the same or more rows as the character
layout to be hinted more consistently, the hinting of the numeric row
is split from the rest of the symbol layout.

The numeric row hinting is always done in the first row. If the numeric
row is actually enabled, no additional hints will be shown (as hints
are only added to character type keys).

The symbol hinting is now bottom-aligned: hints from the last symbol
row are shown in the last character row.

Overall, only keys matching the expected type (numeric for numeric
row, character elsewhere) are iterated over. This allows also the
bottom row with the layout switch keys to be hinted.

Closes #618

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKey.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKey.kt
@@ -26,7 +26,8 @@ class TextKey(override val data: KeyData) : Key(data) {
     var computedData: TextKeyData = TextKeyData.UNSPECIFIED
         private set
     val computedPopups: MutablePopupSet<TextKeyData> = MutablePopupSet()
-    var computedHint: TextKeyData? = null
+    var computedSymbolHint: TextKeyData? = null
+    var computedNumericHint: TextKeyData? = null
 
     fun compute(evaluator: TextComputingEvaluator) {
         val keyboardMode = evaluator.getKeyboard().mode
@@ -44,7 +45,8 @@ class TextKey(override val data: KeyData) : Key(data) {
         } else {
             computedData = computed
             computedPopups.clear()
-            computedPopups.hint = computedHint?.computeTextKeyData(evaluator)
+            computedPopups.hint = computedSymbolHint?.computeTextKeyData(evaluator)
+            computedPopups.merge(PopupSet(hint = computedNumericHint?.computeTextKeyData(evaluator)))
             if (computed is BasicTextKeyData && computed.popup != null) {
                 computedPopups.merge(computed.popup, evaluator)
             }
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKey.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKey.kt
@@ -27,7 +27,7 @@ class TextKey(override val data: KeyData) : Key(data) {
         private set
     val computedPopups: MutablePopupSet<TextKeyData> = MutablePopupSet()
     var computedSymbolHint: TextKeyData? = null
-    var computedNumericHint: TextKeyData? = null
+    var computedNumberHint: TextKeyData? = null
 
     fun compute(evaluator: TextComputingEvaluator) {
         val keyboardMode = evaluator.getKeyboard().mode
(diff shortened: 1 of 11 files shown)


ostrya:
@patrickgold did you already have a chance to have a look at this and maybe point out the issues that I missed?

patrickgold:
@ostrya Nope sry, the OOM issue is currently fully taking away my time. As I am completely reworking the layout and keyboard view part though currently, I will see if I don't include it directly into my PR, as the merging logic has also changed quite a bit, will inform you on what I decide!

patrickgold:
@ostrya #734 is now merged in. I've temporarily settled to re-use the old key hint merge method, but I am now open to reworking the key hint merge logic. Could you adjust the changes to resolve the merge conflict, so I can have a detailed look at your practical implementation? This would be incredibly helpful, thanks in advance :)

ostrya:
Hi @patrickgold , I have rebased my change. Since the way the key's popups are generated has been changed, I have now also had to touch the computation logic in TextKey to introduce two different hints: a numeric hint, and a "symbol" hint (name is open for discussion).

This allows for the (Neo2) case where the symbol layout has the same number of character rows, and thus the numeric hint line conflicts with the symbol hint line.

For now, I chose to give the symbol hint preference, but it might also be possible to make this configurable.

patrickgold:
Thanks for adapting the changes, really appreciate it! I've tried your changes out and everything works like it should. For testing purposes I also merged in your neo2 layouts in the local branch to test how they react with each other and it works really well. For the merging part, I have only seen one issue, I will suggest a change for it so you can review it.

> For now, I chose to give the symbol hint preference, but it might also be possible to make this configurable.

I think leaving it as symbols-prioritized is the best, because most neo2-aylout users will have the numeric row in characters enabled anyways.

app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKey.kt
```diff
@@ -44,7 +45,8 @@ class TextKey(override val data: KeyData) : Key(data) {
         } else {
             computedData = computed
             computedPopups.clear()
-            computedPopups.hint = computedHint?.computeTextKeyData(evaluator)
+            computedPopups.hint = computedSymbolHint?.computeTextKeyData(evaluator)
+            computedPopups.merge(PopupSet(hint = computedNumericHint?.computeTextKeyData(evaluator)))
```
patrickgold, line 49:
```suggestion
            val symbolHint = computedSymbolHint
            val numericHint = computedNumericHint
            if (symbolHint != null && numericHint != null) {
                computedPopups.hint = symbolHint.computeTextKeyData(evaluator)
                numericHint.computeTextKeyData(evaluator)?.let { computedPopups.relevant.add(it) }
            } else if (symbolHint != null) {
                computedPopups.hint = symbolHint.computeTextKeyData(evaluator)
            } else if (numericHint != null) {
                computedPopups.hint = numericHint.computeTextKeyData(evaluator)
            }
```

Merging the hints this way avoids the need to temporarily allocate a new `PopupSet` just for merging in the numeric hint.

ostrya:
Since the topic with currency symbols came up in the neo2 PR, I was wondering - should we maybe also merge the popups of the symbol keys to the character keys?

E.g. if we have a character key "x" with no original popups, the numeric hint "1" with popups "°" and "¹", and the symbol hint "…" with no further popups, we could merge it to the "x" so that it has hint "…", and popups "…", "1", "°" and "¹". What do you think?

patrickgold:
> Since the topic with currency symbols came up in the neo2 PR, I was wondering - should we maybe also merge the popups of the symbol keys to the character keys?

This is actually an accepted proposal somewhere in the mess of issues... ah, found it: #618 . If you want, you can add it to the merging part, as long as it is toggle-able in the preferences (and should default to disabled if not otherwise configured by the user).

ostrya:
@patrickgold I have added the feature as I think it was intended in the proposal.  Please have a look, also regarding the new setting and the way it is made available for the final calculation.

patrickgold:
Thanks for fixing the mentioned issues! Still, there are some issues I experienced with the order the hint keys appear in the popup. Especially with the "Accent is prioritized", where the hint is sometimes prioritized over the main character key. Additionally, when merging the hint popups, the symbol shown on the key does not match up with the first selected popup when a character key only has hint symbol keys (if the symbol shown is a hint but the first selected is a accent that's ok when the hint pref is set to "Accent is prioritized" or "Smart prioritization"). Could you have a look over it and try to improve the order the popups are arranged when several sites merge in their keys? Thanks in advance!

ostrya:
Sure, I will have a look.

patrickgold:
Thanks for reworking the popup merging method! I've reviewed the changes and I am impressed, the popup order makes much more sense now. I've tested various different combis of hint pref settings and symbol popup merging and in every case it worked as it should.

From my side this feature is now ready to be merged, together with neo2. I am just waiting for your approval that everything is ready from your side.

ostrya:
Yes, the PR is ready from my side :-)

PR #608, opened 2021-04-10
Fix glide logic in TextInputManager

### LINKED ISSUE #581: Space typed word after glide
Whenever you glide a word and manually type something immediately after, it won't be separated. 
This is okay if someone wants to add something to the previous word, like typing "as" after gliding "where" to form "whereas", but in other use cases it would be convinient to have that space.
Is this intended behaviour? In such case, maybe making it optional could help. 

For an example of what this means, Google implemented it into its keyboard.

### LINKED ISSUE #595: Cancel Word-Deletion Behaviour
@X-yl Adding you because you seem to be the expert in regards to Swipe-typing

You have added a very smart thing: If the user presses `delete`, after swiping, the whole word will be deleted.

HOWEVER this behavior isn't cancelled when the user starts typing normally. This means that this...
1. Swipe
2. type multiple word
3. try to delete a character
4. accidenntally delete the whole last word

...is a common occurrance.

In my opinion, this is a very important fix.

#### Environment information
- FlorisBoard Version: 0.3.10
- Device: Samsung Galaxy S9
- Android version, ROM: 10

### LINKED ISSUE #604: Suggestions replace previous word
#### Short description
Sometimes when I click on a suggestion provided after glide typing, it replaces the previous word before space. It happens randomly, and I'm not sure what triggers it.

#### Steps to reproduce
1. Open some app, start typing (swipe)
2. Click on some suggested word
3. See it replace previous word

#### Environment information
- FlorisBoard Version: 0.3.10-beta06
- Install Source: F-Droid
- Device: Oppo K3
- Android version, ROM: ColorOS v7.1(10)

patrickgold:
Fixes #581
Fixes #595
Fixes #604

@X-yl I've investigated into several glide suggestion issues and now changed some parts of the glide suggestion logic in the TextInputManager. Most importantly, I've eliminated the `hackyGlideSuggestionSkip` variable, because it was indeed very hacky :) When a glide word is now typed in, the phantom space flag is purposely not set, this way we can use the suggestion click logic without injecting some extra logic for glide typing. Furthermore, when manual typing in a letter after a glided word, a space is now set (except for characters like !?. etc.). The delete word feature now also resets when either clicking on a provided suggestion or when manual typing in a letter.

Could you check if I didn't accidentally remove a useful feature/functionality for the behavior directly after a glide word was typed in and also check if you can also not reproduce the bugs anymore? This would be very helpful!

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/core/EditorInstance.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/core/EditorInstance.kt
@@ -224,27 +224,6 @@ class EditorInstance private constructor(
                 ic.commitText(" ", 1)
             }
             ic.commitText(text, 1)
-            isPhantomSpaceActive = true
-            wasPhantomSpaceActiveLastUpdate = false
-            ic.endBatchEdit()
-            true
-        }
-    }
-
-    /**
-     * Replaces the previous word with the given [text]. Used to correct gestures.
-     */
-    fun commitGestureCorrection(text: String): Boolean {
-        val ic = inputConnection ?: return false
-        return if (isRawInputEditor) {
-            false
-        } else {
-            ic.beginBatchEdit()
-            markComposingRegion(Region(this, cachedInput.getWordForIndex(-1).start, cachedInput.getWordForIndex(-1).end))
-            ic.commitText(text, 1)
-            markComposingRegion(null)
-            isPhantomSpaceActive = true
-            wasPhantomSpaceActiveLastUpdate = false
             ic.endBatchEdit()
             true
         }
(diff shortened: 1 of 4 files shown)


X-yl:
Wow, that is a lot simpler than what I came up with, I feel pretty silly now haha! Thanks very much (and especially thank you for getting rid of `hackyGlideSuggestionSkip`, that was really quite terrible code). The only issue I spot is that this doesn't correctly deal with deleting gestures that contain an apostrophe. For example if you glide `doesn't` and then press backspace, it only removes the `t` and not the rest of the word. Other than that I don't think there are any issues!

patrickgold:
Thanks for checking the fix out, really appreciate it!

> The only issue I spot is that this doesn't correctly deal with deleting gestures that contain an apostrophe. For example if you glide doesn't and then press backspace, it only removes the t and not the rest of the word.

This issue is there because the whole delete words precisely feature the `handleDeleteWord()` method relies on is jank af (Set the "Delete swipe" gesture to "Delete words precisely" and try it out in a text field and you'll see what I mean haha). I will fix that feature with a separate PR though, the glide suggestion fix is unrelated to this.

PR #634, opened 2021-04-12
Update of SK layout

majso:
- adding ä character


patrickgold:
Thanks for your change, just reviewed it and it is ready to be merged.

Issue #677, opened 2021-04-15
OutOfMemory (OOM) Crash


patrickgold:
This issue is a collection for all OOM (OutOfMemory) related crashes, as the number of duplicates regarding this topic is getting out-of-hand. Further crash reports with a stacktracte similar to the one attached below will be marked as duplicate and linked to this issue.

In general, these are the known areas in v0.3.10 which are most likely the sources of OOM crashes and which need fixing for v0.3.11:
- **Suggestions**: The suggestions code's memory management is not great and will be replaced soon by other, more efficient methods. It is planned that the new implementation is written in C++, which avoids unnecessary overhead of the JVM and thus the memory usage should be much lower.
- **GlideTyping**: The most relevant memory leak (and really big resource hog) has already been fixed in #625, smaller memory leaks of event listeners will be fixed later.
- **LayoutManager**: The layout manager does a pretty good job at caching and preventing memory leaks, though the loading of layout json files can be improved to lower the stress on the CPU and to avoid potential garbage.
- **Re-initialization when rotating the screen**: The re-initialization of the service class when rotating the screen has some pretty nasty memory leaks and thus over the time a lot of garbage accumulates with also leads to OOM errors.
- **Settings UI**: Even though this is a small memory leak and definitely not relevant considering the garbage the above areas accumulate, the Settings UI has some small memory leaks as well which should be fixed.

---

```
java.lang.OutOfMemoryError: Failed to allocate a 24 byte allocation with 2027432 free bytes and 1979KB until OOM, target footprint 268435456, growth limit 268435456; failed due to fragmentation (largest possible contiguous allocation 24379392 bytes)
	at kotlin.reflect.jvm.internal.KParameterImpl.getType(KParameterImpl.kt:43)
	at kotlin.reflect.jvm.internal.KCallableImpl.callDefaultMethod$kotlin_reflection(KCallableImpl.kt:137)
	at kotlin.reflect.jvm.internal.KCallableImpl.callBy(KCallableImpl.kt:112)
	at com.squareup.moshi.kotlin.reflect.KotlinJsonAdapter.fromJson(KotlinJsonAdapter.kt:111)
	at com.squareup.moshi.internal.NullSafeJsonAdapter.fromJson(NullSafeJsonAdapter.java:41)
	at com.squareup.moshi.CollectionJsonAdapter.fromJson(CollectionJsonAdapter.java:81)
[...]

PR #711, opened 2021-04-19
OutOfMemory: Fix 1

### LINKED ISSUE #699: Bug report
~~~ 1618610515704.stacktrace ~~~

java.lang.NullPointerException
	at dev.patrickgold.florisboard.ime.clip.provider.FileStorage.cloneURI(FileStorage.kt:42)
	at dev.patrickgold.florisboard.ime.clip.provider.FlorisContentProvider.insert(FlorisContentProvider.kt:77)
	at android.content.ContentProvider$Transport.insert(ContentProvider.java:287)
	at android.content.ContentResolver.insert(ContentResolver.java:1557)
	at dev.patrickgold.florisboard.ime.clip.provider.ClipboardItem$Companion.fromClipData(FlorisDatabase.kt:123)
	at dev.patrickgold.florisboard.ime.clip.FlorisClipboardManager.onPrimaryClipChanged(FlorisClipboardManager.kt:214)
	at android.content.ClipboardManager.reportPrimaryClipChanged(ClipboardManager.java:222)
	at android.content.ClipboardManager$2.handleMessage(ClipboardManager.java:67)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:169)
	at android.app.ActivityThread.main(ActivityThread.java:6521)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)

patrickgold:
This PR is an initial attempt at reducing the memory strain and improving the codebase. This PR alone will not introduce noticeable changes for the crashes (I think). You are of course free to try the debug APK out and see if anything regarding the OOM issues have changed. Note that this PR assumes you have glide typing and suggestions disabled (will be addressed by future PRs).

During inspections of memory allocations, a few things could be seen by me:
- The average memory consumption over a larger timeframe is increasing very slowly, which means in general the GC is able to pick up and remove unused objects and the number of memory leaks is low.
- When typing a lot in a relatively short timeframe, a lot of memory allocations are done, which leads to enormous spikes in memory usage. This can easily lead to hitting the max. RAM usage limit before the GC has the ability to get rid of the objects, which then throws the infamous OOM exceptions. The primary goal in the next `oom-fix2` PR will be to remove the need to allocate so many new objects and also experiment with moving memory-intensive objects such as the layout and event objects to native code (C++). This should remove a lot of unneeded overhead for objects in the JVM and is especially useful for objects which exists in hundreds, if not thousands, such as KeyData and InputKeyEvent.

Additionally this PR cleans up quite some code (removing/adding spaces, new lines, etc., which I noticed during reading through the code).

---

Related #677 

---

Fix #699

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/debug/LogTopic.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/debug/LogTopic.kt
@@ -33,5 +33,7 @@ object LogTopic {
 
     const val IMS_EVENTS: FlogTopic =           1u
     const val KEY_EVENTS: FlogTopic =           2u
+    const val GLIDE: FlogTopic =                512u
+    const val CLIPBOARD: FlogTopic =            1024u
     const val CRASH_UTILITY: FlogTopic =        2048u
 }
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/clip/ClipboardHistoryItemAdapter.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/clip/ClipboardHistoryItemAdapter.kt
@@ -97,7 +97,7 @@ class ClipboardHistoryItemAdapter(
                 viewHolder.imgView.visibility = GONE
                 // For very large images, this can take a bit
                 FlorisClipboardManager.getInstance().executor.execute {
-                    val resolver = FlorisBoard.getInstance().context.contentResolver
+                    val resolver = FlorisBoard.getInstance().contentResolver
                     val inputStream = resolver.openInputStream(uri!!)
 
                     val drawable = Drawable.createFromStream(inputStream, "clipboard URI")
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/clip/ClipboardInputManager.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/clip/ClipboardInputManager.kt
@@ -1,8 +1,6 @@
 package dev.patrickgold.florisboard.ime.clip
 
 import android.annotation.SuppressLint
-import android.os.Handler
-import android.os.Looper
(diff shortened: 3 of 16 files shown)


Glitchy-Tozier:
What exactly do you store? I can think of storing the information for what buttons are pressed, but that shouldn't lead to memory problems if that data is cleared once it's not needed anymore.

patrickgold:
> What exactly do you store? I can think of storing the information for what buttons are pressed, but that shouldn't lead to memory problems if that data is cleared once it's not needed anymore.

Depends on what is enabled. If the big memory hogs are disabled, I still map all available layouts, load all used layouts for all subtypes, cache the current input string, preload the emoji keyboard, load the extended popup mappings, etc. One issue is that the layouts are unnecessarily recomputed when the keyboard is shown, another is that I have to recompute the cached input string for every key press (this is actually needed to have context about the current word, its caps state, etc.).

> but that shouldn't lead to memory problems if that data is cleared once it's not needed anymore.

This is true but objects remain in memory until the Garbage Collector decides to sweep across the memory and de-allocate all unused objects. When an object has just one nasty reference to it which I forgot to clear, the GC cannot clear the item anymore, leading to a memory leak. Also objects in Java have unnecessary overhead bytes for each and every object, which piles up with time. For this reason I will move memory intensive objects away from Java and to native C++, where I have full control over memory allocation/de-allocation, pointers, etc. and there's no overhead for objects.

Glitchy-Tozier:
I see, thank you for the explanation!

PR #714, opened 2021-04-19
Add support for Colemak-DHm layout

henningBunk:
I added the Coelmak-DHm layout which is described on this website: https://colemakmods.github.io/mod-dh/

It calls it "Mod DH" but on other sited everybody calls it DHm, so I went with the shorter and more common name. As the author I credited the original author which is also named in above site:

> The original Mod-DH was developed and launched by SteveP in October 2014

Cheers! 👋

### DIFF
--- a/app/src/main/assets/ime/text/characters/colemak-dhm.json
+++ b/app/src/main/assets/ime/text/characters/colemak-dhm.json
@@ -1,0 +1,46 @@
+{
+  "type": "characters",
+  "name": "colemak-dhm",
+  "label": "Colemak-DHm",
+  "authors": [ "SteveP" ],
+  "direction": "ltr",
+  "arrangement": [
+    [
+      { "code":  113, "label": "q" },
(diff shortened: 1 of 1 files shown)


patrickgold:
Thanks for your PR! Sorry for the late review, but in the last week I was focused on finishing the layout engine rework (#734). With the above PR the syntax for the layouts also slightly changes, see https://github.com/florisboard/florisboard/blob/master/app/src/main/assets/ime/text/characters/colemak.json as a example. Most notably is that the `shift` attribute is not existent anymore and has been replaced with a `case_selector`. Once that is fixed I will happily merge your changes into master!

henningBunk:
Thanks for the reply. I will have a look at the changes soon and edit the PR accordingly.

patrickgold:
Closing this PR as it is stale. Should you want to adapt the layout to the new syntax, feel free to re-open this PR.

PR #718, opened 2021-04-20
Recycle EmojiKeyViews to reduce memory usage

X-yl:
This PR decreases memory usage by using a `RecyclerView` to recycle `EmojiKeyView`s instead of creating one for each emoji. This reduces memory usage by about 15-20% when suggestions and glide typing are disabled (in my somewhat limited short tests)

Let me know if you'd like me to make any changes.

Related: #677

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/media/emoji/EmojiKey.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/media/emoji/EmojiKey.kt
@@ -26,6 +26,10 @@ class EmojiKey(override val data: KeyData) : Key(data) {
     var computedPopups: PopupSet<EmojiKeyData> = PopupSet()
         private set
 
+    companion object {
+        val EMPTY = EmojiKey(EmojiKeyData.EMPTY)
+    }
+
     fun dummyCompute() {
         computedData = data as? EmojiKeyData ?: computedData
         computedPopups = PopupSet(relevant = (data as? EmojiKeyData)?.popup ?: listOf())
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/media/emoji/EmojiKeyAdapter.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/media/emoji/EmojiKeyAdapter.kt
@@ -1,0 +1,59 @@
+package dev.patrickgold.florisboard.ime.media.emoji
+
+import android.view.View
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.flexbox.FlexboxLayoutManager
+
+
+class EmojiKeyAdapter(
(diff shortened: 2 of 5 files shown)


patrickgold:
Thanks for your PR! This is true, the EmojiKeyViews are also a huge memory hog. I am currently rebuilding the KeyboardView logic and also parts of the layout mechanism (the layout is built every time the window is shown, which is a huge waste of resources). When I am merging my keyboard view rework, I will review and merge your changes!

X-yl:
Sure. Out of curiosity though, what are you changing in KeyboardView?

patrickgold:
The branch is still in highly experimental state, but the "new" keyboard view works more similarly to the KeyboardView the Android system provided (is deprecated now). Namely: the KeyboardView does not have child views anymore, this is being taken care of by the "new" KeyboardView by directly drawing on the canvas. This also unifies the currently split touch/swipe gesture logic (some parts are in KeyView, some in KeyboardView), which has bugged me for a while now, as adding new features gets increasingly difficult if the logic is spread so much across two view classes. The keyboard layout and all layout/runtime theme parameters will be saved in a ComputedKeyboard object, which only gets updated if really necessary (not like I discovered every time the window opens). Even if I need to rebuild a layout (due to a preference change for instance), I do not need to rebuild the UI and triggering the Android measure-layout-draw cycle, which is very expensive on the performance if done too often.

For now this breaks glide typing, popups, extended popups and gestures, but with the way I structure the ComputedKeyboard I see no problem in adjusting those features so they work again flawlessly. With the rework I am also very cautious not to leave around object references or clear them after usage, to avoid unnecessary object allocations, etc.

X-yl:
Hmm, yeah that's probably for the best. It seems to me like the Android view/layout system is really not designed for this sort of thing, and flattening the view hierarchy would probably help a lot with performance and memory usage. Will you be implementing something similar for the emoji keyboard? If so then this PR is somewhat pointless and I'd suggest closing it. Let me know if there's anything you would like some help with, I seem to have a lot of free time right now lol.

patrickgold:
> Will you be implementing something similar for the emoji keyboard?

No, because the emoji stuff is completely separated (and are only planned to be reworked by me in v0.5.0) and thus I highly appreciate your changes, which will be very useful in reducing the memory strain. The main problem is that the emoji keyboard is entirely different to the character in almost all areas (especially because it has tabs and can scroll), so I will develop the "new" KeyboardView only for all character based layouts for now.

X-yl:
Sorry for the delay, I just rebased on the latest master (and also fixed some issues I didn't notice, oops).

patrickgold:
Thanks for rebasing your changes! I tested it out locally and then initial RAM usage is indeed a bit lower. There is another major issue with emojis though which I detected: the emoji file is re-parsed every time the device rotates, causing unnecessary memory allocations upon device rotation. This is though not topic of this PR, so I will merge your changes into master.

PR #727, opened 2021-04-21
Improve bug_report.md

Glitchy-Tozier:
Make everything use the same number of placeholder-periods (3). Also use "…" instead of three ".". It looks stupid in monospace, but it's nice in a regular font and is less annoying to delete.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -22,9 +22,9 @@ Guide to a good bug-report:
 Describe the bug in a short but concise way.
 
 #### Steps to reproduce
-1. Go to '...'
-2. Click on '....'
-3. Scroll down to '....'
+1. Go to '…'
+2. Click on '…'
+3. Scroll down to '…'
 4. See error
 
 #### Environment information


Glitchy-Tozier:
Two extra comments:
1. Why is there an empty line after the **assignee**-attribute before the `---`?
2. Why is there an empty line at the end of Environment Information? That should be removed, right?

patrickgold:
Thanks for your PR, really appreciate it! Didn't even notice I was so inconsistent with the dots, whoops :) Will merge your changes into master.!

> Why is there an empty line after the assignee-attribute before the ---?

GitHub originally created the template and everything between the first and second dotted lines is used to automatically add this template to the "New issue" list and to automatically add the labels. As I don't know if this line is needed or not (GitHub can be very picky about if it wants a new line or not in some Markdown syntax), I suggest we keep it there. The user will never see it anyways, so it won't be seen either.

> Why is there an empty line at the end of Environment Information? That should be removed, right?

It is best practice to end a file with an empty line. You will see this in pretty much all source files, and also in markdown files. It doesn't matter for the formatting if an empty line is at the end, so again, I suggest we keep it there.

Glitchy-Tozier:
Interesting, didn't know that.

Is the reasoning just to let people "code on" without needing to hit enter? Is there a reasoning?

PR #728, opened 2021-04-21
Improve crash_report.md

Glitchy-Tozier:
Make everything use the same number of placeholder-periods (3). Also use "…" instead of three ".". It looks stupid in monospace, but it's nice in a regular font and is less annoying to delete, especially when using a phone.

### DIFF
--- a/.github/ISSUE_TEMPLATE/crash_report.md
+++ b/.github/ISSUE_TEMPLATE/crash_report.md
@@ -20,9 +20,9 @@ Guide to a good crash-report:
 Describe what you were doing that could've led to the crash.
 
 #### Steps to reproduce
-1. Go to '...'
-2. Click on '....'
-3. Scroll down to '....'
+1. Go to '…'
+2. Click on '…'
+3. Scroll down to '…'
 4. See crash
 
 <!-- Paste the generated crash log below -->


Glitchy-Tozier:
Also, why is there an empty line after the **assignee**-attribute before the `---`?

patrickgold:
Thanks for your PR! Same as in #727 , will merge your changes into master.

Glitchy-Tozier:
If you don't mind helping me with GitHub, why do I receive the suggestion of deleting some branch?
I searched and apparently it wouldn't be destructive to delete the branch, but I'm not sure I understand the reason for why I would want to delete branches.

image

patrickgold:
That's pretty common for patch branches: you want to change something in the base repo, so you branch off. There you make your changes and create a PR (like you have done correctly). Then I review the PR and if everything checks out, merge the changes into master. As soon as I merge your changes into master, your patch branch becomes obsolete, as the changes are included in master and you can safely delete the patch branch. Of course you can keep the branch, but if you make frequent changes to a repo the patch branches will clutter your branch list real fast (imagine I wouldn't delete branches after merging, my branch list would be in the mid-hundreds by now haha). GitHub just tries to assist you and gives you a UI button to do so conveniently.

Glitchy-Tozier:
Thank you, I'll delete it then!

One day I really need to learn how exactly GitHub works... Maybe I'll need to start a project of my own to motivate me.

PR #734, opened 2021-04-22
Layout engine rework

### LINKED ISSUE #513: It's possible to open multipe pop-ups
Is this intentional? 

It seems like a handy system.
IF it's intended to work this way, i suggest always placing the most recent popup on top of all the other ones. Currently, there is no dynamic system that regulates this. For example, the `o`-pop-ups are always on top of the `u` ones.

#### Environment information
- FlorisBoard Version: 0.3.9
- Device: Samsung Galaxy S9
- Android version, ROM: 10

### LINKED ISSUE #515: Pop-up Won't Disappear
#### Short description
The pop-up that appears when you hold the return button does not disappear (visually). I have no idea how this can be reproduced. I opened up the keyboard and was faced with this issue. Half an hour later, it was fixed by itself. [Here's a video of the bug](https://i.imgur.com/CfEz3sY.mp4).

#### Environment information
- FlorisBoard Version: 0.3.9
- Install Source: GitHub
- Device: A Samsung Galaxy Note
- Android version, ROM: 7, Stock (Samsung's)

### LINKED ISSUE #676: Feature request for ENTER and SPACE button.
- Currently ENTER and SPACE button heights are too much big. It doesn't look comfortable with the other part of keyboard.
It would be nice to have option to change its width and height. 

thank you.

### LINKED ISSUE #730: Bug on Label and Letter Previews!
#### Short description
The label for alphabets in not shown in the keyboard app , In the central kurdish we have some vowel words (ھ - ێ - ڵ - ڕ and ە ) some of them have a same preview for example ( ە and ە ) this two words is diffrent but both of them are shown in same shape ( ە ) and also this ( ئ and ﺋ ) or letter Ha ( ھ ).

for that we are use the Label to them becuase of label the user is know which letter is Hah and which one is small vowel A for your information to understand me what i mean when i created the Central Kurdish layout to FlorisBoard i used this 

      { "code":  1607, "label": "ھ" },

      { "code":  1574, "label": "ﺋ" },

      { "code":  1749, "label": "ﻪ" },

in the keyboard app the label is not shown as you see above i used ( ھ ) for letter Hah and ( ﺋ ) for letter Alif.. etc 

but i don't know why is just show the code of letter and Label is not shown and that's difficult to users how to know which word is Hah and which one is vowel A look at the screenshot both of words is looks like eash other but they are so diffrent in writing...

[](https://postimg.cc/T5bpQj92)

The word in left is Hah and another one is vowel A,
For Hah it must be shown ( ھ ) label  and for vowel A it must be show ( ﻪ )

In the GBoard that is have a diffrent layout the Hah is shown correctly as ھ but in FlorisBoard is NOT :( 

this one is Gboard that show Hah correctly 

[](https://postimg.cc/f363dgcq)

I hope you understand me and fix that small bug for Labels (^_^)

#### Steps to reproduce
1. Go to 'Keyboard app'
2. Click on 'Writing something'
3. Scroll down to '…'

#### Environment information
- FlorisBoard Version: curently 3.11 beta01 and all other versions have same bug
- Install Source: GitHub
- Device: Realme X2 Pro
- Android: 10 , Realme UI 1.0

patrickgold:
This PR is a major rework in the way FlorisBoard loads assets, builds layouts and renders the keyboard. As the changes were so interconnected with each other, they are all included in one PR. See the sections below for what is changing.

## AssetManager

The asset manager got a rework and its method names and structure are more clear than before + the Result API usage got de-cluttered. Additionally, I switched from Moshi to kotlinx.serialization. Both libraries have its flaws, but kotlinx-serialization integrates a bit better with Kotlin source code and is a tad faster.

## Layout engine 1: The computed keyboard

Until now, all layouts have been loaded from different layouts to a computed layout, which then was assigned to the KeyboardView. This logic had serious flaws though:
1. The computed layout was rebuilt every time a subtype switch or window show event occurred - one major factor which played into the OOM crashes (note that there are a lot of other factors which also cause OOM crashes, so the OOM story is far from finished with this).
2. Every time a computed layout was applied to the KeyboardView, the whole UI was destroyed for that KeyboardView and rebuilt - a super expensive operation which side effects are large number of object allocations and possible memory leaks.
3. The syntax and structure of the old layout wasn't flexible enough for further expansion. Thus, it has been reworked.

## Layout engine 2: The UI
As mentioned above, each KeyboardView was a ViewGroup and thus had UI children. When rebuilding a layout, the whole Android measure/layout/draw cycle has to be triggered, which is pretty expensive, especially if the views are detached and attached frequently. Now, each KeyboardView is a plain View which renders the layout directly on the drawing canvas, circumventing the problem to recreate the UI every time the layout changes. Furthermore, the whole TextInputUi has been flattened, as seen below:

<table>
<tr>
<th>Old</th>
<th>New</th>
</tr>
<tr>
<td>
<pre>
&lt;TextInputUi&gt;
&nbsp;&nbsp;&lt;Smartbar ../&gt;
&nbsp;&nbsp;&lt;ViewFlipper ..&gt;
&nbsp;&nbsp; &nbsp;&lt;EditingKeyboardView ../&gt;
&nbsp;&nbsp; &nbsp;&lt;LoadingPlaceHolderKeyboardView ../&gt;
&nbsp;&nbsp; &nbsp;&lt;KeyboardView characters /&gt;
&nbsp;&nbsp; &nbsp;&lt;!--all other keyboard modes... --&gt;
&nbsp;&nbsp;&lt;/ViewFlipper&gt;
&lt;/TextInputUi&gt;
</pre>
</td>
<td>
<pre>
&lt;TextInputUi&gt;
&nbsp;&nbsp;&lt;Smartbar ../&gt;
&nbsp;&nbsp;&lt;KeyboardView alllayouts /&gt;
&lt;/TextInputUi&gt;
</pre>
</td>
</tr>
</table>

By doing this, I can avoid performance expensive layout measurement round-trips and I have fewer UI elements allocated in memory.

Minor other improvements:
- The height for the space and enter key is auto-adjusted when the theme is in borderless mode, because it looks weird if the keys are in full height while all other keys have no border around them. Closes #676 
- It is now impossible to open multiple popups at once in the text context and the popup state should reset correctly. CLoses #513, Closes #515
- The `label` defined in the layout json file is now used for the display rather than using the code to generate a character for it. Closes #730 
- Keys can now have multiple code points by using the "multi_text_key" class identifier.

## Layout engine 3: The logic
The new layout logic tries to abstract the layout, keyboard, key, etc. so other input layouts, especially non-Latin ones have a better base to start with. The KeyboardView idea is designed to render all sorts of layouts, even if they require a completely different structure than we know from Latin-based "traditional" layouts. Furthermore the variation selection and caps selection syntax in the layout.json files have changed and are way more clear. Selectors like this cań easily be created to have dynamic keys for other features too.


patrickgold:
This PR's modifications are now entering a stabilizing phase and I have fully re-implemented all the basics again. A lot of stuff in terms of memory management and performance has changed and hopefully improved for everyone. This PR of course only addresses the core of FlorisBoard, features like glide typing, suggestions and clipboard issues will be addressed at a later stage. Additionally the layout json syntax got an update because of the layout engine rework and it is now easier to define variant-dependent as well as case-dependent keys.

If you want to try out the debug build, you can download it [here](https://github.com/florisboard/florisboard/pull/734/checks). Be aware that the touch logic is still a bit wonky and that glide typing, gestures and emojis don't work in the current debug build, which I will fix in the next two days. If you have feedback about the debug build, feel free to comment in this PR! I will directly use it to improve this PR's changes before merging into master.

patrickgold:
This PR's main effort - to rewrite the layout logic and engine is now complete. In order to test out the changes and see if crashes etc. occur, I will release these changes to the beta branch now. What still does not work though and needs fixing in the next beta is:

- Glide typing (the touch logic bridge)
- Swipe actions which involve two pointers (space+shift and shift-swipe)

Other stuff which is currently broken:
- Emoji extended popups (won't see a fix until v0.5.0)
- Editing/selection keyboard (may or may not see a fix before v0.4.0 release)

Everything else should be implemented and working again, though of course bugs may occur.

PR #750, opened 2021-04-24
Spanish Readme

another-sapiens:
Heya! 
I seriously love this app, but didn't know how to contribute. It was only when I noticed the F-Droid description is not translated that I got the idea of doing it myself. 

Hope this helps the Spanish speaking community find your amazing keyboard.


patrickgold:
Thanks for your translation of the store description into Spanish, really appreciate it! I just wanted to say that translations are managed only on Crowdin but then realized that this isn't even possible for the store description (at the moment), so I will accept your changes :)

As for the language code used, it seems that by [this document](https://support.google.com/googleplay/android-developer/table/4419860?hl=en) only `es-ES` or `es-419` can be used (F-Droid adheres to this standard too afaik). If possible, could you move your translations to the `es-ES` directory? Other than that, I see no issue preventing me from merging your changes :)

another-sapiens:
Hi! 
I've tried to translate it using Crowdin too, but ended up doing it manually through the file as I couldn't find it.

About the language code, the official F-droid app's source has multiple folders for spanish in /metadata, including es_AR and es_MX. It would seem that those codes are accepted for other apps, too. 

I can change it anyways if you want, though.

patrickgold:
> About the language code, the official F-droid app's source has multiple folders for spanish in /metadata, including es_AR and es_MX. It would seem that those codes are accepted for other apps, too.

Hmm good to know. Is your translation generally applicable for Spanish or is it specifically for the Argentinian dialect? Base don that you should choose the `es_ES` or `es_AR` tag respectively.

another-sapiens:
Other Spanish dialects should be able to understand it perfectly, afaik. It's mostly a matter of preference, as someone might want it to feel more native to their region and use the `es-ES` if they're from Spain, for example, and they wouldn't be able to if it's already occupied by mine. 

Don't quote me on this, but I think I've seen es-AR on some apps and the Readme was displayed correctly.

We could either test it out like this and see if it works, or just change it to `es_AR`. Or `es-ES`, which is found on many apps.

Luensche:
@patrickgold Is it possible to add this files to Crowdin somehow? Not for this pr but I think other translations would be helpful as well.

another-sapiens:
@Luensche I've seen it implemented in some other apps, but it seems to take a few extra steps. TrackerControl's dev said something about needing something similar to this file: https://framagit.org/tom79/fedilab/blob/develop/crowdin.yml

I don't know what he meant because I'm new to this, but he's already implemented it. The commit is not too old, in case someone wants to see how he did it.

patrickgold:
@Luensche It is definitely possible to add this to Crowdin though the language code mapping must be done manually and even then it takes some time to see that everything works from the translation to actually showing up the correct file in the app store (the language codes can be picky sometimes when done via Fastlane). I will definitely do it sometime, though currently I am trying to fix OOM and other issues, which takes away most of my time.

@another-sapiens I will now merge your changes into master, thanks again for translating!

Luensche:
@patrickgold But do you want to receive prs for other languages?

PR #769, opened 2021-04-27
Improve ime activation

olLenz:
I did two changes in this PR:

1. Move the functions `checkIfImeIsEnabled` and `checkIfImeIsSelected` to a util class to reduce the huge size of the FlorisBoard class a bit
2. Get the result for a selection in the input method picker via a broadcast receiver instead of fetching it multiple times with random time delays


patrickgold:
Thanks for your PR! I've had a quick look over it and it looks pretty good! Using a broadcast receiver is waaay better than my wonky timer solution, really appreciate that you replaced this bit of code :)

Note that I will merge #734 first (which will happen in a few days) and then I will resolve potential merge conflicts in this PR.

patrickgold:
I've reviewed your changes and will momentarily merge your changes into master. They will be shipped later today with v0.311-beta02! Thanks again for your contribution!

Issue #420, opened 2021-03-02
Custom Domains


jetrosut:
Currently, only ".com", ".gov", ".edu", ".net", and ".org" are supported by FlorisBoard. Though, everyone has different TLDs and SLDs that they use commonly. An option where users can enter custom domains could be implemented.

artfulrobot:
Yes! Or, tbh, an option to disable that feature. The number of times i long press dot, end up picking .com and then having to press delete 4 times then find the alt characters button... just because i wanted to add a slash character in a url!

jetrosut:
> The number of times i long press dot, end up picking .com and then having to press delete 4 times then find the alt characters button

You could move your finger away from the pop up [to cancel out of it](https://github.com/florisboard/florisboard/issues/478#issuecomment-803943826).

artfulrobot:
Being able to cancel popups without landing on another random character is good, but it's currently awkward to do in my opinon.

For me, the domain shortcuts is just a papercut that gets in the way. Customisable TLDs would be better than what we have, but being able to bring up the normal punctuation menu would be better.

PR #795, opened 2021-05-03
Set main uri ending to .de and add .eu ending

Luensche:
I am wondering why ".com" is the first pop entry for the German layout. Mostly ".de" will be used for normal usage. In addition, this adds the ".eu" ending which is also used more frequently.

### DIFF
--- a/app/src/main/assets/ime/text/characters/extended_popups/de.json
+++ b/app/src/main/assets/ime/text/characters/extended_popups/de.json
@@ -96,12 +96,13 @@
     },
     "uri": {
       "~right": {
-        "main": { "code": -255, "label": ".com" },
+        "main": { "code": -255, "label": ".de" },
         "relevant": [
           { "code": -255, "label": ".ch" },
-          { "code": -255, "label": ".de" },
           { "code": -255, "label": ".at" },
-          { "code": -255, "label": ".net" }
+          { "code": -255, "label": ".com" },
+          { "code": -255, "label": ".net" },
+          { "code": -255, "label": ".eu" }
         ]
       }
     }


Luensche:
Is it possible to make a difference for de_DE, de_AT and de_CH? If yes, maybe it is a nice feature to change the uri for this locales to have their tld first. Maybe it's also good to add .fr for Swiss users.

Glitchy-Tozier:
> I am wondering why ".com" is the first pop entry for the German layout.

I suspect because it's the most important domain-ending. Personally, I visit `.com`-sites **WAY more often** than `.de`-sites, `.at`-sites, and `.ch`-sites combined.

And I wouldn't want `.com` to be moved.

> In addition, this adds the ".eu" ending which is also used more frequently.

This is an addition I agree with. `.org` should be added as well, in my opinion.

Luensche:
>I suspect because it's the most important domain-ending. Personally, I visit .com-sites WAY more often than .de-sites, .at-sites, and .ch-sites combined.

I think this is user dependent. If you are a developer or someone who read English texts very often, you will use ".com" probably the most. But if you are a "simple German-speaking user" ".de "will be the most.

Luensche:
Perhaps this could also be made customizable in the settings. I requested something similar in #796.

Luensche:
Found out that there is already an issue for customized domains: #420

What shall we do with this pr?

patrickgold:
Difficult to say, tbh. At the end of the day static domain definitions will have to make way for "domain_slots" (same system as with the currency symbols), though this time the domain set can be customized in runtime (unlike the static list of currency sets). Also Glitchy's point with `.com` being more often used is true (for me at least). So I guess we will drop this PR if it is okay for you and focus on making the domains customizable sometime in the future?

Luensche:
I will close this PR and hope, that the point in the future is not to fare away.

PR #817, opened 2021-05-05
System / internal user dictionary

### LINKED ISSUE #109: Private User Dictionary
It would be pretty cool to have a user dictionary that isn't the built in system dictionary. 

Anysoft keyboard has this feature 

private-store

patrickgold:
This PR is a draft for adding the ability to suggest words from either the system user dictionary (shared dictionary between all keyboards for custom entries) and an internal user dictionary dictionary (also custom defined entries but not shared with other keyboards). The suggested words integrate with the next-word predictions (but are currently prioritized).

The system/internal dictionaries are auto-enabled by default and available to manage in Settings > Typing > Dictionary group:
user-dictionary-prefs-preview

If you want, you can try out the debug build and give feedback in this PR!

Closes #109

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/UserDictionary.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/UserDictionary.kt
@@ -227,7 +227,8 @@ class SystemUserDictionaryDatabase(context: Context) : UserDictionaryDatabase {
 
         override fun query(word: String, locale: Locale?): List<UserDictionaryEntry> {
             return queryResolver(
-                selection = "${UserDictionary.Words.WORD} LIKE '%$word%' AND (${UserDictionary.Words.LOCALE} = '$locale' OR ${UserDictionary.Words.LOCALE} = '${locale?.language}' OR ${UserDictionary.Words.LOCALE} IS NULL)",
+                selection = "${UserDictionary.Words.WORD} LIKE ? AND (${UserDictionary.Words.LOCALE} = ? OR ${UserDictionary.Words.LOCALE} = ? OR ${UserDictionary.Words.LOCALE} IS NULL)",
+                selectionArgs = arrayOf("%$word%", locale.toString(), locale?.language.toString()),
                 sortOrder = SORT_BY_FREQ_DESC,
             )
         }
@@ -235,31 +236,34 @@ class SystemUserDictionaryDatabase(context: Context) : UserDictionaryDatabase {
         override fun queryAll(): List<UserDictionaryEntry> {
             return queryResolver(
                 selection = null,
+                selectionArgs = null,
                 sortOrder = SORT_BY_FREQ_DESC,
             )
         }
 
         override fun queryExact(word: String): List<UserDictionaryEntry> {
             return queryResolver(
-                selection = "${UserDictionary.Words.WORD} = '$word'",
+                selection = "${UserDictionary.Words.WORD} = ?",
+                selectionArgs = arrayOf(word),
                 sortOrder = null,
             )
         }
 
         override fun queryExact(word: String, locale: Locale?): List<UserDictionaryEntry> {
             return queryResolver(
-                selection = "${UserDictionary.Words.WORD} = '$word' AND (${UserDictionary.Words.LOCALE} = '$locale' OR ${UserDictionary.Words.LOCALE} = '${locale?.language}' OR ${UserDictionary.Words.LOCALE} IS NULL)",
+                selection = "${UserDictionary.Words.WORD} = ? AND (${UserDictionary.Words.LOCALE} = ? OR ${UserDictionary.Words.LOCALE} = ? OR ${UserDictionary.Words.LOCALE} IS NULL)",
+                selectionArgs = arrayOf(word, locale.toString(), locale?.language.toString()),
                 sortOrder = null,
             )
         }
 
-        private fun queryResolver(selection: String?, sortOrder: String?): List<UserDictionaryEntry> {
+        private fun queryResolver(selection: String?, selectionArgs: Array<out String>?, sortOrder: String?): List<UserDictionaryEntry> {
             val resolver = applicationContext.get()?.contentResolver ?: return listOf()
             val cursor = resolver.query(
                 UserDictionary.Words.CONTENT_URI,
                 PROJECTIONS,
                 selection,
-                null,
+                selectionArgs,
                 sortOrder
             ) ?: return listOf()
             return parseEntries(cursor).also { cursor.close() }
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/UserDictionary.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/UserDictionary.kt
@@ -16,23 +16,32 @@
 
 package dev.patrickgold.florisboard.ime.dictionary
 
+import android.content.ContentValues
 import android.content.Context
 import android.database.Cursor
+import android.net.Uri
 import android.provider.UserDictionary
 import androidx.room.ColumnInfo
 import androidx.room.Dao
 import androidx.room.Database
+import androidx.room.Delete
 import androidx.room.Entity
+import androidx.room.Insert
 import androidx.room.PrimaryKey
 import androidx.room.Query
 import androidx.room.RoomDatabase
 import androidx.room.TypeConverter
 import androidx.room.TypeConverters
+import androidx.room.Update
+import dev.patrickgold.florisboard.ime.extension.ExternalContentUtils
 import dev.patrickgold.florisboard.util.LocaleUtils
 import java.lang.ref.WeakReference
 import java.util.*
 
 private const val WORDS_TABLE = "words"
+
+private const val FREQUENCY_MIN = 1
+private const val FREQENCY_MAX = 255
 
 private const val SORT_BY_WORD_ASC = "${UserDictionary.Words.WORD} ASC"
 private const val SORT_BY_WORD_DESC = "${UserDictionary.Words.WORD} DESC"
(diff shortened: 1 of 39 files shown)


app/src/main/res/xml/prefs_typing.xml
```diff
@@ -69,6 +69,42 @@
...
+        <SwitchPreferenceCompat
+            android:defaultValue="true"
+            app:dependency="suggestion__enabled"
+            app:key="suggestion__enable_floris_user_dictionary"
+            app:iconSpaceReserved="false"
+            app:title="@string/pref__dictionary__enable_internal_user_dictionary__label"
+            app:summary="@string/pref__dictionary__enable_internal_user_dictionary__summary"/>
+
+        <Preference
+            app:dependency="suggestion__enable_floris_user_dictionary"
```
sabzo, line 100: is `android:defaul`t assumed to be false? On the click listener it sets to true, but what's the default?

app/src/main/java/dev/patrickgold/florisboard/settings/fragments/TypingInnerFragment.kt
```diff
@@ -0,0 +1,50 @@
...
+    override fun onPreferenceTreeClick(preference: Preference?): Boolean {
+        return when (preference?.key) {
+            PrefHelper.Dictionary.MANAGE_SYSTEM_USER_DICTIONARY -> {
+                val intent = Intent(USER_DICTIONARY_SETTINGS_INTENT_ACTION)
+                startActivity(intent)
+                true
+            }
+            PrefHelper.Dictionary.MANAGE_FLORIS_USER_DICTIONARY -> {
+                // NYI
+                true
```
sabzo, line 45: There's no visual element to the user that this option has been enabled. Perhaps there should be a switch on the Dictionary preference value item, an on and off.

app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/UserDictionary.kt
```diff
@@ -0,0 +1,155 @@
...
+            return parseEntries(cursor).also { cursor.close() }
+        }
+
+        private fun parseEntries(cursor: Cursor): List<UserDictionaryEntry> {
+            if (cursor.count <= 0) {
+                return listOf()
+            }
+            val idIndex = cursor.getColumnIndex(UserDictionary.Words._ID)
+            val wordIndex = cursor.getColumnIndex(UserDictionary.Words.WORD)
+            val freqIndex = cursor.getColumnIndex(UserDictionary.Words.FREQUENCY)
```
sabzo, line 133: Not sure where the frequency is being determined from, but it seems quite aggressive when a new word is being added, that it outweighs the other frequencies (seems closer to 255 than others).

If a user has added a new word what should the freq be? One way is to give it a high score, but the word maybe uncommonly used and wouldn't have a higher frequency otherwise.  Nevertheless this is still a vast improvement to have user added words.

app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
```diff
@@ -446,12 +451,16 @@ class TextInputManager private constructor() : CoroutineScope by MainScope(), In
                 activeDictionary?.let {
                     launch(Dispatchers.Default) {
                         val startTime = System.nanoTime()
-                        val suggestions = it.getTokenPredictions(
+                        val suggestions = queryUserDictionary(
+                            activeEditorInstance.cachedInput.currentWord.text,
+                            florisboard.activeSubtype.locale
+                        ).toMutableList()
```
sabzo, line 457: This eventually adds words to internal dictionary as user is typing without any special actions on behalf of the user. How is the freq being set here?

app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/UserDictionary.kt
sabzo: It seems a high frequency maybe the norm, considering that [Google's AOSP](https://android.googlesource.com/platform/packages/apps/Settings/+/6b8e658/src/com/android/settings/inputmethod/UserDictionaryAddWordActivity.java#36) default is 250:  
```java
private static final int FREQUENCY_FOR_USER_DICTIONARY_ADDS = 250;
```

app/src/main/res/xml/prefs_typing.xml
patrickgold: Oops, forgot to add the default attribute here, thanks for pointing this out!

app/src/main/java/dev/patrickgold/florisboard/settings/fragments/TypingInnerFragment.kt
patrickgold: That's because I am already prototyping with the internal manager UI (and will add it to the PR before merging it), so for debug-only preference stubs I don't add UI indications of stuff not working yet.

app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/UserDictionary.kt
patrickgold: That's a good question actually. As I am prototyping the manager UI for both the system and internal manager, I planned to allow the user to decide what value for the frequency should be used (and pre-fill it with 255, in case the user does not want to set this value himself). As 250 seems to be the default value though, I think I will use that as the pre-filled value.

The system user dictionary UI does not even allow for a frequency to be set, I assume it uses 250 as the default.

app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
patrickgold: Just for clarification - the user dictionary is a dictionary where people manually put their custom strings (and possibly with shortcuts) in, the keyboard does not automatically place stuff in there? I have the system user dictionary enabled for Gboard and OpenBoard and both don't seem to place their learned words in it, there's only stuff I manually added to the dictionary either in the keyboard UI or in the Language & Input screen. The automated learning stuff would be the current/next-word predictions (which is an entirely different topic).

app/src/main/res/xml/prefs_typing.xml
patrickgold: Never mind, this preference specifically does not have a default value set because it carries no value (is used as a UI element to open the user dictionary UI, no more).

app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
sabzo: yes [user dictionary](https://developer.android.com/reference/android/provider/UserDictionary) is 

> A provider of user defined words for input methods to use for predictive text input. Applications and input methods may add words into the dictionary. Words can have associated frequency information and locale information.

This dictionary will be accessible by other apps, where as the "internal" dictionary wouldn't. They're all user dictionaries, one is system wide, one is app specific (FlorisBoard).

It seems we can add to the system wide dictionary, but we can't add to the internal dictionary yet manually. But in the code highlighted in this comment, you're now taking the words being currently typed and saving it into the internal dictionary (database)? If so what is/was the frequency used.  

Actually what do lines 455-457 (above) do?

app/src/main/java/dev/patrickgold/florisboard/settings/fragments/TypingInnerFragment.kt
sabzo: Ok, what if this were triggered as well by longClicking on a word? A user in mid-sentence may not want to open the settings just to add their word(s).

patrickgold: The preference UI will use an dialog UI style for adding/modifying an entry data. I guess I can display this dialog when the user long presses an suggestion, but I have to build some kind of helper (AlertDialogs can only be displayed within an activity for whatever reason)

app/src/main/java/dev/patrickgold/florisboard/settings/UdmActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+                USER_DICTIONARY_TYPE_FLORIS -> R.string.settings__udm__title_floris
+                USER_DICTIONARY_TYPE_SYSTEM -> R.string.settings__udm__title_system
+                else -> R.string.settings__title
+            }
+        )
+        supportActionBar?.setDisplayHomeAsUpEnabled(true)
+
+        dictionaryManager.loadUserDictionariesIfNecessary()
+
+        binding.fabAddWord.setOnClickListener {  }
```
sabzo, line 126: I'm guessing this is to be added and I'm probably checking too early :) I like the FAB however.

app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
patrickgold: The above highlighted lines ask the user dictionary if it has words similar to the current input (so a read-only operation). Currently the frequency is used to sort the results (high to low freq) but it is kinda ignored afterwards (I have to adapt the results merging a bit for this to work properly).

On a technical level the internal dictionary already works (I imported the exported system dict entries) and suggestions from entries in the internal user dictionary are being displayed. Also, when using the Android Studio database explorer you can see the innerings of the user dictionary (database).

Currently I am working on the add/edit dialog and on populating the RecyclerView to show the entries in the UI.

app/src/main/java/dev/patrickgold/florisboard/settings/UdmActivity.kt
patrickgold: Yes, I am currently working on that :)

app/src/main/java/dev/patrickgold/florisboard/ime/dictionary/UserDictionary.kt
patrickgold: I am now using 128 as the default frequency, as it is pretty well in the middle and in my opinion better than using 250, where the word would be prioritized pretty much every time.

app/src/main/java/dev/patrickgold/florisboard/settings/UdmActivity.kt
patrickgold: The UI is now in and working, though input validation in the add/edit dialog boxes is currently missing, meaning you can currently insert anything you want. Will fix that tomorrow.

PR #833, opened 2021-05-08
Re-implement glide typing for new keyboard view

### LINKED ISSUE #802: [Bug]: Glide Typing seems to be broken in v0.3.11-beta03
#### Short description
In the current beta v0.3.11- beta03, glide typing seems to be broken, i.e. the keyboard's behaviour is like glide typing was enabled (although it obviously is enabled in _FlorisBoard's_ settings, for sure).

##### Remarks
1. To ensure that there's no interdependency with my environment, I installed both versions ("official" one taken from F-Droid repo – currently v0.3.10 – & beta taken from GitHub) in parallel, so that I'm able to compare them. In v0.3.10, glide typing works fine!
2. I'm really impressed that _FlorisBoard_ doesn't seem to rely on google's libs for glide/gesture typing – thumbs up! My CustomROM comes with AOSP keyboard by default; for which I had to install google's gesture typing libs afterwards. I did that by installing the magisk module "Install gesture typing enabled AOSP keyboard". But _FlorisBoard_ keeps glide typing available even with that Magisk module disabled - chapeau! :-)

#### Steps to reproduce
1. Go to 'settings'
2. Enable 'activate glide typing' (or something similar – it's "Glide Typing aktivieren" in german)
3. Leave settings
4. Try to use glide typing
5. Observe that it doesn't work

#### Environment information
- FlorisBoard Version: v0.3.11-beta03
- Install Source: GitHub
- Device: Poco X3 NFC
- Android: ArrowOS 11 Vanilla
- Google Framework: microG v0.2.18.204714
- other Frameworks & Mods: Magisk, LSPosed (each with some mods, e.g. XPrivacyLua)

patrickgold:
Closes #802 

@X-yl if possible could you have a look what I am missing? I tried around a lot, but in the end the glide trail is drawn, but no words are generated and I don't really know where in the glide typing implementation the error is. Thanks in advance!

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/text/gestures/GlideTypingManager.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/text/gestures/GlideTypingManager.kt
@@ -6,8 +6,8 @@ import dev.patrickgold.florisboard.ime.core.Subtype
 import dev.patrickgold.florisboard.ime.extension.AssetManager
 import dev.patrickgold.florisboard.ime.extension.AssetRef
 import dev.patrickgold.florisboard.ime.extension.AssetSource
-import dev.patrickgold.florisboard.ime.keyboard.Key
 import dev.patrickgold.florisboard.ime.text.TextInputManager
+import dev.patrickgold.florisboard.ime.text.keyboard.TextKey
 import kotlinx.coroutines.*
 import org.json.JSONObject
 
@@ -18,7 +18,6 @@ import org.json.JSONObject
 class GlideTypingManager : GlideTypingGesture.Listener, CoroutineScope by MainScope() {
 
     private var glideTypingClassifier = StatisticalGlideTypingClassifier()
-    private val initialDimensions: HashMap<Subtype, Dimensions> = hashMapOf()
     private var currentDimensions: Dimensions = Dimensions(0f, 0f)
     private lateinit var prefHelper: PrefHelper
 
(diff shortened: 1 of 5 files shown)


X-yl:
Sure, no problem, I'll take a look.

app/src/main/java/dev/patrickgold/florisboard/ime/text/gestures/GlideTypingManager.kt
```diff
@@ -18,7 +18,7 @@ import org.json.JSONObject
 class GlideTypingManager : GlideTypingGesture.Listener, CoroutineScope by MainScope() {
 
     private var glideTypingClassifier = StatisticalGlideTypingClassifier()
-    private val initialDimensions: HashMap<Subtype, Dimensions> = hashMapOf()
+    private val initialDimensions: Dimensions = Dimensions(80000.0f, 600000.0f)
```
X-yl, line 21: Haven't had time to clone and try it out but uh.. what's going on here exactly?

This was my way to avoid rebuilding the glide classifier when the dimensions of the keyboard change but the layout is the same. For example when switching to one handed mode or rotating the keyboard, there's no need to rebuild the gesture classifier, we can just map the new coordinates onto the old keyboard dimensions with a simple linear transform and pass the transformed coordinates to the gesture classifier.

So, I mapped each subtype to the dimensions the gesture classifier was initialized with. If you hard code it to 80000x60000 px I don't think it would work :P

patrickgold: Well that's a byproduct of the fact that since the layout rework upon initialization of the keyboard I don't know the dimensions of the keyboard yet (the view is unmeasured), but I need to pass the layout to the gesture classifier. I don't really know where to pass the initial dimensions, as it seems that the object which needs the initial coordinates is generated before I know how big the view is. Initially the values were 0,0 but I just tried some random values to see if it does something.

patrickgold: Additionally there's now only one keyboard view for all subtypes, so in my opinion no mapping is required (that's why I removed the map and made it a Dimension object only). The idea with the 0..1 normalization upfront sounds pretty interesting, is definitely a cleaner solution for the problem.

patrickgold: Ok I tried around a bit more and found out that just removing the initial dimensions and normalizing part does the trick it seems. You are using the initial dimensions in the normalizeX/Y methods only anyways, so the Pruner seems to generate off some other dimensions. The words are now generated perfectly, even if the size changes due to one-handed mode.

X-yl: OHHH I figured it out, `TextKey` is reused when switching to one handed mode now! So the `TextKey`s that Pruner has a reference to actually change their coordinates now. Ah that's a good solution, just wanted to point out though that you forgot to get rid of `currentDimensions` and `updateDimensions` which are now useless.

patrickgold: Ahhh that's why it works now I see. I didn't want to get rid of the current dimension because I didn't wanted to break things. :) But yeah, you are right, updating the dimensions is now pretty useless, will push a fix PR for this.

PR #842, opened 2021-05-09
Make normalizeBoxSide have a non-zero minimum longestSide

X-yl:
Noticed this issue while testing glide typing. If longestSide is ever zero, as sometimes happens because the dictionary contains "words" like "yyy" and "ggg", it causes NaN issues.


patrickgold:
Thanks for your PR! I've reviewed your changes and will merge your changes into master!

PR #850, opened 2021-05-10
Stop glide suggestions disappearing and remove the redundant first suggestion

X-yl:
No idea how I didn't notice it before but after glide typing a word, the suggestions are immediately replaced with the regular suggestions instead of showing the glide typing suggestions.

Another issue I noticded is that somewhat redundantly, the committed word shows as the first suggesion. I changed it to show the second best guess instead.


patrickgold:
Thanks for your PR! I've reviewed the changes and will merge it into master now. Will be released in v0.3.12 now (together with the glide shift hotfix).

PR #866, opened 2021-05-14
Composing input method (and Korean as the first subject)

### LINKED ISSUE #305: Add Korean layout please
Would love to see a Korean language option, I really like this keyboard and I have Korean friends whom I'm talking to. I'd appreciate were the Korean layout implemented! Thank you very much for this wonderful open source keyboard and looking forwrard to next updates!

### LINKED ISSUE #607: Implement a common manager for composing input methods
Many input methods requires multiple inputs to compose a character, for example, Telex for Vietnamese (#183), Indic scripts (#368) or Korean (#305). Therefore, it could be helpful to have a common manager for them that reads from a config file. Each row in the config file represent a transformation rule, with the LHS being the sequence of keys pressed and the RHS is the output character.

For example, config for Telex can look like:

```json
{
  "_depth": 2,
  "aa": "â",
  "as": "á",
  "âs": "ấ",
  "..."
}
```

I suppose we can use [getTextBeforeCursor][get] and [setComposingText][set] to implement this. Here is a pseudocode of how I think I would implement it:

```kotlin
fun onKeyPressed(pressedKey: Char, config: Json) {
    depth = config["_depth"]
    pattern = getTextBeforeCursor(depth - 1) + pressedKey
    if pattern is a key in config {
        replacedText = composingText.replace(pattern, config[pattern])
        setComposingText(replacedText)
    }
}
```

[get]: https://developer.android.com/reference/android/view/inputmethod/InputConnection#getTextBeforeCursor(int,%20int)
[set]: https://developer.android.com/reference/android/view/inputmethod/InputConnection#setComposingText(java.lang.CharSequence,%20int)

Hayleia:
Supposed to close #607 and close #305.

I tried to do a pass on documentation comments but maybe I still forgot some.

Some thoughts:
- this was already mentioned in the composing issue, but there is a line in EditorInstance.kt::commitText that calls a line in Florisboard.kt that seems a bit costly for what it does, might need to save a shortcut somewhere but I didn't know where to put it.
- I'm also wondering if all the composers in imeConfig.composers are not wasting space when unused instead of being released.
- about the regex for old subtypes, apparently according to the debugger, the code went in Subtype::fromString before getting to that regex so I had to support the old subtype format there, not sure if that's expected.

Breaking change: subtypes in config.json need a composer key. Mentioning this for any currently open issue adding layouts.

I put the Appender and the WithRules compoers in the file that declares the interface because they're kinda the default ones, and I put HangulUnicode in its separate file because it's specific. Not really defending that choice, just explaining why there are several files but not one for each.

Not related to composers but to Korean, I had to factor some things in TextKeyData to add support for shift-characters that are not necessarily uppercase non-shift-characters.

(Also you might want to squash the history because that branch is more "me figuring out how to android, how to florisboard, how to kotlin, and how to kotlinx serialization specifically" and less "me trying to get a clean history for a clean pull request".)


patrickgold:

I've had a quick look at your changes and judging from that the changes look promising. I will locally test out and give feedback on your changes tomorrow after I've released beta01.

patrickgold:
Thanks for your PR, I really appreciate it! I've just tested it out on my testing devices and (at least what an English speaking person can tell) works really really well, I'm really loving it!

There are some issues though which needs to be addressed. I will add comments to the appropriate code sections so it is clearer what feedback belongs where.

The commitText() part is a thing which will most likely stay in there for a while, I will see if I come up with a better place for it.

> I put the Appender and the WithRules compoers in the file that declares the interface because they're kinda the default ones, and I put HangulUnicode in its separate file because it's specific. Not really defending that choice, just explaining why there are several files but not one for each.

This is also how I've done it with the `KeyData` interface and totally ok with me.

> Also you might want to squash the history because that branch is more "me figuring out how to android, how to florisboard, how to kotlin, and how to kotlinx serialization specifically" and less "me trying to get a clean history for a clean pull request"

Yeah normally I am very open for just leaving the history as it is (as it showcases how the feature to-be-added was created), but in this case I might have to do some clean up :)

> about the regex for old subtypes, apparently according to the debugger, the code went in Subtype::fromString before getting to that regex so I had to support the old subtype format there, not sure if that's expected.

The regex is checked in the preferences initialization, so it should detect it there. You've come up with another solution though, which does not break the subtype configuration for other people when upgrading, so I think leaving this as it is is the best solution here.

app/src/main/assets/ime/config.json
```diff
@@ -1,5 +1,13 @@
 {
   "package": "dev.patrickgold.florisboard",
+  "composers": [
+    {
+      "$": "appender"
+    },
+    {
+      "$": "hangul-unicode"
+    }
+  ],
```
patrickgold, line 10: Not an error per se but it would look better if this was formatted this way:

```json
{ "$": "appender" },
{ "$": "hangul-unicode" }
```

app/src/main/assets/ime/text/characters/extended_popups/ko.json
```diff
@@ -0,0 +1,75 @@
+{
+  "type": "characters/extended_popups",
+  "name": "ko",
+  "authors": [ "patrickgold" ],
```
patrickgold, line 4: As you've designed the Korean layout files, you should also put your own name in the authors list :) I'm just adding a comment to this file, this applies to all layout files though.

app/src/main/assets/ime/text/characters/korean.json
```diff
@@ -0,0 +1,41 @@
...
+      { "$": "up_low_text_key", "code": 12610, "label": "ㅂ", "codeUp": 12611, "labelUp": "ㅃ" },
+      { "$": "up_low_text_key", "code": 12616, "label": "ㅈ", "codeUp": 12617, "labelUp": "ㅉ" },
+      { "$": "up_low_text_key", "code": 12599, "label": "ㄷ", "codeUp": 12600, "labelUp": "ㄸ" },
+      { "$": "up_low_text_key", "code": 12593, "label": "ㄱ", "codeUp": 12594, "labelUp": "ㄲ" },
+      { "$": "up_low_text_key", "code": 12613, "label": "ㅅ", "codeUp": 12614, "labelUp": "ㅆ" },
+      { "$": "auto_text_key", "code": 12635, "label": "ㅛ"},
+      { "$": "auto_text_key", "code": 12629, "label": "ㅕ"},
+      { "$": "auto_text_key", "code": 12625, "label": "ㅑ"},
+      { "$": "up_low_text_key", "code": 12624, "label": "ㅐ", "codeUp": 12626, "labelUp": "ㅒ" },
+      { "$": "up_low_text_key", "code": 12628, "label": "ㅔ", "codeUp": 12630, "labelUp": "ㅖ" }
```
patrickgold, line 18: This is what `case_selectors'' are useful for, see the https://github.com/florisboard/florisboard/blob/master/app/src/main/assets/ime/text/characters/colemak.json for example.

app/src/main/java/dev/patrickgold/florisboard/ime/text/composing/HangulUnicode.kt
```diff
@@ -0,0 +1,102 @@
+package dev.patrickgold.florisboard.ime.text.composing
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+@SerialName("hangul-unicode")
+class HangulUnicode: Composer {
```
patrickgold, line 8: Language-formatting (not an error but confusing to read): `HangulUnicode: Composer` indicates you have a HangulUnicode method which returns a Composer. Correct: `HangulUnicode : Composer`, meaning the HangulUnicode extends Composer.

app/src/main/java/dev/patrickgold/florisboard/ime/text/composing/Composer.kt
```diff
@@ -0,0 +1,54 @@
...
+    val name: String
+    val label: String
+    val toRead: Int
+
+    fun getActions(s: String, c: Char): Pair<Int, String>
+}
+
+@Serializable
+@SerialName("appender")
+public class Appender: Composer {
```
patrickgold, line 20: Same here, a space before the colon should be added. Additionally in Kotlin everything by default is `public`, so no need to add `public` before `class`. The same is also for all other classes in this file.

app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKeyData.kt
```diff
@@ -358,12 +349,49 @@ class AutoTextKeyData(
...
+        BasicTextKeyData(type, code, label, groupId, popup)
+    @Transient override val upper: BasicTextKeyData =
+        BasicTextKeyData(type, codeUp, labelUp, groupId, popup)
+
+    override fun toString(): String {
+        return "${UpLowTextKeyData::class.simpleName} { type=$type code=$code label=\"$label\" codeUp=$codeUp labelUp=\"$labelUp\" groupId=$groupId }"
+    }
+}
+
+
```
patrickgold, line 394: This is what `case_selectors` https://github.com/florisboard/florisboard/blob/f3f95ae28270cd8c6077bf4c1612c408c3a79596/app/src/main/java/dev/patrickgold/florisboard/ime/keyboard/KeyData.kt#L52-L81 are for, I suggest reverting all changes within the `TextKeyData.kt` file.

app/src/main/res/layout/settings_fragment_typing_subtype_dialog.xml
```diff
@@ -40,6 +40,18 @@
...
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:paddingHorizontal="8dp"
+                android:text="@string/settings__localization__subtype_composer" />
+
+            <androidx.appcompat.widget.AppCompatSpinner
+                android:id="@+id/composer_spinner"
+                android:layout_width="match_parent"
+                android:layout_height="48dp" />
+
```
patrickgold, line 54: Don't put 2 spinners in one linear layout, this messes up the padding. Wrap them in an additional linear layout like the other spinners too:

```xml
        <LinearLayout
            android:layout_width="match_parent"
            android:layout_height="wrap_content"
            android:orientation="vertical"
            android:paddingTop="8dp">

            <TextView
                android:id="@+id/textView"
                android:layout_width="match_parent"
                android:layout_height="wrap_content"
                android:paddingHorizontal="8dp"
                android:text="@string/settings__localization__subtype_composer" />

            <androidx.appcompat.widget.AppCompatSpinner
                android:id="@+id/composer_spinner"
                android:layout_width="match_parent"
                android:layout_height="48dp" />

        </LinearLayout>
```

Hayleia:
Thanks for your time, your patience, and your comments here.
And by the way, I never said it but thanks for giving me this opportunity to learn about everything I learnt while working on that pull request. Even the useless UpLowTextKeyData I just reverted was not a waste of time because I made my very first steps in kotlinx serialization there (then second steps with the Composers obviously).

I think I fixed everything if I didn't forget anything.

> The regex is checked in the preferences initialization, so it should detect it there. You've come up with another solution though, which does not break the subtype configuration for other people when upgrading, so I think leaving this as it is is the best solution here.

About this, I put breakpoints in fromString and on the line that checks the regex, and the code went in fromString (and crashed there) before getting to the regex check. Maybe I messed up something somewhere but yeah, I also expected it to get in the regex check first and it didn't happen, that's why I edited fromString.

patrickgold:
Perfect, thanks for reworking the mentioned issues!

Testing out your changes in the last three days, it worked pretty well, with the exception of the (new) Firefox URL bar, where the character insertion interferes with the URL suggestions, thus deleting the inputted character immediately. I think the problem is somewhere in the `commitText()` method. Furthermore, you define the `doCommitText()` as a local fun inside the `commitText()` fun - while technically possible, please move it out (local funs are not a good practice and should be avoided) and pass all required parameters (except `ic`, just get it like all other funs do too). This I think is the only issue with this PR before I can merge it in.

Hayleia:
Oops, I didn't test Firefox's bar. Seems to work with my latest commit though (just tried a fix at random, didn't expect it to actually work).
And I didn't know about that good practice. I was already pleased that my Kotlin worked at all, but yeah I should read about that too now.

Hayleia:
BTW, not sure how related to that pull request this is (maybe some changes here individually are ok, and some changes elsewhere individually are ok, but they don't go along together) but I noticed some performance issues with the latest commits on master.

Basically the branch composing1 as it is (without the latest commits) gives this:
https://user-images.githubusercontent.com/7904776/119248759-a94e4e00-bb93-11eb-80ce-ef570f0e50c7.mp4
So, it works, and the keyboard is responsive (key popups appear right when there is the dot that indicates the keypress).

But then if I import the latest changes from master, I get this.
https://user-images.githubusercontent.com/7904776/119248778-c6831c80-bb93-11eb-8f08-85898307965a.mp4
It still works but there is a noticeable delay between the actual keypress and the keypress being registered by the keyboard.

Note that my device is pretty old, it's a Galaxy Note 3 on Android 9 (Havoc 2.9).

patrickgold:
> well my phone hates the line that starts with "paint.setShadowLayer"

Hmm it seems that the shadow layer functionality is a pretty costly operation, works really smooth on newer devices, but seems to lag on you old device. Maybe I have to find another solution for this, thanks for letting me know of this!

As for your PR, I think all your changes are good now, will test it out later though for a final review.

patrickgold:
I've pushed a fix commit for the shadow (17dde536d9245ea6de689382cb7f4e9e8ea86118), could you try this one out if this fixes the performance issues on your device? Thanks in advance!

Hayleia:
I just tried. It doesn't seem much better unfortunately (I recorded a video but I don't seem to manage to send it now and I have to go, but I can send it later). Sorry :(

patrickgold:
Hmm that's really strange. As this issue with the elevation does have nothing to do with your composers though, I think I will merge in your changes now and then see if I can find another solution for the elevation separately.

PR #875, opened 2021-05-14
Reinitialize pruner when layout changes

### LINKED ISSUE #854: [Bug] Number row impacts glide typing accuracy
On the latest beta, enabling the number row severely impacts the accuracy of glide typing.

It appears the calculation of which key is hovered over is incorrectly calculated as the key position is offset

X-yl:
Closes #854.

This issue was that previously I assumed if the subtype is the same, the layout is the same. However, it is possible for the layout to change without the subtype changing, such as when enabling or disabling the number row. This change forces the pruner to be reinitialized if the previous layout is different.


patrickgold:
Thanks for your PR!  I've reviewed your changes and will merge them into master.

PR #908, opened 2021-05-18
Copy version string to clipboard on click on the version

### LINKED ISSUE #881: About section, longpress to copy version number
Just a small feature which can be useful to some.
Allow user to copy version number on longpress in about section:

Luensche:
Fixes #881

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/settings/AboutActivity.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/settings/AboutActivity.kt
@@ -17,17 +17,24 @@
 package dev.patrickgold.florisboard.settings
 
 import android.annotation.SuppressLint
+import android.content.ClipData
+import android.content.ClipboardManager
+import android.content.Context
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
 import android.view.MenuItem
 import android.webkit.WebView
+import android.widget.Toast
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.widget.Toolbar
 import dev.patrickgold.florisboard.R
 import dev.patrickgold.florisboard.databinding.AboutActivityBinding
+import dev.patrickgold.florisboard.ime.clip.FlorisClipboardManager
+import dev.patrickgold.florisboard.ime.core.FlorisBoard
 import dev.patrickgold.florisboard.util.AppVersionUtils
+import dev.patrickgold.florisboard.util.checkIfImeIsSelected
 
 class AboutActivity : AppCompatActivity() {
     private lateinit var binding: AboutActivityBinding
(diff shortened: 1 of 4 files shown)


Luensche:
I'm still doing something wrong with the toasts. At least they don't show up. Does anyone have a hint for me?

app/src/main/java/dev/patrickgold/florisboard/settings/AboutActivity.kt
```diff
@@ -42,8 +49,22 @@ class AboutActivity : AppCompatActivity() {
         val toolbar = findViewById<Toolbar>(R.id.toolbar)
         setSupportActionBar(toolbar)
 
-        // Set app version string
-        binding.appVersion.text = "v" + AppVersionUtils.getRawVersionName(this)
+        // Set app version string and onClickListener
+        val appVersion = "v" + AppVersionUtils.getRawVersionName(this)
+        binding.appVersion.text = appVersion
+        binding.appVersion.setOnClickListener {
+            val florisBoard = FlorisBoard.getInstance()
```
X-yl, line 56: Calling this will cause an exception if Florisboard is not the default keyboard
```suggestion
```

app/src/main/java/dev/patrickgold/florisboard/settings/AboutActivity.kt
```diff
@@ -42,8 +49,22 @@ class AboutActivity : AppCompatActivity() {
...
+            val isImeSelected = checkIfImeIsSelected(this)
+            if (isImeSelected) {
+                FlorisClipboardManager.getInstance().addNewPlaintext(appVersion)
+                Toast.makeText(florisBoard, R.string.about__version_copied__title, Toast.LENGTH_LONG).show()
```
X-yl, line 60: The context should be `this`.
```suggestion
                Toast.makeText(this, R.string.about__version_copied__title, Toast.LENGTH_LONG).show()
```

app/src/main/java/dev/patrickgold/florisboard/settings/AboutActivity.kt
```diff
@@ -42,8 +49,22 @@ class AboutActivity : AppCompatActivity() {
...
+            } else {
+                val clipboard: ClipboardManager = getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+                val clip = ClipData.newPlainText("Florisboard version", appVersion)
+                clipboard.setPrimaryClip(clip)
+                Toast.makeText(florisBoard, R.string.about__version_copied__title, Toast.LENGTH_LONG).show()
```
X-yl, line 65: Same here.
```suggestion
                Toast.makeText(this, R.string.about__version_copied__title, Toast.LENGTH_LONG).show()
```

app/src/main/java/dev/patrickgold/florisboard/settings/AboutActivity.kt
```diff
@@ -42,8 +49,22 @@ class AboutActivity : AppCompatActivity() {
         val toolbar = findViewById<Toolbar>(R.id.toolbar)
         setSupportActionBar(toolbar)
 
-        // Set app version string
-        binding.appVersion.text = "v" + AppVersionUtils.getRawVersionName(this)
+        // Set app version string and onClickListener
+        val appVersion = "v" + AppVersionUtils.getRawVersionName(this)
+        binding.appVersion.text = appVersion
+        binding.appVersion.setOnClickListener {
```
X-yl, line 55: Just a suggestion, you could apply this to the area including the icon, app name and the version, as otherwise it's a kinda small area which is clickable.

X-yl (commented):
This should hopefully fix your issue with the toast not showing.

Glitchy-Tozier:
Btw, the original request was to copy the version-nr on **long-press**.

I think this is a good solution instead of immediately copying the version nr.
However, there should probably be some immediate feedback, such as darkening the "copy-button"-area.

Luensche:
> 
> 
> This should hopefully fix your issue with the toast not showing.

Unfortunately, the toast still does not work. Maybe it's just because of my environment?

Luensche:
> However, there should probably be some immediate feedback, such as darkening the "copy-button"-area.

I need a hint for this how to implement it. Simply color changing won't be enough I guess.

X-yl:
> Unfortunately, the toast still does not work. Maybe it's just because of my environment?

Hmm I dunno. It works for me after the change you made.

> I need a hint for this how to implement it. Simply color changing won't be enough I guess.

[Take a look at this answer](https://stackoverflow.com/a/38984185). Set that on `headArea` (also you should rename it to `head_area` so it is snake_case like all the other IDs). Also I recommend adding `<corners android:radius="15dp">` or something within the `<shape>` to give it some nice rounded corners.

Luensche:
>Hmm I dunno. It works for me after the change you made.

I don't know why, but now the toast works.

patrickgold:
Thanks for your PR! I've reviewed it and it works pretty well! I will merge it into master now and release it with v0.3.13-beta02!

PR #920, opened 2021-05-22
Fix the default popup for some letters

### LINKED ISSUE #912: Wrong popup placed as default in Greek layout
#### Short description
After v0.3.13-beta01 update the default placed popup for "υ" and "ι" letters (they're the only ones to have multiple accents) has changed on Greek layout (for the respective capitals too). Probably the same thing happens on other layouts too, but I don't know how they're supposed to be so as to check :/

Here's v0.3.13-beta01
Screenshot_20210519-031643_Bromite

And here's v0.3.12 (same case for "ι")
Screenshot_20210519-031716_Bromite

#### Environment information
- FlorisBoard Version: v0.3.13-beta01
- Install Source: GitHub
- Device: Xiaomi Redmi Note 8
- Android: 11, custom rom

tsiflimagas:
This pr fixes a recently introduced minor issue with the order of popups for letters "ι" and "υ" in greek layout. I tried it locally and it works as expected. Closes #912

### DIFF
--- a/app/src/main/assets/ime/text/characters/extended_popups/el.json
+++ b/app/src/main/assets/ime/text/characters/extended_popups/el.json
@@ -20,8 +20,8 @@
         ]
       },
       "ι": {
+        "main": { "$": "auto_text_key", "code":  943, "label": "ί" },
         "relevant": [
-          { "$": "auto_text_key", "code":  943, "label": "ί" },
           { "$": "auto_text_key", "code":  912, "label": "ΐ" },
           { "$": "auto_text_key", "code":  970, "label": "ϊ" }
         ]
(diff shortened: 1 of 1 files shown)


patrickgold:
Thanks for your PR! While this works for your configuration, this may not be true if someone has enabled multiple symbol hints. You can enforce a "main" character, like here in German for example:

https://github.com/florisboard/florisboard/blob/04d3af64841ff969165c233956c9c359c53f0fbe/app/src/main/assets/ime/text/characters/extended_popups/de.json#L60-L66

This way you also help the "smart prioritization" in choosing the correct main popup character.

tsiflimagas:
Thanks for letting me know about this ability, it seems quite useful! I happen to miss new stuff, like this, sometimes, mostly due to not really knowing coding :/
It should be good now I think :)

patrickgold:
> I happen to miss new stuff, like this, sometimes, mostly due to not really knowing coding :/

That's more on me by forgetting to tell you this in the initial issue :) Also the extended popup mapping used here is unique to FlorisBoard, so it's fully ok if you don't know about the project's specifics!

>  It should be good now I think :)

Yes, it is! Will merge it into master.

PR #928, opened 2021-05-23
fix hint merge logic (#872)

### LINKED ISSUE #872: 1, 2, and 3 missing on Dvorak
#### Short description

The numbers 1, 2, and 3 are missing as the options on long-press on ',. keys on Dvorak layout. 
They are also not displayed the way the other digits are.

 
Screenshot_20210515-100139_Firefox_Beta

#### Environment information
- FlorisBoard Version: 0.3.11-beta04
- Install Source: GitHub
- Device: Xiaomi Mi 9T
- Android: LOS 17.1

ostrya:
The merge of the hints depends on the underlying main key. Especially, hints should only be shown for character keys, and if the hint is identical to the main key, it should not be added at all. Since the actual main key is only evaluated on demand with TextKey#compute, all corresponding hint merge logic needs to be moved there too.

Closes #872


patrickgold:
Thanks for fixing the issue for Dvorak. I've tested it out and will push it to master!

PR #963, opened 2021-05-30
Fix getting composer from name

Hayleia:
The `composers[composerNames.indexOf(activeSubtype.composerName)]` was always a bad idea and I knew it when I wrote it, but I thought "it's stupid but it's probably going to move to a better place anyways and for now it works" so I left it.

Except it doesn't work.

The `composers` list is unsorted, and the `composerNames` list is sorted, so their order is the same by pure chance in the current state of Florisboard, but that line still requires fixing before adding more composers. Otherwise this will happen.
2021-05-30 10_46_24-FlorisBoard – FlorisBoard kt  FlorisBoard app

This commit fixes it by adding a map to read the current composer from its name without sorting being a problem.
(Also it uses the constant Appender.name (that is already available for other purposes) rather than hardcoding "appender" when forcing the position of that composer to the first position in the list.)


patrickgold:
Thanks for your PR, really appreciate it! It's always those nasty little bugs that can cause a lot of confusion later on if not fixed, so extra thanks for fixing this one! (I would definitely need a good time searching for this sneaky one haha). Will merge it into master.

PR #966, opened 2021-05-31
Major input logic overhaul

### LINKED ISSUE #883: Rapid-fire deleting is delayed
#### Short description
When holding the delete-button to rapid-fire-delete (individual characters), the deletion-process is slow to start and keeps going for half a second after releasing the button.
It seems to lag behind whatever the user is doing.

#### Environment information
- FlorisBoard Version: 0.3.12
- Device: Samsung Galaxy S9
- Android: 10

### LINKED ISSUE #943: TextKeyBoardView rendering issues (in terms of performance)
#### Short description
The current implementation of the TextKeyboardView drawing methods works well, but in terms of performance is a real slowdown. This is especially noticeable since I merged in 845739015680cb31c1062cae19231144d43d001f, which causes severe lags (on older devices), because the shadows get redrawn way too often. This issue's goal is to find a way to cache as much as possible for the rendering process to make FlorisBoard more performant on older devices (where this issue is most noticeable).

Note: the v0.3.13-btea02 release will NOT contain key shadow support to minimize the lag in this beta release. v0.3.13-beta03 will introduce a new rendering concept and re-add support for key shadows. (#921)

#### Steps to reproduce
- N/A

#### Environment information
- FlorisBoard Version: 0.3.12-beta02 or newer
- Device: pretty much all, really noticeable though only on older devices.
- Android: Affects all API levels

### LINKED ISSUE #953: Keystrokes not registering
#### Short description
When hitting 2 keys fast, the first one is not typed. The phone vibrates 2 times, the key pop-up appears for both of the keys, but only the second key appears to have been registered. Now I have to type way slower or just redo words altogether because some letters are missing.

#### Steps to reproduce
1. Go to: I managed to reproduce this bug in all text fields I could find.
2. Click on: 2 sepperate keys really fast, one after the other.
4. See error: only the second letter appears

#### Environment information
- FlorisBoard Version: v.0.3.13-beta02
- Install Source: F-Droid
- Device: OnePlus 7T
- Android: 11, Stock

https://user-images.githubusercontent.com/23136219/119986066-161a7b80-bfcc-11eb-9be8-84f494fe1d61.mp4

### LINKED ISSUE #957: Space bar swipe gestures to move cursor lag greatly with suggestions enabled
#### Short description
Coming over from [openboard](https://github.com/dslul/openboard), I've grown used to swiping the spacebar left & right to navigate the cursor around quite aggressively. The issue is, though FlorisBoard can do the same, when the experimental suggestions feature is enabled it seems the cursor lags quite a bit if I try moving around too fast; sometimes being 2-3 seconds behind. 

Not sure if this is helpful, I've noticed too that it seems that the underline that shows the word you're currently on seems to "jitter" a lot more than openboard does when swiping the spacebar; and its especially apparent when I do so aggressively. It hasn't crashed however and is very usable if I move slowly, but it is quite inconvenient as is sometimes.

#### Steps to reproduce
1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

#### Environment information
- FlorisBoard Version: 0.3.12
- Install Source: F-Droid
- Device: Pixel 5
- Android: 11, Stock

*Thought I'd also add, this is an awesome project! I've never seen a keyboard as customizable, feature-packed, on top of being FOSS. Thank you for your work!*

patrickgold:
This PR is a major overhaul of the input logic and improves quite some aspects of when a re-draw is requested and how the input state is saved in the runtime. This should help reduce the amount of lags, especially for moving gestures like delete swipe or space bar swipe. The TextKeyboardView currently still uses the Android Canvas API to render itself, in the future it may be considerable to build the rendering with a SurfaceView or TextureView.

Additionally this PR fixes the "keys being eaten" bug and should make typing on FlorisBoard possible again.

If you want you can try out the debug artifact and can give me feedback directly in this PR. Once this PR is merged I will release v0.3.13-beta03 to test out these changes on a larger scale.

Closes #883
Closes #943
Closes #953
Closes #957


Glitchy-Tozier:
It's insane how much coding you're able to get done!

PR #973, opened 2021-05-31
Update kurdish.json

GoRaN909:
Made some changes of words position and added popup for some default characters.


patrickgold:
Thanks for your changes! I have tested them out locally for syntax etc. and everything checks out, will merge them into master.

PR #976, opened 2021-05-31
Update kurdish.json

GoRaN909:
Added stretched button (Kashida) to support all Kurdish layouts.

### DIFF
--- a/app/src/main/assets/ime/text/characters/mod/kurdish.json
+++ b/app/src/main/assets/ime/text/characters/mod/kurdish.json
@@ -1,26 +1,32 @@
 {
   "type": "characters/mod",
-  "name": "kurdish",
   "label": "کوردی",
+  "name": "kurdish",
   "authors": [ "GoRaN" ],
   "direction": "rtl",
   "arrangement": [
+
     [
+      { "code": 1600, "label": "kasheda", "popup":
+      { "main": { "code": 8204, "label": "key__view_half_space" }
+
+      } },
       { "code":    0, "type": "placeholder" },
       { "code":   -5, "label": "delete", "type": "enter_editing" }
     ],
     [
       { "code": -202, "label": "view_symbols", "type": "system_gui" },
-      { "$": "variation_selector",
-        "default":  { "code": 1567, "label": "؟", "groupId": 1 },
-        "password": { "code": 1548, "label": "،", "groupId": 1 },
-        "email":    { "code":   64, "label": "@", "groupId": 1 },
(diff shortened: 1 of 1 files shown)


app/src/main/assets/ime/text/characters/mod/kurdish.json
```diff
@@ -1,26 +1,32 @@
...
       { "code":   -5, "label": "delete", "type": "enter_editing" }
     ],
     [
       { "code": -202, "label": "view_symbols", "type": "system_gui" },
-      { "$": "variation_selector",
-        "default":  { "code": 1567, "label": "؟", "groupId": 1 },
-        "password": { "code": 1548, "label": "،", "groupId": 1 },
-        "email":    { "code":   64, "label": "@", "groupId": 1 },
-        "uri":      { "code":   47, "label": "/", "groupId": 1 }
-      },
```
patrickgold, line 19: You must use a variation selector for keys based on the current key variation, please leave this one in in this style. The old style which you inserted is not supported anymore.

app/src/main/assets/ime/text/characters/mod/kurdish.json
```diff
@@ -1,26 +1,32 @@
...
+      { "code":   64, "label": "@", "groupId": 1, "variation": "email_address" },
+      { "code": 1548, "label": "،", "groupId": 1, "variation": "normal" },
+      { "code": 35, "label": "#", "groupId": 1, "variation": "password" },
+      { "code":   47, "label": "/", "groupId": 1, "variation": "uri" },
       { "code": -210, "label": "language_switch", "type": "system_gui" },
       { "code": -213, "label": "switch_to_media_context", "type": "system_gui" },
-      { "code":   32, "label": " " },
-      { "code":   46, "label": ".", "groupId": 2 },
```
patrickgold, line 23: Same goes here, only that the inserted "old" variation selector is not needed because they are the same key for every variation.

app/src/main/assets/ime/text/characters/mod/kurdish.json
```diff
@@ -1,26 +1,32 @@
 {
   "type": "characters/mod",
-  "name": "kurdish",
   "label": "کوردی",
+  "name": "kurdish",
```
patrickgold, line 4: Semantically the "name" attribute should be above the "label".

GoRaN909:
You right please ignore this one and close the PR see #977  i fixed that erorrs :)

PR #977, opened 2021-05-31
Update kurdish.json

GoRaN909:
Added stretched button (Kashida) to support all Kurdish layouts.


patrickgold:
Thanks for correcting the mentioned issues, will merge the changes into master now :)

Issue #931, opened 2021-05-23
Add Vietnamese Keyboard


zT2Dz:
Your keyboard looks perfect for me, and I really want to use it in my everyday text conversations, but it doesn't support Vietnamese for now. Could you please add Vietnamese in the next update?. The Vietnames keyboard has the QWERTY design and Telex/VNI input encoding method. Thanks for reading.
Some helpful links: 
An opensource Android Keyboard: https://github.com/AgeOfMobile/Vietnamese-LatinIME
Unikey source: https://www.unikey.org/source.html

patrickgold:
thanhhocse96 did rework Telex by a lot and enhanced Telex. Could you please have a look at the latest beta release 0.3.16-beta02 and check it out?

patrickgold:
Also this issue is technically a duplicate of #183, so to avoid too many duplicate threads I will close this issue. If you have feedback on Telex please share it in #1841!

Issue #183, opened 2021-01-14
Support Vietnamese Input Method


ddtuan99:
Hi, first of all, thanks you guys a lot for your excellent work on the project!

About the feature request, i think the most popular Vietnamese input method on mobile is [Telex](https://en.wikipedia.org/wiki/Telex_(input_method)).

Rules of the method:
> Because the Vietnamese alphabet uses a complex system of diacritical marks, Telex requires the user to type in a base letter, followed by one or two characters that represent the diacritical marks: 

Non-tonal diacritics 
Character | Keys pressed | Sample input | Sample output
-- | -- | -- | --
ă | aw | trawng | trăng
â | aa | caan | cân
đ | dd | ddaau | đâu
ê | ee | ddeem | đêm
ô | oo | nhoo | nhô
ơ | ow | mow | mơ
ư | uw or w | tuw or tw | tư

Tone markings 
Tone | Keys added to syllable | Sample input | Sample output
-- | -- | -- | --
Ngang (level) | z or nothing | ngang | ngang
Huyền (falling) | f | huyeenf | huyền
Sắc (rising) | s | sawcs | sắc
Hỏi (dipping-rising) | r | hoir | hỏi
Ngã (rising glottalized) | x | ngax | ngã
Nặng (falling glottalized) | j | nawngj | nặng

> If more than one tone marking key is pressed, the last one will be used. For example, typing asz will return "a". (Thus z can also be used to delete diacritics when using an input method editor.) To write a tone marking key as a normal character, one has to press it twice: her becomes hẻ, while herr becomes her. 

There is another popular input method - [VNI](https://en.wikipedia.org/wiki/VNI#Input_methods) (I personally found it's frequently used on desktop more than on mobile). it uses numerical keys along the top of the keyboard (as opposed to the numpad) for entering tone marks and diacritical marks.

I really hope that Telex input method will soon be available in Floris Board.

If it can be implemented but considered low priority or there is lack of project's resource, could you introduce me some keyword/concept on how to implement it so i  can contribute the feature myselft?

patrickgold:
Thanks for the detailed feature request, I really appreciate it! Non-Latin layouts/languages will definitely be added to FlorisBoard, but are not included in the current milestone (v0.4.0), as this milestone is already too packed. I will assign this issue to v0.5.0.

> If it can be implemented but considered low priority or there is lack of project's resource, could you introduce me some keyword/concept on how to implement it so i can contribute the feature myselft?

[...]

PR #978, opened 2021-06-01
Basic Telex for Vietnamese

Hayleia:
I am not a Vietnamese language user so this pull request is a draft and I would like Vietnamese language users to test the debug artifact (once generated) and report if it seems usable.

This pull request adds vi-VN as default subtype, with qwerty as a layout and specific popups (ăâ on a, ôơ on o, etc), Vietnamese đồng as currency, and a Basic Telex composer.

So if you want to type phượng.
- "ph" is just "ph".
- ư can be typed either with just w, or with a popup on u, or with Telex syntax with uw, or even just with w.
- ợ needs two steps. First you need ơ which can be typed as a popup on o or with Telex syntax with ow. Then you add j.
- "ng" is just "ng".

Beware, the composer is called Basic Telex and not just Telex for a reason. There is no free order. you can not type ongwj and expect ợng. You need owjng, w first, j next, both right after the o (or ơjng using a popup, and j right after the ơ).

Video example.
https://user-images.githubusercontent.com/7904776/120271316-64b96580-c2ab-11eb-81ed-68f160589ea9.mp4

Feedback needed on:
- popup order
- relevant currencies
- usability of basic Telex


ddtuan99:
Thank you so much for your effort! 

I've just tested a bit. I feel (not so sure, I think I need more time on it) that basic Telex is ok. 

Right now, just one thing that I feel strange when typing in Firefox address bar, for example I have typed 'a' and the address bar is showing autocomplete url, say 'abc.com' then if I type another 'a' (to get â), the autocomplete url is gone but I still get only 'a', ie. I need to type 3rd 'a' to get â if there is an autocomplete url in the address bar. This does not happed with non-combine sequence, I mean, 'az' will give me 'az' correctly.

I have exclusively used [Open Board](https://github.com/OpenBoard-org/OpenBoard) (only support popup Vietnames layout, no Telex) for a while, I think they have quite complete character in popups (fare more than Floris board having right now).

Because of that I'm lakcing Telex's typing expreience so I hope I have time to look in to [Laban Key](https://play.google.com/store/apps/details?id=com.vng.inputmethod.labankey&hl=en_US&gl=US) - the most popular Vietnamese Keyboard, to give you more useful feedback. Also, hope there will be more people test your work.

Hayleia:
Not that Firefox URL bar again :(
Ok I'll try to have a look at it.

And I didn't know that popup-only was a viable option too. I mean, that's a lot of characters.
image
But this wouldn't be hard to add.

Thanks for the feedback!

ddtuan99:
I think popup characters won't have high priority if we support Telex (I'm sorry for being not clear in previous comment).

Right now, I personally think Free order Telex ("phuongwj" is equivalent to "phuwowjng") has 2nd priority. 
First priority is the ability to fallback to non-Vietnamese word. For example, I want to get "how", the expected sequence I need to type is "howw", that mean "how" will give me "hơ", typing another "w" will discard previous combination of "ow" so I will get "how".

Another feature I just remember when writting about 2nd priority is that "uow" will produce "ươ" (not "uơ") directly. I think this convention exist because in Vietnamese, "uơ" is invalid.

ddtuan99:
> And I didn't know that popup-only was a viable option too. I mean, that's a lot of characters.

It's extremely hard at first but I'm completely familiar with it now. The price of privacy.

Hayleia:
Ok, once the checks pass (well, they did before I was done typing this), the new debug artifact should work on firefox and should have these fallback rules.
There is one annoying rule though. Because ư can be obtained with uw and also just w, if you want to have uw as a result, you need to type uuww.
I also added the specific rule (and its fallback rule) for uow you mentioned.

Thanks again for the comments!

ddtuan99:
Thank you for your great work again. I'm using the new one. I will give you more feedback when I can.

patrickgold:
Thanks for your work on Vietnamese, really appreciate it!

As this PR is still marked as a draft, I will wait with my code/feature review until you feel ready. Giving feedback on your Telex implementation itself is difficult for me, because I am not familiar with these input type. Though should you have any questions or need assistance on the technical side, feel free to ask!

Hayleia:
Well I'm not familiar with it either, that's why I mentioned here and in issues #183 and #931 that it would require Vietnamese language users to validate that the debug asset seems usable. Hence why it's a draft, I wouldn't want it to be merged if it is not validated by users.

Huy-Ngo:
I would like to test this but I can't compile due to some CMake version incompatibility :cry:

patrickgold:
> I would like to test this but I can't compile due to some CMake version incompatibility cry

FlorisBoard requires CMake 3.10.2 or newer, then no incompatibilities should occur. What's the error you get when trying to compile?

Ashpex:
> I would like to test this but I can't compile due to some CMake version incompatibility

If you simply want to test it, here is an easier way: Go to [Checks](https://github.com/florisboard/florisboard/pull/978/checks) (when you are logged in GitHub) and look for the artifacts tab, you should be able to download the generated build there.

Huy-Ngo:
> FlorisBoard requires CMake 3.10.2 or newer, then no incompatibilities should occur. What's the error you get when trying to compile?

Sorry, I closed the shell so I can't paste the error log here, which was quite long anyway. It says I use too old version and it requires 3.19 or newer, but mine is 3.21 so I suppose it uses some flags that were deprecated in newer versions, or it is because I haven't been updating Android Studio (← off-topic discussion). Shouldn't matter for me anyways, because:

> If you simply want to test it, here is an easier way [...]

Thanks, this is simpler.

-----

As for feedback: This is neat, I don't think there is any problem (aside from a random glitch which would delete the composing character) with this layout. Though, I think I would try it for one week or two to be surer. I think @ddtuan99 can give some reviews as he(?) has been using it for 3 months tho.

The long delay led to these two conflicts so I hope this is resolved soon.

patrickgold:
Hmm the CMake issue is still weird but using the debug artifact APK is the simpler way I agree.

The two merge conflicts shouldn't be too much of a hassle to fix, considering the time which went by since this PR was opened there are very few conflicts.

> Though, I think I would try it for one week or two to be surer.

Yup this would be great, thanks! Sadly I do not have any knowledge in Vietnamese so besides checking the technical side I cannot test out this layout.

Huy-Ngo (approved):
So I've been trying for > 2 weeks now and while it's not exactly the same as TELEX IME as implemented in Unikey, it is even better than what I expected at first. I'm totally comfortable with this layout.

patrickgold:
Thanks Huy-Ngo for the feedback! From my side I am totally fine with merging this layout in, @Hayleia would just have to fix the merge conflicts first.

Hayleia:
Great to hear 😄
I think I successfully rebased my changes on some recent commit (not the latest one because I did it yesterday then wasted time with github stuff, but at least there doesn't seem to be conflicts) and I tried it quickly on my phone and it seemed to work.
So I removed the draft status.

patrickgold:
Thanks a lot for fixing the conflicts!

I've reviewed the changes and didn't find anything that stands out as being in the way of merging this layout in. Before merging though I will test this layout out on my local testing devices to ensure that everything runs smoothly (will do this probably tomorrow).

patrickgold:
I've reviewed your changes locally on my testing devices and everything works fine. Will merge your changes now in, will be available in 0.3.14-beta02 tomorrow!

quyleanh:
Thank you for PR.
Is there any plan for full Telex support. @Hayleia

Hayleia:
I wouldn't mind being the one who writes the code for it but I don't know the full specs (and I don't want to read some C/C++ code from Unikey to understand the full specs) so I don't think I'll be able to do it. I would need an exhaustive list of tests ("phượng":["phuongwj","phwowngj","phuwowjng",...]) that I could run to verify that the code does exactly what it's supposed to do on every relevant case for example and I didn't find that bundled with Unikey's source code.

quyleanh:
@Hayleia thank you for your kind explanation.  
Could you please share how will you implement the full Telex support? @patrickgold

Huy-Ngo:
I'm not florisboard developer, but:
I have *tried* to read [UniKey source code][uk], and I'd say it's not easy to just read that and, well, it's not easy to read it as there is little to no docstring in the source code. Characters are defined by hex code without notes on what it is; variables and macros are named in a way that may be intuitive to the developer at the time, but a totally new comers would struggle to find out their meaning. The source code is also more than 15 years old and doesn't compile on my machine with the instruction—maybe it used some nonstandard or deprecated syntax.

(Disclaimer: I'm not a C++ developer either so maybe if you're one you might have a better chance at reading it than me.)

[uk]: https://www.unikey.org/source.html

That is to say, implementing TELEX by just investigating the (only FLOSS) implementation. As far as I'm aware, late-correction is also not documented anywhere either. As Hayleia said, without a test suite it's  hard to verify the code work.

------

Well, let's not <i>bàn lùi</i> here. Excuse me for adding proposal as a comment in this PR but I'm not sure this is issue-worthy.

Is it a good idea to implement IMEs as external libraries @patrickgold ? It can be called by a simple function as

```kotlin
getTextFromIme("phuongwj", "telex") // output: phượng
```

If yes, is there constraint about implementation language? Does it have to be a JVM language, or just any language can do? I think a JVM language might be more compatible with the software but I don't have a good time with JVM languages myself. Python+Jython might also be an option.

And @quyleanh , if you're interested, you might help with building such IME for TELEX.

patrickgold:
> I have tried to read UniKey source code, and I'd say it's not easy to just read that and, well, it's not easy to read it as there is little to no docstring in the source code.

Ok so I had a bit of a look into the linked source code and I must say it is an okay C++ code, I have seen much worse things than this. The main problem is the sheer size of the source code, which would require a lot of effort to get into the structure of the program and understand how it works. And the character tables are a lot of Unicode mappings, where I do not really know how and why these are arranged like that without proper docstrings.

> Is it a good idea to implement IMEs as external libraries @patrickgold ? It can be called by a simple function as

What exactly do you mean with "implement IMEs as external libraries"? For me an IME is an application which ships its separate InputMethodService and functions stand-alone, whereas a library just provides some functionality to a service/app and cannot run on its own.

If it is the latter I see it as an advanced input event/text processor and I fully agree, that could and should be moved into a separate library. A stable API interface for this would have to be designed for this, which is as abstract as possible to allow special implementations for various non-Latin and all Latin based input.

If you want to implement this text processor you have my full support and I will try to assist you as best as possible (from the technical side at least). Note though that currently a lot of code is changing in the core to accommodate the new extension system I will introduce in the next 5-6 betas of the prefs rework. With full extension support FlorisBoard will be able to install extensions (themes, dictionaries, layout json files, simple composers and maybe more in the future). This especially affects the asset manager, layout manager, spelling manager, all of the old Prefs UI, and the Keyboard Core/UI implementation at the moment.

> If yes, is there constraint about implementation language?

If the library interfaces with the FlorisBoard IME in runtime then the language constrains are Kotlin/JVM or C/C++/JNI. The former is much easier to write and is faster for small tasks, whereas the latter is really useful when handling big amounts of data or needing high performance processing power. Note thea I have/had an OutOfMemory disaster when loading too much data into the Java heap with word suggestions, so you must be careful with loading too much code mappings at once if you are in Kotlin/JVM. You ca also split your library and utilize the best of both worlds.

If you want to write a tool when building which constructs the code mappings or other assets needed for the library, then you can choose Python or any other scripting language of course, as long it is installable on F-Droid's build servers.

dkellner:
Thanks for all the work you've put into this! FlorisBoard seems to be the only FLOSS Android keyboard featuring Telex at the moment.

I'm probably just blind, but how do I enable this running 0.3.14-beta08? I've added a new subtype in "Languages & Layouts" with the following settings:

- Primary language: Vietnamese (Vietnam)
- Popup mapping: vi-VN
- Characters layout: QWERTY
- Symbols primary layout: Western
- Symbols secondary layout: Western
- Currency set: Vietnamese dong
- Numeric layout: Western Arabic
- Numeric (advanced) layout: Western Arabic
- Number row layout: Western Arabic
- Phone primary layout: Simple telpad input
- Phone secondary layout: Simple telpad input (advanced controls)

This gives me the right popups (e.g. `u`'s popup contains `ư` and `7`), but Telex does not work, e.g. pressing `o` twice results in `oo` (not in `ô`) and `uj` just prints `uj` instead of `ụ`.

Hayleia:
I think composers were temporarily broken during the settings rework (any 3.14 beta version), which means if you want any language that requires them, you need to use the stable 3.13.

patrickgold:
Yup composers are currently broken as in there's not extension integration, thus FlorisBoard's beta UI currently does not offer an option for it either. This is something I will address before releasing stable 0.3.14.

dkellner:
Thanks for the clarification! In that case I'll just wait for 0.3.14, as 0.3.13 does not include Telex yet.

smartlitchi:
> I wouldn't mind being the one who writes the code for it but I don't know the full specs (and I don't want to read some C/C++ code from Unikey to understand the full specs) so I don't think I'll be able to do it. I would need an exhaustive list of tests ("phượng":["phuongwj","phwowngj","phuwowjng",...]) that I could run to verify that the code does exactly what it's supposed to do on every relevant case for example and I didn't find that bundled with Unikey's source code.

I think I found an [exhaustive list of tests](https://github.com/BambooEngine/ibus-bamboo/blob/master/src/github.com/BambooEngine/bamboo-core/bamboo_test.go) in the source code of ibus-bamboo, the IME I use on linux.

quyleanh:
@tuyenvm I'm sorry for mentioning you since you're seem busy with OpenKey. 
Could you please take a look at this pull request discussion, and give some ideas how to implement full support Telex in florisboard with OpenKey engine?
Thank you very much.

PR #982, opened 2021-06-02
Add the Halmak keyboard layout

dessalines:
Adding Halmak, an AI designed keyboard layout.

https://github.com/MadRabbit/halmak


patrickgold:
Thanks for your pull request! I've reviewed the changes and the syntax is ok. I've noticed though that, while technically correct, the third row has 10 keys, which leaves very little space per key to display itself. I wonder if it would be possible to remove (or move) the slash so the key hitboxes in the third row have more room? If you say that Halmak must have those 10 keys in the third row than we can merge it that was either, it just looks and feels a bit strange IMO.

dessalines:
Ah, I'd just assumed 10 keys for every row is best, but you're right its not necessary, as long as they're available from a symbols page. This is good to go then.

patrickgold:
Thanks for fixing the mentioned issue! I will merge your changes now, will release them in beta04.

PR #1006, opened 2021-06-07
Move new clipboard items with the same content to the beginning

### LINKED ISSUE #991: Don't keep duplicates in clipboard history
Like gboard does, it would be useful I think to not keep duplicated content in clipboard history, but move the existing item at the top if it's copied/cut again. Also, it seems that gboard doesn't apply this for images too, so I'm wondering if it's doable in general, so it could probably be done in FlorisBoard.

Luensche:
Closes #991

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/clip/FlorisClipboardManager.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/clip/FlorisClipboardManager.kt
@@ -110,10 +110,7 @@ class FlorisClipboardManager private constructor() : ClipboardManager.OnPrimaryC
         if (clipboardPrefs.enableHistory) {
             val clipboardInputManager = ClipboardInputManager.getInstance()
 
-            val historyElement = history.firstOrNull {
-                (it.data.type == ItemType.IMAGE && it.data.uri == newData.uri)
-                    || (it.data.type == ItemType.TEXT && it.data.text == newData.text)
-            }
+            val historyElement = history.firstOrNull { it.data.type == ItemType.TEXT && it.data.text == newData.text }
             if (historyElement != null) {
                 moveToTheBeginning(historyElement, newData, clipboardInputManager)
             } else {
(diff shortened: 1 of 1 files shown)


tsiflimagas:
Cool! Thank you very much for taking the time to implement this, it works as i expected it, great work! Looking at the code, I see (if I get it right) you implemented this for images also, as I wondered in the issue, but it seems to not work. So I only report this, if I didn't get something wrong.

Luensche:
Yes, you are right. It does not work as expected for images. I will have a look into it.

Luensche:
I removed the logic for images. Comparing text is very easy, but for images you can't compare this way. You first have to get the image by URI and then compare it. I'm not so deep in android and florisboard, that I know how to do it the best way. Maybe @X-yl or @patrickgold can give me a hint.

Luensche:
Another thought just occurred to me: is it possible that a URI is generated only for different blobs? Does that perhaps also save some memory?

X-yl:
Unfortunately you can't compare images by the URI, since each image will have a different one. I think you will have to load the Bitmap and use Bitmap#sameAs() to compare them. Not sure if it's worth the performance cost of having to load and compare every image on the clipboard every time you copy an image, though. 

If you want to be clever, you could try and get a quick hash of the images (or maybe just a partial hash) when they're first copied to florisboard and compare the hash later on.

Luensche:
> Unfortunately you can't compare images by the URI, since each image will have a different one.

Yeah, that's what I found out already. :persevere:

I don't know if it is worth to compare each image.

patrickgold:

Comparing images requires you to load both images and compare them pixel by pixel (or try to use hashes). IMO, it isn't worth the effort, most of the time text will be used anyways. Let me know if you are ready from your side, I will then review your changes and if everything checks out merge them!

tsiflimagas:
I guess I now know why gboard doesn't compare images either. But as Patrick said, doing it only for text is more than enough (I always use clipboard for text tbh). Thank you for this useful feature!

patrickgold:
Just checked out the changes and everything is working fine for text. As I will release beta04 now, I am merging in this PR. Thanks again for your work!

PR #1011, opened 2021-06-08
Code cleanup + some polish

X-yl:
Firstly I just wanted to make all the old clipboard code I wrote which used `Executor` and `Handler` use Kotlin coroutines, to make it consistent with the rest of the project.

I also have improved bitmap loading in the clipboard manager a lot -- loading smaller samples in the preview and preventing jerking while the images are loading. Scrolling should be much smoother. 

I have also taken Etienne's advice and cached the ideal word lengths, this should be a decent performance improvement. 

Finally I wanted to make the glide trail fade based on time, not based on how far the pointer travels. It's a much needed bit of polish that makes glide typing look much smoother, while also getting rid of the fairly pointless max length preference.

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/clip/ClipboardHistoryItemAdapter.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/clip/ClipboardHistoryItemAdapter.kt
@@ -1,6 +1,10 @@
 package dev.patrickgold.florisboard.ime.clip
 
-import android.graphics.drawable.Drawable
+import android.animation.ValueAnimator
+import android.content.ContentResolver
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import android.net.Uri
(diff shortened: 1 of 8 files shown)


patrickgold:
Thanks for your PR, really appreciate it! The glide trail and glide typing in general now feels a bit smoother, great job!

The only thing I don't like is that the drawn glide trail is now sooo thick - previously the glide trail was less prominent and thinner. (Although it seems like Gboard also has this thick trail, but still it looks a bit strange IMO).

X-yl:
So sorry, I totally forgot I opened this, haha. I've tweaked the reduction factor a bit, could you take another look and see if it's better?

patrickgold:
Sorry for the very long delay of the review of your changes, I was a bit distracted with the spell checker implementation! I've re-reviewed and tested out the changes and it is much better than before, thanks for fixing it. I will merge your changes into master and will release it in beta05.

Issue #76, opened 2020-12-08
Add Flow layout when implementing glide/swype typing


jeremiah-miller:
I just learned about the Flow keyboard after seeing [this issue](https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/2545) on ASK, and while I haven't spent a ton of time on it, my initial thought is it's super promising and if the reviews on google play are anything to go by, it's really good. That said, the emoji keyboard is seriously lacking, and there's no dark theme option. Perhaps that layout could be added to florisboard, and the glide typing as done in Flow could be used as an inspiration for implementing it in florisboard?

I'd love to help, however I'm not an android dev so I may not be able to help as much as I'd like :(

https://sourceforge.net/p/flowkeyboard/code/HEAD/tree/
https://play.google.com/store/apps/details?id=com.enchantedcode.flow

patrickgold:
Thanks for bringing up this new concept of text input. I've never heard of it, but on first look is a truly new way of inputting text on a phone screen. This is worth trying out at some time in FlorisBoard but would require a completely separate rendering and touch logic for this layout. At the current state I'm unsure if this concept will make it into FlorisBoard or if one should just use the original IME, but I'll definitely keep an eye on it. Note though that v0.4.0 (next planned major release which adds composing suggestions and hopefully glide typing) is due to start with developing, so I'll probably don't have time for looking at this topic until early 2021.

> I'd love to help, however I'm not an android dev so I may not be able to help as much as I'd like :(

Even if you are not an Android dev, you can still help by testing out new features, suggest ideas, report bugs and crashes, etc. This help is incredibly useful in improving FlorisBoard and for me it would not be possible to do this alone :)

jeremiah-miller:
Yeah, I tried looking through the source code of flow and florisboard to see if I could do it myself but I got lost real quick. I was hoping that since glide typing hasn't been implemented yet that it would be easy to use a similar algorithm for all the swipe typing in florisboard regardless of layout, but it sounds like that won't be possible which is unfortunate. 

I'll keep an eye out for anything, it's already a pretty solid keyboard from what I've seen so keep up the good work :+1:

patrickgold:
> I was hoping that since glide typing hasn't been implemented yet that it would be easy to use a similar algorithm for all the swipe typing in florisboard regardless of layout

[...]

PR #1015, opened 2021-06-10
Add keyboard layouts optimized for swype input

jeremiah-miller:
Layouts were found [here](http://sangaline.com/post/finding-an-optimal-keyboard-layout-for-swype/) and [here](https://jasmcole.com/2017/06/04/swype-right/), added as mentioned in #76

### DIFF
--- a/app/src/main/assets/ime/text/characters/dghp.json
+++ b/app/src/main/assets/ime/text/characters/dghp.json
@@ -1,0 +1,41 @@
+{
+  "type": "characters",
+  "name": "dghp",
+  "label": "DGHP",
+  "authors": [ "jeremiah-miller", "sangaline" ],
+  "direction": "ltr",
+  "arrangement": [
+    [
+      { "$": "auto_text_key", "code":  100, "label": "d" },
+      { "$": "auto_text_key", "code":  103, "label": "g" },
+      { "$": "auto_text_key", "code":  104, "label": "h" },
+      { "$": "auto_text_key", "code":  112, "label": "p" },
+      { "$": "auto_text_key", "code":   97, "label": "a" },
+      { "$": "auto_text_key", "code":  115, "label": "s" },
+      { "$": "auto_text_key", "code":  106, "label": "j" },
+      { "$": "auto_text_key", "code":  114, "label": "r" },
+      { "$": "auto_text_key", "code":  107, "label": "k" },
+      { "$": "auto_text_key", "code":  110, "label": "n" }
+    ],
+    [
+      { "$": "auto_text_key", "code":  105, "label": "i" },
+      { "$": "auto_text_key", "code":  113, "label": "q" },
+      { "$": "auto_text_key", "code":  118, "label": "v" },
+      { "$": "auto_text_key", "code":  117, "label": "u" },
+      { "$": "auto_text_key", "code":  119, "label": "w" },
+      { "$": "auto_text_key", "code":   99, "label": "c" },
+      { "$": "auto_text_key", "code":  108, "label": "l" },
+      { "$": "auto_text_key", "code":  120, "label": "x" },
+      { "$": "auto_text_key", "code":  109, "label": "m" }
+    ],
+    [
+      { "$": "auto_text_key", "code":  116, "label": "t" },
+      { "$": "auto_text_key", "code":  121, "label": "y" },
+      { "$": "auto_text_key", "code":   98, "label": "b" },
+      { "$": "auto_text_key", "code":  101, "label": "e" },
+      { "$": "auto_text_key", "code":  122, "label": "z" },
+      { "$": "auto_text_key", "code":  102, "label": "f" },
+      { "$": "auto_text_key", "code":  111, "label": "o" }
+    ]
+  ]
+}
--- a/app/src/main/assets/ime/text/characters/nalmy.json
+++ b/app/src/main/assets/ime/text/characters/nalmy.json
@@ -1,0 +1,41 @@
+{
+  "type": "characters",
+  "name": "nalmy",
+  "label": "NALMY",
+  "authors": [ "jeremiah-mille", "jasmcole" ],
+  "direction": "ltr",
+  "arrangement": [
+    [
+      { "$": "auto_text_key", "code":  122, "label": "z" },
+      { "$": "auto_text_key", "code":  118, "label": "v" },
+      { "$": "auto_text_key", "code":  100, "label": "d" },
+      { "$": "auto_text_key", "code":  110, "label": "n" },
+      { "$": "auto_text_key", "code":   97, "label": "a" },
+      { "$": "auto_text_key", "code":  108, "label": "l" },
+      { "$": "auto_text_key", "code":  109, "label": "m" },
+      { "$": "auto_text_key", "code":  121, "label": "y" },
+      { "$": "auto_text_key", "code":  120, "label": "x" },
+      { "$": "auto_text_key", "code":  106, "label": "j" }
+    ],
+    [
+      { "$": "auto_text_key", "code":  107, "label": "k" },
+      { "$": "auto_text_key", "code":  103, "label": "g" },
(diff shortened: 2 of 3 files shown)


dashezup:
Nice I would like to try it. I've been using Sangaline's one for a while (by patching Gboard APK to include that layout).

Was those layouts ever been include in any of other keyboard app? or this could be the first one?

Sangaline didn't name it in that blog article. @sangaline would you name the best swype keyboard layout you designed? such like after your name Sangaline or just DGHP?

jeremiah-miller:
It was referred to as DGHP in the paper they wrote, so that's what I went with. Personally i think DGHPAS is nicer, but this is fine too http://arxiv.org/abs/1503.06300

dashezup:
> It was referred to as DGHP in the paper they wrote, so that's what I went with. Personally i think DGHPAS is nicer, but this is fine too http://arxiv.org/abs/1503.06300

that's ok then, I didn't know this before.

jeremiah-miller:
How do you pronounce it in your head/if you were to say it out loud? I think i'm gonna go with pronouncing it like "dope", as if it was "dough" without any vowels and then 'p' on the end. But i'm curious if you have any other ideas :) (this is just small talk, feel free to not respond or whatever lol)

dashezup:
yeah it's a bit strange to try to pronounce DGHP.

I named it Sangaline on my Gboard patch (been using it longer than 1 year iirc). I'm more comfortable to call it with its inventor's name, I also use Dvorak (more specifically Programmer Dvorak on desktop/laptop) which is also named after the inventor's name. But since Sangaline named it DGHP in that paper it may be fine to just call it DGHP.

It's kind of fun to make others to confused when you were using a custom keyboard layout. although in most cases they wouldn't surprised cuz I use Dvorak on desktop, they mostly thought it's also Dvorak on my phone.

sangaline:
I'm excited to see this, thank you so much for putting it together and looping me in. I've actually never had the opportunity to use the keyboard myself, and I'm very much looking forward to giving it a try!

It's unfortunate that the keyboard doesn't start with something more pronounceable or memorable. We called it DGHP in the paper because we needed a convenient way to refer to that specific layout, but I wasn't really thinking about the implications as a user-facing name at the time. I agree with @jeremiah-miller that DGHPAS seems a bit nicer than DGHP. I wouldn't be bold enough to suggest "sangaline" as a name myself, but you certainly won't hear any complaints from me if you decided to go with that :-). It has the advantage that Googling ["sangaline keyboard"](https://www.google.com/search?q=sangaline+keyboard) brings up the [blog post](http://sangaline.com/post/finding-an-optimal-keyboard-layout-for-swype/), the open-source [dodona keyboard design/simulation framework](https://github.com/sangaline/dodona), a [Hacker News discussion about the keyboard](https://news.ycombinator.com/item?id=9355276), and [the paper](https://arxiv.org/abs/1503.06300) as the first four results. Similar searches for ["dghp keyboard"](https://www.google.com/search?q=dghp+keyboard) and ["dghpas keyboard"](https://www.google.com/search?q=dghp+keyboard) turn up nothing relevant.

jeremiah-miller:
I think "sangaline keyboard" is the best bet, but i'll be away from my computer till sunday night or monday, so if someone else could make the change i'd really appreciate it, if not i can do it then!

dashezup:
> I've actually never had the opportunity to use the keyboard myself, and I'm very much looking forward to giving it a try!

This patched old version of gboard apk is the one I'm using. it's mainly based on [the script in this gist](https://gist.github.com/empathicqubit/c075158e915e4d85651bb83d4196cab4), layout named Sangaline. I haven't try to patch gboard apk for a while because it has signature verification and it's a boring work to bypass it by editing smali codes.

https://ezup.dev/tmp/2wdipiy5_com.google.android.inputmethod.latin_26890814.apk (link available for 30 days)

It has symbols hidden as well. It seems that Florisboard does not have options to hide those elements to make it clean like this for now.



> I think "sangaline keyboard" is the best bet, but i'll be away from my computer till sunday night or monday

you can also edit on GitHub web UI. And this may could be "squash and merge".

jeremiah-miller:
cc @jasmcole

jasmcole:
> cc @jasmcole

Cool! Thanks for the ping - I'll have to give this a go.

jeremiah-miller:
Have you all been using these and comparing them? So far it seems to me that nalmy is easier for real life use, since it's primary goal is to minimize finger travel, whereas the sangaline board is trying to minimize errors. That's great, except it does that by making swipe patterns longer and more complex, which seems to make swiping less comfortable or convenient. It would be interesting to see the error rate reduction that nalmy has when running it through dodona

dashezup:
> Have you all been using these and comparing them? So far it seems to me that nalmy is easier for real life use, since it's primary goal is to minimize finger travel, whereas the sangaline board is trying to minimize errors.

I wondered about that as well, that's a good summary. I've only been using Sangaline layout, tried NALMY a little bit, and I don't think I can get used the two layouts. from my experience:

- with sangaline layout, for lots of words, I need to across almost the whole width of the keyboard to type them
- dvorak is just bad for swype, QWERTY may be fine for swype typing especially comparing to dvorak
- it's easier to achieve touch-typing-alike effect with sangaline layout, because most of used letters are either located at left side, right side or center which are easily to be located. while NALMY has most used letters at the center, most uncommon letters on left side and right side, usually you need to pay more attention on the center of the keyboard with eyes while typing (idk if this still matter if the one get used to this NALMY layout).

regarding the finger traveling problem on Sangaline, one hand mode may be the solution for that but I'm not fan of it. I feel more comfortable when the keyboard is centered on phone screen. it should be fine on phones with smaller size of screen, such like old iPhone or iPhone SE (4.7 or 5.5 inch), but iPhone iOS is not that friendly for using modified Apps compare to Android.

btw it should be **swype** instead of **swipe** you may edit the issue title as well

jeremiah-miller:
I think you're right that sangaline might be better with small keyboards, but even then it would be annoying to always be swyping back and forth. 

> while NALMY has most used letters at the center, most uncommon letters on left side and right side, usually you need to pay more attention on the center of the keyboard with eyes while typing (idk if this still matter if the one get used to this NALMY layout).

I've been using the [flow keyboard](https://play.google.com/store/apps/details?id=com.enchantedcode.flow) for a while and it's like that (but takes that philosophy of most-used-in-the-center to the next level). Once you get used to it it's super fast and easy, if it had a better emoji selector it's all I'd use

patrickgold:
Sorry for the long delay on the review of your additions, I was distracted by the spell checker implementation the last few weeks.

I've reviewed and tested out the layouts on my local testing rig and everything looks fine, the codes match up with the labels and auto-capitalization works also fine. Thanks for the work, really appreciate it :)

If it is ready for you I can merge in the changes and add it to the beta05 release (which will come out in the next few days after working through the backlog).

jeremiah-miller:
No worries, yeah it's all set to integrate! Thanks!

patrickgold:
Perfect, will merge it then!

jeremiah-miller:
cc @jasmcole I've been using NALMY for about a month now and I think I've got a pretty good hang of it, however I think swapping 'R' and 'S' would help reduce a lot of errors (so I don't type "it" when I want "is" or vice versa). And there might be a few other small optimizations here and there, but I think that's one of the biggest ones. I'll add any more I think of here and then try out a tweaked version of it and if it's better then I'll see about getting it added. If anyone else has any tweaks in mind drop them here too so I can try them out.

Also if this isn't the best place for this kind of discussion I'm happy to move it, I wasn't sure where would be best since I also want to get ideas from other people who might've been using NALMY for a bit now

jeremiah-miller:
"She"/"the" are another one that would be fixed by swapping 'r' and 's'

Some other ones with less obvious solutions:

- "I'm"/"I'll" are way to easy to mix, could be fixed by moving 'm' or 'l' to the bottom of the keyboard

- "dipping"/"doing"/"dropping" are extremely similar, but i have no idea what a simple fix could be

PR #1021, opened 2021-06-12
added support for direct boot

### LINKED ISSUE #485: Enable keyboard at startup (Direct boot mode)
If a password is set on the system, it's impossible to type the password with florisboard since the keyboard doesn't start at boot.
It would be nice to make florisboard start at boot.
Simple keyboard has this feature and ccording to this issue https://github.com/rkkr/simple-keyboard/issues/85 the keyboard needs to support Direct Boot Mode.

### LINKED ISSUE #1019: Ability to start on boot
Please make florisboard accessible without unlocking so it can be used to type password on boot.

### LINKED ISSUE #1046: FlorisBoard does not load after powering on
#### Short description
Describe the bug in a short but concise way.
I let my phone die for the first time after installing FlorisBoard.  Upon restart, I could not and still cannot enter my passphrase because apparently there is no keyboard.

#### Steps to reproduce
1. Install FlorisBoard, 
2. Make it default keyboard
3. Uninstall other keyboards
4. Let phone die or restart phone
5. No keyboard popup to enter passphrase

#### Environment information
- FlorisBoard Version: ~No idea, cannot check either since I cannot enter my phone~ 0.3.12
- Install Source: F-Droid
- Device: Samsung Galaxy S10e
- Android: 11

empratyush:
[Direct boot](https://developer.android.com/training/articles/direct-boot) allow an app to work before [Credential protected storage is available](https://www.youtube.com/watch?v=fb2Zwmc3Sp4), I have not done anything excellent because that's useless IMHO, UI works in default mode and override it with user configuration as soon credential protected storage is available.

Project might consider using device protected storage instead of credential protected storage for configuration and other non sensitive stuff but that's up to you. I am no one to make that decision.

This is BASIC implementation more testing and development might be needed.

I have tested it on Poco f1 and Pixel 4 running on android 11.

Closes #485 
Closes #1019 
Closes #1046

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -33,7 +33,8 @@
         <service
             android:name="dev.patrickgold.florisboard.FlorisImeService"
             android:label="@string/floris_app_name"
-            android:permission="android.permission.BIND_INPUT_METHOD">
+            android:permission="android.permission.BIND_INPUT_METHOD"
+            android:directBootAware="true">
             <intent-filter>
                 <action android:name="android.view.InputMethod"/>
             </intent-filter>
--- a/app/src/main/java/dev/patrickgold/florisboard/FlorisApplication.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/FlorisApplication.kt
@@ -17,6 +17,12 @@
 package dev.patrickgold.florisboard
 
 import android.app.Application
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import android.content.IntentFilter
+import android.os.Build
+import androidx.core.os.UserManagerCompat
 import dev.patrickgold.florisboard.crashutility.CrashUtility
 import dev.patrickgold.florisboard.debug.Flog
 import dev.patrickgold.florisboard.debug.LogTopic
@@ -28,6 +34,8 @@ import dev.patrickgold.florisboard.ime.theme.ThemeManager
 import dev.patrickgold.florisboard.res.AssetManager
 import dev.patrickgold.florisboard.res.FlorisRef
 import timber.log.Timber
+import java.lang.Exception
+import java.lang.ref.WeakReference
 
 @Suppress("unused")
 class FlorisApplication : Application() {
@@ -64,6 +72,35 @@ class FlorisApplication : Application() {
         } catch (e: Exception) {
             CrashUtility.stageException(e)
             return
+        }
+
+        /*Register a receiver so user config can be applied once device protracted storage is available*/
+        if(!UserManagerCompat.isUserUnlocked(this) && Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            registerReceiver(BootComplete(), IntentFilter(Intent.ACTION_USER_UNLOCKED))
+        }
+    }
+
+    fun init() {
+        CrashUtility.install(this)
+        val prefs = Preferences.initDefault(this)
+        val assetManager = AssetManager.init(this)
+        SubtypeManager.init(this)
+        DictionaryManager.init(this)
+        ThemeManager.init(this, assetManager)
+        prefs.initDefaultPreferences()
+    }
+
+    private class BootComplete : BroadcastReceiver() {
+
+        override fun onReceive(context: Context?, intent: Intent?) {
+            if(Intent.ACTION_USER_UNLOCKED == intent?.action){
+                try {
+                    (context as FlorisApplication).unregisterReceiver(this)
+                    context.init()
+                } catch (e : Exception) {
+                    e.fillInStackTrace()
+                }
+            }
         }
     }
 }
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/clip/FlorisClipboardManager.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/clip/FlorisClipboardManager.kt
@@ -268,42 +268,46 @@ class FlorisClipboardManager private constructor() : ClipboardManager.OnPrimaryC
      * @param context Required to register as an onPrimaryClipChangedListener of ClipboardManager
      */
     fun initialize(context: Context) {
-        systemClipboardManager = (context.getSystemService(CLIPBOARD_SERVICE) as ClipboardManager)
-        systemClipboardManager.addPrimaryClipChangedListener(this)
+        try {
+            systemClipboardManager = (context.getSystemService(CLIPBOARD_SERVICE) as ClipboardManager)
+            systemClipboardManager.addPrimaryClipChangedListener(this)
 
-        val cleanUpClipboard = Runnable {
-            if (!prefs.clipboard.cleanUpOld) {
-                return@Runnable
+            val cleanUpClipboard = Runnable {
+                if (!prefs.clipboard.cleanUpOld) {
+                    return@Runnable
+                }
+
+                val currentTime = System.currentTimeMillis()
+                var numToPop = 0
+                val expiryTime = prefs.clipboard.cleanUpAfter * 60 * 1000
+                for (item in history.asReversed()) {
+                    if (item.timeUTC + expiryTime < currentTime) {
+                        numToPop += 1
+                    } else {
+                        break
+                    }
+                }
+                for (i in 0 until numToPop) {
+                    history.removeLast().data.close()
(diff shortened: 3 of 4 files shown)


Glitchy-Tozier:
I feel like that might add a permission that many users might not need... What are the things an app can do using this functionality?

thestinger:
It's a "normal" permission, not a "dangerous" permission. It's not supposed to be user-facing and does not have any user-facing privacy or security implications. An app can already start itself after boot other ways.

thestinger:
According to https://developer.android.com/training/articles/direct-boot#notification, you're supposed to use `ACTION_USER_UNLOCKED` if the app can have foreground processes rather than only background ones. It seems like that's the right approach here.

https://developer.android.com/reference/android/content/Intent#ACTION_USER_UNLOCKED

thestinger:
Doesn't seem that requires the boot completed permission per documentation but I haven't checked. If it doesn't, it's one of countless examples of why that permission is meaningless since you can listen for other broadcasts or start themselves other ways.

Here's the upstream explanation of the permissions:

https://developer.android.com/reference/android/Manifest.permission#RECEIVE_BOOT_COMPLETED

> Allows an application to receive the Intent.ACTION_BOOT_COMPLETED that is broadcast after the system finishes booting. If you don't request this permission, you will not receive the broadcast at that time. Though holding this permission does not have any security implications, it can have a negative impact on the user experience by increasing the amount of time it takes the system to start and allowing applications to have themselves running without the user being aware of them. As such, you must explicitly declare your use of this facility to make that visible to the user.

It's important for Android developers to understand this about all the "normal" permissions. They do not have a real privacy or security model / goal but rather are primarily a way to either reduce attack surface, help with static analysis or make developers think twice before doing something.

This applies to things like the low-level Bluetooth permission too. The security model comes from user consent per pairing and via the Location permission for doing scanning (not required for pairing) in a way that reveals location information (Android 12 will add a form of scanning not requiring it, but BT pairing without Location has been possible for quite a while, despite apps preferring to be invasive even today). Anyway, that's just how "normal" permissions work. They do not mean what people interpret them to mean...

thestinger:
Doesn't seem that requires the boot completed permission per documentation but I haven't checked that. If it doesn't, it's one of countless examples of why that permission is meaningless since you can listen for other broadcasts or start themselves other ways. Anyway, that permission is meaningless and is simply a way to discourage starting stuff at boot because it slows it down and is quite invasive when too much stuff is doing it. It's not a permission gating whether an app can start itself in the background. The only way to stop that is force stop. Not having that permission is far from equivalent to being force stopped at boot, even though people wrongly interpret it that way. The app can still start itself.

Glitchy-Tozier:
Sounds like it could be acceptable, thank you for the explanation!

Let's see what Patrick thinks when he gets around to going through the pull requests.

empratyush:
@thestinger  
"does not have any user-facing privacy or security implications"
People still freak out for unknown reason.

"you're supposed to use ACTION_USER_UNLOCKED"
Actually there is two boot complete constant (LOCKED_BOOT_COMPLETED and BOOT_COMPLETED) and by design BOOT_COMPLETED get called only when device is unlocked.
anyway applied for suggestion.

@Glitchy-Tozier 
I do understand your hints about adding a permission is not good. 
So implementation has been improved.

thestinger:
@empratyush It may still be better to listen for it via the manifest in case there's some kind of race.

patrickgold:
Thanks for your PR, I really appreciate it! I've read a bit into the direct boot topic and it is a bit more complicated than I thought, especially due to the device vs credential protected storage thingie. I've looked at your changes (haven't tried them out yet), but on first look I think the broadcast receiver should be registered in the manifest, it just looks out of place how it is currently. I also have to think of how to handle the fact that no user-specific prefs and data is available at boot mode and possibly adjust the core managers here. I will experiment myself with this topic and will update you what I will do to support direct boot.

empratyush:
> I've read a bit into the direct boot topic and it is a bit more complicated than I thought, especially due to the device vs credential protected storage thingie.

At first glance it seems complicated. but it's actually very easy and straightforward. although it's depend on what we want to achieve. If the goal is to support direct boot there is no need to care about keeping/moving user settings/preferences into device protected storage. In fact gboard does the same what i have done. like it shows gboard default UI before user unlock for the first time and then initialize preferences once device credential protacted storage is available.

> I've looked at your changes (haven't tried them out yet), but on first look I think the broadcast receiver should be registered in the manifest, it just looks out of place how it is currently.

Actually initial implementation was via manifest, but i saw people started complaining about it. 

So thought a bit about what i am trying to do and just to support direct boot there is no need to register it via manifest. like if user select it as default keyboard it gonna work (with default ui configuration) even when device is not unlocked yet . But if this is not default keyboard it will not gonna start at boot nor gonna run unless device protacted storage is not avaiilabe or user select it as keyboard manually.

In short what i am trying to say is current implementation is about supporting direct boot just to improve user experience while other ways might be about enforcing direct boot (like on a alarm clock app where you do want to have alarm on time).

If you want i can revert the previous changes.

>I also have to think of how to handle the fact that no user-specific prefs and data is available at boot mode and possibly adjust the core managers here. I will experiment myself with this topic and will update you what I will do to support direct boot.

okay, afaik few apps move there required data to device protected storage while other does not. (Clock & alarm apps does and Gboard doesn't).

If you do decide how you want to get it done let me know, i will try to do it that way.

patrickgold:
Ok so while experimenting I discovered that [this attribute](https://github.com/florisboard/florisboard/pull/1021/files#diff-7fa6aef292187a049f7a4d6060d8df3ba212d838789c78940bd363344b1c38cdR37) is the only magic thing for direct boot on a basic level (ignoring user specific config for now). As soon as this attribute is set, an app/service can launch immediately after boot, even when locked. That's surprisingly easy... The broadcast receivers in the manifest are only then necessary if your service isn't automatically starting on boot (like alarm etc), an IME is started by the system so nothing necessary here. Good now I think I understand what's happening here.

I think I can go with your PR, because it essentially adds support for direct boot and that's the most important thing. In general I am also for a default config keyboard while in this state like Gboard does it too, no complicated stuff with moving some configs to the device protected storage. The only thing we have to account for is what happens when the password is lets say in Greek while the default config only shows the QWERTY keyboard? This may cause some trouble, but I will handle this after merging your changes in first.

As for the default behavior, I just tested it out and it works flawless. Could you just fix the code formatting (especially the spaces before/after keyboards) [here](https://github.com/florisboard/florisboard/pull/1021/files#diff-218242fbe4440ca956d8f9b3c88c32f76ad66f7cc09bc202625a7f436b9a1ab6R64-R75) and [here](https://github.com/florisboard/florisboard/pull/1021/files#diff-0fe7fbbafbe86de35da6458603dfeceeff4a17f75c49b24f24f56aa1f7f81a50R78-R105)? If that's fixed I can merge your changes in.

thestinger:
You can probably choose a sensible layout based on the locale rather than needing any device encrypted data of your own.

empratyush:
> Ok so while experimenting I discovered that [this attribute](https://github.com/florisboard/florisboard/pull/1021/files#diff-7fa6aef292187a049f7a4d6060d8df3ba212d838789c78940bd363344b1c38cdR37) is the only magic thing for direct boot on a basic level (ignoring user specific config for now). As soon as this attribute is set, an app/service can launch immediately after boot, even when locked. That's surprisingly easy... The broadcast receivers in the manifest are only then necessary if your service isn't automatically starting on boot (like alarm etc), an IME is started by the system so nothing necessary here. Good now I think I understand what's happening here.

That's what i was saying.



> As for the default behavior, I just tested it out and it works flawless. Could you just fix the code formatting (especially the spaces before/after keywords) [here](https://github.com/florisboard/florisboard/pull/1021/files#diff-218242fbe4440ca956d8f9b3c88c32f76ad66f7cc09bc202625a7f436b9a1ab6R64-R75) and [here](https://github.com/florisboard/florisboard/pull/1021/files#diff-0fe7fbbafbe86de35da6458603dfeceeff4a17f75c49b24f24f56aa1f7f81a50R78-R105)? If that's fixed I can merge your changes in.

sorry about that, i am really really bad with formatting especially spacing. although i have fixed that. 
So you can now merge these changes. (hopefully)

patrickgold:
Perfect, thanks for fixing! Will merge your changes into master, should be available for use in beta06!

PR #1091, opened 2021-07-23
Keyboard data logic rework

### LINKED ISSUE #1045: Turkish only characters: Ğ and her buddy İ
Turkish has capitalized I with dot on it "İ" letter. 
i, İ, ı, I.
on florisboard capitalized İ shows as I like in english. nothing wrong with Ğ letter I just put on title to take attention lol

### LINKED ISSUE #1089: Duplicate hints for international phonetic alphabet
#### Short description
The hints double, showing two circles instead of one. See sreenshots

#### Environment information
- FlorisBoard Version: 0.3.13-beta6
- Device: Samsung Galaxy S9
- Android: 10

patrickgold:
This PR reworks the KeyData logic so popups can also fully use variation and case selectors. This is required to fully implement the Turkish layout, but also may be needed in the future. This rework also allows for a more general use of the KeyData interface and unifies the text and emoji data base (kinda).

Additionally fixes #1045 and fixes #1089

### DIFF
--- a/app/src/main/assets/ime/text/characters/extended_popups/tr.json
+++ b/app/src/main/assets/ime/text/characters/extended_popups/tr.json
@@ -29,7 +29,10 @@
         "main": { "$": "auto_text_key", "code":  287, "label": "ğ" }
       },
       "i": {
-        "main": { "$": "auto_text_key", "code":  305, "label": "ı" },
+        "main": { "$": "case_selector",
+          "lower": { "code":  305, "label": "ı" },
+          "upper": { "code":   73, "label": "I" }
+        },
         "relevant": [
           { "$": "auto_text_key", "code":  303, "label": "į" },
           { "$": "auto_text_key", "code":  236, "label": "ì" },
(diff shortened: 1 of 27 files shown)


Glitchy-Tozier:
I may be wrong, but the dotted-circle-fix seems a bit hacky. Why do you have to check that the circle already exists? Shouldn't the script automatically only add one circle?

patrickgold:
> I may be wrong, but the dotted-circle-fix seems a bit hacky. Why do you have to check that the circle already exists? Shouldn't the script automatically only add one circle?

I wouldn't call it "hacky" per se, just a quick solution to the problem that some layout json files do define the dotted circle for the circumflex marks but others don't. Also if someone in the future adds some circumflex characters, it doesn't matter if dotted circles are included in the json label or not, it will always be rendered correctly.

Glitchy-Tozier:
> that some layout json files do define the dotted circle for the circumflex marks but others don't.

Ah, in that case it makes perfect sense!  :)

Issue #140, opened 2021-01-09
[Question] Subtype language selection


bushrang3r:
Great to see a new opensource keyboard on the scene 👍

I have installed Japanese as a language on my phone but am unable to add it as a subtype in FlorisBoard settings. Ir simply doesn't show as an option.
I wonder if I'm missing something, or if it's not possible at the moment.
I need this to be able to display the language selector button on the keyboard.
Thanks

patrickgold:
No, besides Arabic and Persian FlorisBoard does currently not support any non-latin languages. FlorisBoard's set of languages will definitely expand, but I think that adding Japanese won't happen until after v0.4.0. I will add it to the corresponding milestone.

bushrang3r:
Thanks Patrick,
Understood. No worries. And thanks.
I actually already have a Japanese keyboard installed but switching between keyboards on Android is just not as quick as  switching languages 'on the fly' with the language switching button. But I can live with it for now.

lethallunatic:
@bushrang3r 
I have reported an issue #168 to @patrickgold regarding the spacebar for switching languages. You can however use the spacebar to change to another keyboard, depending on the other keyboard you may also switch back to florisboard.

MTSYuki:
A japanese input would be wonderful.
Currently I use Mozc for Android. It's UI the same as Gboard, also open source but is not updated since 4/10/19.

Is gonna be a hard work to put suggestions, japanese language use a high number of characters.

Hayleia:
I just tried Mozc to get an idea of how this is supposed to work (I'm not a Japanese language user at all). From what I see, the layout is Latin, Latin letters are converted to hiragana only, and if the user intended to write katakana or kanji, then there is a suggestion bar at the top to convert hiragana to katakana/kanji?

The part where latin characters are converted to hiragana is doable pretty easily now (see video (yes i'm definitely typing in canadian french)), but the part where hiragana can be converted to katakana/kanji from the smart bar would probably require autocorrect/spellcheck to be stable and then it would only be a matter of using the correct dictionary I guess?
https://user-images.githubusercontent.com/7904776/119452735-f9194a80-bd36-11eb-88e8-8c5962478b46.mp4

MTSYuki:
@Hayleia That exactly how the Mozc works, yes, but there another layout for typing.

One of this is the 12 Keys, which is very similar to older phones (The only diference is the way you type the characters, you need to slide, instead of pressing the same button)

[...]

PR #1109, opened 2021-07-27
Add Japanese JIS layout

Waelwindows:
Implements the Japanese JIS layout and corresponding language features in FlorisBoard. This PR does not implement an IME as it's out-of-scope.

This PR should be a step towards closing #140.


patrickgold:
Ah I have the issue: The JSON (de)serializer is a bit inflexible for polymorphic classes, and all sub classes have to be defined here: https://github.com/florisboard/florisboard/blob/9d364f99e251ec9d2531f8f6a2d9be372247a5a2/app/src/main/java/dev/patrickgold/florisboard/res/AssetManager.kt#L54-L69

As your selectors only inherit from `AbstractKeyData`, just add them to the `polymorphic(AbstractKeyData::class) { ... }` block and it should work.

Waelwindows:
That seems to fix the JSON issue. However, `KanaSelector::compute` seems to be crashing FlorisBoard over due to `AbstractMethod` error, which I assume is due to some evaluator not implementing `ComputingEvaluator::evaluateKanaType` and `ComputingEvaluator::evaluateKanaSmall`.

Got any ideas where that may be?

patrickgold:
The computing evaluator for the text context is defined here, and I think the Kana part is missing: https://github.com/florisboard/florisboard/blob/9d364f99e251ec9d2531f8f6a2d9be372247a5a2/app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt#L118

What I don't understand though is why it compiled in the first place.

Waelwindows:
I added overrides for the evaluator in `TextInputManager` but the problem still continues.

Here's the relevant part from the crash report
```
java.lang.AbstractMethodError: abstract method "dev.patrickgold.florisboard.ime.keyboard.KanaType dev.patrickgold.florisboard.ime.keyboard.ComputingEvaluator.evaluateKanaType()"
	at dev.patrickgold.florisboard.ime.keyboard.KanaSelector.compute(KeyData.kt:154)
	at dev.patrickgold.florisboard.ime.text.keyboard.TextKey.compute(TextKey.kt:38)
	at dev.patrickgold.florisboard.ime.text.keyboard.TextKeyboardView.computeKeyboard(TextKeyboardView.kt:850)
	at dev.patrickgold.florisboard.ime.text.keyboard.TextKeyboardView.onLayout(TextKeyboardView.kt:784)
	at android.view.View.layout(View.java:22844)
	at android.view.ViewGroup.layout(ViewGroup.java:6389)
	at android.widget.LinearLayout.setChildFrame(LinearLayout.java:1829)
```

patrickgold:
Found the source: the computing evaluator here https://github.com/florisboard/florisboard/blob/9d364f99e251ec9d2531f8f6a2d9be372247a5a2/app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKeyboardView.kt#L70

does not define the kana evaluation methods, thus it crashes.

> Isn't there a way to have a default implementation for interfaces in Kotlin?

Yes, write ` = value` (where value is the default value) after the interface fun definition.

app/src/main/assets/ime/text/characters/jis.json
```diff
@@ -0,0 +1,212 @@
...
+          "small":  { "code": 12353, "label": "ぁ" }
+        },
+        "kata": {
+          "normal": { "code": 12450, "label": "ア" },
+          "small":  { "code": 12449, "label": "ァ" }
+        },
+        "half_kata": {
+          "normal": { "code": 65393, "label": "ｱ" },
+          "small":  { "code": 65383, "label": "ｧ" }
+        }
```
patrickgold, line 29: You need to add a `"$": "kana_size_selector"` here for each key within the hira, kata and half_kata, as you expect an `AbstractKeyData` and have to give the JSON deserializer a hint what data structure you have

Waelwindows:
Ok, i've added in the implementations to the evaluators and now it doesn't crash :tada:

app/src/main/assets/ime/text/characters/jis.json
Waelwindows: Yeah figures why that key wasn't working.

patrickgold:
The commit shows for me, though it is the first where you use an email address matching your GitHub it seems, as before the profile pic was blank. This happens sometimes, but as long as the commit history and working tree is upright, nothing to worry about too much (especially because we will probably squash the commits at the end anyways).

Waelwindows:
Yeah probably.

Anyways, how does your JSON logic work with optionals? It seems to be bugging out when I omit `KanaSizeSelector.small` even tho it's optional.

Waelwindows:
I'm thinking of adding a key for ゛ and ゜ and making the character small (for the future 12 flick layout) like how mozc and Google Japanese Input does.

Image for refrence
IMG_20210728_013742


Is there logic with the keycode numbers in [KeyCode.kt](https://github.com/florisboard/florisboard/blob/master/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyCode.kt) or am I supposed to choose any number?

Waelwindows:
For some reason, the latest commit where I renamed a bunch of layouts broke the pop-ups.

patrickgold:
> Anyways, how does your JSON logic work with optionals? It seems to be bugging out like the whole json layout is invalid when I omit KanaSizeSelector.small even tho it's optional.

Normally it should just hide the key if the computed key is null, but if it does not work it may bug out.

> Is there logic with the keycode numbers in KeyCode.kt or am I supposed to choose any number?

Are those Unicode characters? If so, use their respective Unicode code (https://unicode-table.com/en/ helps finding the code)

> For some reason, the latest commit where I renamed a bunch of layouts broke the pop-ups.

Will look into the changes, maybe I can find what goes wrong.

patrickgold:
The extended popup mapping for Japanese can't be found, and [this](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS40LjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwianNDb2RlIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJpbXBvcnQgamF2YS51dGlsLkxvY2FsZVxuXG5mdW4gbWFpbigpIHtcbiAgICBwcmludGxuKExvY2FsZShcImphXCIsIFwiSlBcIiwgXCJqaXNcIikudG9MYW5ndWFnZVRhZygpKVxufSJ9) is the problem. The `toLanguageTag()` does something weird and thus the string mapping mismatching. Your language tag is the correct one for me though, so we have to see how we fix the locale to language tag conversion problem.

Waelwindows:
> Are those Unicode characters? If so, use their respective Unicode code (https://unicode-table.com/en/ helps finding the code)

No, they're not unicode characters, i'm talking about the internal keycodes like `KeyCode.SPACE`. 

I'm thinking of adding new internal keycodes for  ゛゜ and making the character small for use in the compositor instead of the combining marks I've put in the layout for now.

 I currently have the kana state change based on `activeState.caps` and `activeState.capsLock` for now until I get the keys working.

patrickgold:
> KeyCode.SPACE

Uses the Unicode code `32`, is only mapped in KeyCode because it is also an often times special key, but also a normal key at once. If they aren't Unicode characters though then use a negative number, I'd suggest `-301` and downwards.

> I currently have the kana state change based on activeState.caps and activeState.capsLock for now until I get the keys working.

Ah that's what the new keys are for, I've wondered why these were mapped to the caps key currently.

As for the locale issue, I am currently working on a solution and may do a PR later on.

patrickgold:
#1117 adds a new FlorisLocale and with this the language tag is now consistent, thus fixing the popup bug. When above PR is merged into master, you must rebase your branch on the latest master commit and the extended popup mapping file should be found and loaded properly.

Waelwindows:
I've rebased master into my branch, but the popup problem still persists. Could you perhaps try to print the resolved locale by `FlorisLocale`? Perhaps there's a mistake on my end...

patrickgold:
Ok I've tested it locally and the initial problem that the layout file hasn't been found is now resolved. The extended popups file is loaded (Thus FlorisLocale is not the problem), though the mappings are not associated correctly with the keys and thus no key has a popup. Ill further investigate into this.

patrickgold:
This line causes the trouble: https://github.com/florisboard/florisboard/blob/8989b7130af26e0e52c184bb23ff35c4d16b9620/app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKey.kt#L67

The problem is that for Latin-based languages this is required, though for Japanese some weird lower-casing stuff happens and converts the character. I am currently not sure how to fix this, except maybe making the keyboard decide if it needs upper/lower-casing or not.

Another issue I see is that (at least for me) some popup characters are just boxes, meaning my device cannot render all characters correctly at all. Either this means my device font does not support these or the code gets messed up somewhere in the processing chain.

Waelwindows:
> Another issue I see is that (at least for me) some popup characters are just boxes, meaning my device cannot render all characters correctly at all. Either this means my device font does not support these or the code gets messed up somewhere in the processing chain.

That's normal, I've added certain characters that either relatively new or don't have a font glyph for them yet for posterity's sake.

Anyways, how should we deal with spaces? Traditionally, all CJK languages, AFAIK, use the full width/ideographic space (U+3000). Should this be mapped as a different keycode in `KeyCode.kt`?

patrickgold:
> I'm not sure why it's required, but if it that's the case you could probably stop converting for any non-latin locale.

Probably, some solution in that direction will be the way to go.

> That's weird, could you give out an example of it's output? Perhaps I could just hardcode it into the popups

The line I linked above is responsible for finding the correct popup mapping for a key and in there the lower-casing went wrong, this is what I meant. The layout itself has only correct output, so nothing wrong here.

> Anyways, how should we deal with spaces? Traditionally, all CJK languages, AFAIK, use the full width/ideographic space (U+3000). Should this be mapped as a different keycode in KeyCode.kt or somehow special cased with the normal regular space?

Should be mapped as a different key code, as it is a different key in the end.

Waelwindows:
Ok, i've added the ideographic space as `KeyCode.CJK_SPACE` in `KeyCode.kt`. For all intents and purposes, it's just a space so it should behave exactly like the regular `KeyCode.SPACE`. I'll try to implement it's behaviour in every that mentions `KeyCode.SPACE` in FlorisBoard.

Waelwindows:
I've just added the functionality for the kana switcher, but for some reason it's a bit laggy. The keyboard state doesn't seem to update untill you do something like press backspace. :thinking: 

Anyways, I wanted to ask you on how I can implement the popups for the kana switcher modifier. I want it to function like these keys in the layout.
Screenshot_20210731-183208_Element

Except the user could explicitly choose between each kana state they want instead of cycling through each one till you reach the desired one. i.e. instead of those three icons the user could choose the kana state directly (かな、カナ、ｶﾅ)

patrickgold:
> Ok, i've added the ideographic space as KeyCode.CJK_SPACE in KeyCode.kt. For all intents and purposes, it's just a space so it should behave exactly like the regular KeyCode.SPACE. I'll try to implement it's behaviour in every that mentions KeyCode.SPACE in FlorisBoard. Edit: I tried implementing, but i'm not sure if in certain situations the CJK_SPACE should act like a regular SPACE. I think it's better if @patrickgold decides when to and how to implement this.

Will look into it when it makes sense that CJK_SPACE should behave like SPACE

> I've just added the functionality for the kana switcher, but for some reason it's a bit laggy. The keyboard state doesn't seem to update untill you do something like press backspace. thinking

If you need to dispatch the current state immediately, call `florisboard.dispatchCurrentStateToInputUi()`. This will propagate the new state to the whole input UI and should cause it to be updated as soon as possible, without any other key press needed.

> Anyways, I wanted to ask you on how I can implement the popups for the kana switcher modifier. I want it to function like these keys in the layout.

The popups for the settings/emoji/one-handed are defined in the default extended popup mapping file (https://github.com/florisboard/florisboard/blob/master/app/src/main/assets/ime/text/characters/extended_popups/%24default.json). Note the special identifier `~left`, which means the key within group `1`. If you want to place the kana switcher popups directly where the settings/emoji etc icons are, then use `~left`, else use the respective label. If you want icons instead of the traditional label within the popup, you need to modify the popup foreground computation method: https://github.com/florisboard/florisboard/blob/d1dd91d5c44e87772a0186c16f551985035e4ed1/app/src/main/java/dev/patrickgold/florisboard/ime/popup/PopupManager.kt#L82-L147

When you have a separate kana switcher base key and it needs an icon instead of the label, you need to modify the foreground computation in TextKeyboardView.kt: https://github.com/florisboard/florisboard/blob/d1dd91d5c44e87772a0186c16f551985035e4ed1/app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKeyboardView.kt#L1097-L1213

Waelwindows:
I've added an icon for the kana switcher button and added the relevant lines in `TextKeyboardView.kt` but the icon doesn't show up on the keyboard :thinking:

app/src/main/assets/ime/text/characters/mod/jis.json
```diff
@@ -0,0 +1,27 @@
...
+      { "code": -202, "label": "view_symbols", "type": "system_gui" },
+      { "$": "variation_selector",
+        "default":  { "code":   44, "label": ",", "groupId": 1 },
+        "email":    { "code":   64, "label": "@", "groupId": 1 },
+        "uri":      { "code":   47, "label": "/", "groupId": 1 }
+      },
+      { "code":  -210, "label": "language_switch", "type": "system_gui" },
+      { "code":  -213, "label": "switch_to_media_context", "type": "system_gui" },
+      { "code":    32, "label": "空白" },
+      { "code": -9710, "label": "かな" },
```
patrickgold, line 22: The switcher is a system button, but the type attribute is not defined, so it defaults to `characters`, which does not allow for showing icons.

```suggestion
      { "code": -9710, "label": "かな", "type": "system_gui" },
```

patrickgold:
In `TextKeyboardIconSet.kt`, you must add your drawable to the `RES_ICON_IDS` array. When doing this changes, the icon shows up, but the scaling of the SVG path data is completely messed up, thus the icon is super small. You may need to adjust the SVG path data so the icon is centered within the 24x24 viewport box.

Waelwindows:
Ok, i've scaled the icon so it fits (This should really be in `CONTRIBUTING.md` as well) and added the relevant entries so it shows up.

> Ok I've tested it locally and the initial problem that the layout file hasn't been found is now resolved. The extended popups file is loaded (Thus FlorisLocale is not the problem), though the mappings are not associated correctly with the keys and thus no key has a popup. Ill further investigate into this.

I've already added the relevant groupID and stuff in the extended popup file, but I can't test it yet till you fix this

> If you need to dispatch the current state immediately, call  `florisboard.dispatchCurrentStateToInputUi()`. This will propagate the new state to the whole input UI 
and should cause it to be updated as soon as possible, without any other key press needed.

I already use `florisboard.dispatchCurrentStateToInputUi()`, in fact, twice! Yet it still doesn't fix the problem.

https://github.com/florisboard/florisboard/blob/786a4864d34d706c66f2a8c1d46a4ca8124db74b/app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt#L653-L687

Other than that, we should fix the popups, perhaps add CJK punctuation marks, and then the layout would be complete.

patrickgold:
I will look at the dispatch state method tomorrow, because that's what wonders me the most what goes wrong here. As for the popup fix, will also see to fix this soon.

patrickgold:
I got the selector working, though there are several changes, so instead of manually commenting I generated a patch file and attached it below. Just remove the`.txt` after downloading (GitHub doesn't like patch file extensions) and apply the patch with 
`git am < 0001-Add-kana-attributes-to-keyboard-state.patch` and the selector cycles through the kana types. The popup issue is still there though, will fix it now.

[0001-Add-kana-attributes-to-keyboard-state.patch.txt](https://github.com/florisboard/florisboard/files/6915185/0001-Add-kana-attributes-to-keyboard-state.patch.txt)

patrickgold:
Workaround for Japanese popups + syntax fix in your extended popups json file:

[0002-Add-workaround-for-Japanese-popups.patch.txt](https://github.com/florisboard/florisboard/files/6916028/0002-Add-workaround-for-Japanese-popups.patch.txt)

Waelwindows:
The popups still aren't showing up on my end :thinking:. Thanks for the patches though, they really help :grin:. 

I'll go ahead and add the rest of the functionality for the kana small button (not needed for JIS layout, but for future layouts).

patrickgold:
> The popups still aren't showing up on my end thinking

For me all popups show up, except the ones defined in `~left` for some reason. Will investigate.

patrickgold:
Are the popups in `~left` supposed to be the popups of the comma key or should they be the popups of the kana switcher key?

Waelwindows:
> Are the popups in `~left` supposed to be the popups of the comma key or should they be the popups of the kana switcher key?

Ideally there should be a group just for the kana switcher key, but for now. I'm currently using `~left` for the kana switcher as the popup choices for the dot aren't used that much in Japanese.

Waelwindows:
> > The popups still aren't showing up on my end thinking
> 
> For me all popups show up, except the ones defined in `~left` for some reason. Will investigate.

False alarm, turns out your patch did fix the issue. I installed the new version but didn't remove the old layout (evident by the region being `[X]`). So the layout itself was a bit bugged till i properly replaced it.

app/src/main/assets/ime/text/characters/extended_popups/ja-JP-jis.json
```diff
@@ -0,0 +1,249 @@
...
+      }
+    },
+    "uri": {
+      "~right": {
+        "main": { "code": -255, "label": ".com" },
+        "relevant": [
+          { "code": -255, "label": ".gov" },
+          { "code": -255, "label": ".edu" },
+          { "code": -255, "label": ".org" },
+          { "code": -255, "label": ".net" }
```
SchnWalter, line 244: These seem to be for US. JP probably doesn't need `.gov` and `.jp` is missing.

Also, `.edu` is not a useful suggestion for most people on this planet:

> Since 2001, new registrants for second-level domain names have been required to be United States–affiliated institutions of higher education.  
> https://en.wikipedia.org/wiki/.edu

Waelwindows: Yeah, these don't make sense since there's no QWERTY input in this JIS layout. I should probably remove them.

Waelwindows:
I've added a group for the kana types. and I've confirmed it works. However, for some reason, it won't wonk on the kana switcher key. :thinking:

Waelwindows:
I have some MVP for how the idea would work, Currently i'm using a sentinel character to trigger the small kana transformation. Here's how it looks like.

]

https://user-images.githubusercontent.com/5228243/128430882-7a2194b8-62cd-442d-a103-b56c4da988dc.mp4

We should figure out how the final implementation should be coded like.

patrickgold:
> I've added a group for the kana types. and I've confirmed it works. However, for some reason, it won't wonk on the kana switcher key. thinking.

You need to add the `KeyCode-KANA_SWITCHER` to the `exceptionsForKeyCodes` array in PopupManager.kt, then the popups will work like defined.
As for the key group ID for kana, would it be possible to use something higher, like `91` instead of `4`? I want to use the lowest numbers to implement key grouping for some theme options later on. If you want to choose another ID this is also ok, it must be lower than `100` though (>=100 is custom IDs defined by the user).

> We should figure out how the final implementation should be coded like.

Does the kana_small toggle key modify an already written key in the text box?

Waelwindows:
> > We should figure out how the final implementation should be coded like.
> 
> Does the kana_small toggle key modify an already written key in the text box?

Yes, just like in the video. That's how most mobile IMEs I've seen do it.

Waelwindows:
The last thing left is how to handle CJK space should be handled and maybe adding CJK punctuation.
@patrickgold any progress on the space?

patrickgold:
> The last thing left is how the CJK space should be handled

Question: Where exactly are you using CJK_SPACE? As of now, I cannot find any uses in the code base and thus I don't know how to write the special handling stuff. My best guess would be that it replaces the SPACE key bar, but I am not exactly sure and want to ask for clarification first before I start working on it.

>  I would like to get this PR merged before the final release for v3.13 is out

That's definitely doable, v0.3.13 is still at least 2 beta releases away, more likely 3 though

Waelwindows:
> > The last thing left is how the CJK space should be handled
> 
> Question: Where exactly are you using CJK_SPACE? As of now, I cannot find any uses in the code base and thus I don't know how to write the special handling stuff. My best guess would be that it replaces the SPACE key bar, but I am not exactly sure and want to ask for clarification first before I start working on it.

Yes, it should act exactly like the normal `SPACE` does. Including the fact that it displays the current language in the key if it's in `character` otherwise an icon should be shown. (future proofing for a future 12 flick key layout)

patrickgold:
Below patch file is what I think all necessary implementation so `CJK_SPACE` works like `SPACE`. The only exception is currently for the audio effect but due to #1142 I cannot edit this without creating merge conflicts, so for now the space bar sounds like a character.

[0001-Add-CJK_SPACE-special-behavior-to-input-logic.patch.txt](https://github.com/florisboard/florisboard/files/6951356/0001-Add-CJK_SPACE-special-behavior-to-input-logic.patch.txt)

Waelwindows:
Great, the patch works fine, just as expected. :+1:

patrickgold:
Perfect, great to hear!

Now from my side this what still needs to be done:
- Improving the display for layouts with more than 4 rows (this is an issue for quite some non-Latin layouts, so it needs to be done anyways)
- Final code review and testing that everything works like it should

From you besides the small refinements (like CJK punctuation) you are still doing I wound need the following please:
- Squashing your commits into 1-3 commits (so it makes sense feature wise), because ~38~ 39 commits is quite a lot.
- Testing if the Japanese characters all output the correct character (if not already done), as I am a non-Japanese speaker and this is quite difficult for me to test :)

Waelwindows:
> Squashing your commits into 1-3 commits (so it makes sense feature wise), because ~38~ 39 commits is quite a lot.

I'll do this when i'm absolutely done with the branch. I know I'm supposed to be using 
```sh
git rebase -i {commit hash}
```
After which, should I force push to my branch ? ~~I thought GitHub automatically handled squashing...~~

 
> Testing if the Japanese characters all output the correct character (if not already done), as I am a non-Japanese speaker and this is quite difficult for me to test :)

I have confirmed that all kana input is correct and working. So this shouldn't be a problem.
I'll go over it one more time ~~, just in case.~~

Waelwindows:
There's something that's been bothering me about the punctuation layout. Specifically the the comma and period. It's almost like they're hard-coded in and I can't change them nor their popups. Could you look into this @patrickgold?

patrickgold:
Normally instead of merging master into your branch to update the base you update the local master branch with the latest commits from remote, then within your feature branch you issue:

```
git rebase master
```

If there are merge conflicts, then you'd fix them and do `git rebase --continue` (I think, but git prints the exact command in the terminal). After this, you force-push the branch and GitHub updates the PR.

Now as you've merged master in several times, rebasing shouldn't be performed in this branch anymore.

> There's something that's been bothering me about the punctuation layout. Specifically the the comma and period. It's almost like they're hard-coded in and I can't change them nor their popups. Could you look into this @patrickgold?

Like you also did for the JIS layout, you need to add custom modifier layouts for symbols/symbols2 layouts. There, you can change out the comma and period.

Waelwindows:
I've added an option to the composer that allows for "sticky" key transformation. It's currently just a public variable in the class. 
https://github.com/florisboard/florisboard/blob/687de2cd6059f3b211c3d2f8f074d7cf37fae9e0/app/src/main/java/dev/patrickgold/florisboard/ime/text/composing/KanaUnicode.kt#L13
How should we expose this option to the user?

Waelwindows:
I've removed `half_kata` from `KanaSelector` and made a new selector called `WidthSelector` instead because other CJK languages (like Korean) also have half-width characters. Now what's left is to decouple width from the other kana stuff.

Waelwindows:
I've refactored the code base to remove all mentions of `KanaType` and `half_kata` and the code works just fine :partying_face:.

I tried implementing the popup that previously was half_kata to use `multi_key_text` but it didn't work. The codes by themselves work just fine tho.

patrickgold:
>  How should we expose this option to the user?

What exactly is this sticky mode used for? I don't really understand its use and thus it is difficult for me to find an implementation for it in the UI.

> I tried implementing the popup that previously was half_kata to use multi_key_text but it didn't work. The codes by themselves work just fine tho.

I think you misunderstood what the `multi_text_key` is here for: You use it to execute two keys at once, while the intended and implemented use allows for characters with multiple code points (e.g. emojis) to be defined (but it is still 1 character and 1 action at most).

Below patch file is me improving the half-width implementation and naming, updating the keyboard state docs, re-introducing the KANA_HALF_KATA key which sets both the isKanaKata and isCharHalfWidth flag and cleaning up some parts of new code. With this the kana switcher now cycles through smoothly again, like before. I've renamed the width selector to `char_width_selector`, as this makes it more clear which width we are changing. Your implementation of the half width key stays in there, but for now no layout uses it (as in your particular case it makes more sense to add a separate key for this).

[0001-Improve-half-width-implementation-Clean-up-code-base.patch.txt](https://github.com/florisboard/florisboard/files/6961101/0001-Improve-half-width-implementation-Clean-up-code-base.patch.txt)

Waelwindows:
> but for now no layout uses it (as in your particular case it makes more sense to add a separate key for this).

Yeah, I'll probably add a width switcher key for the symbol layouts.

Waelwindows:
> > How should we expose this option to the user?
> 
> What exactly is this sticky mode used for? I don't really understand its use and thus it is difficult for me to find an implementation for it in the UI.

The sticky keys option in `KanaUnicode` is to select which method to deal with combining marks. With `sticky` is set to off, the combining marks act like a toggle, i.e. pressing the same combining mark twice toggles the mark on the character). When `sticky` is set to on, pressing the same combining mark won't toggle the mark, it'll just do nothing.

As you can see, this option would rarely need to be changed. I feel like it would best to have a window for composer options so that user could configure the composer with the options they would like. Perhaps maybe a button to a window/tab next to the composer option in the layout window?

patrickgold:
> The sticky keys option in KanaUnicode is to select which method to deal with combining marks. With sticky is set to off, the combining marks act like a toggle, i.e. pressing the same combining mark twice toggles the mark on the character). When sticky is set to on, pressing the same combining mark won't toggle the mark, it'll just do nothing.

Ah I understand now, thanks for explaining.

> As you can see, this option would rarely need to be changed. I feel like it would best to have a window for composer options so that user could configure the composer with the options they would like. Perhaps maybe a button to a window/tab next to the composer option in the layout window?

Yes, this would be a solution I agree. Is the sticky behavior very important that one can choose or can it wait until I get to re-write the Settings UI (and add preferences for composers)? Else I'd suggest either choosing the more common one or adding a temporary shadow composer which has sticky behavior, while the original doesn't.

Waelwindows:
> Is the sticky behavior very important that one can choose or can it wait until I get to re-write the Settings UI (and add preferences for composers)?

No, it's not that important. the current defaults are sufficient, this is all just in for customization purposes. So, it can wait for that instead. We'll just merge the PR first then expose the customization options in UI for `KanaUnicode` later

GoRaN909:
> Perfect, great to hear!
> 
> Now from my side this what still needs to be done:
> 
> * [ ]  Improving the display for layouts with more than 4 rows (this is an issue for quite some non-Latin layouts, so it needs to be done anyways)
> * [ ]  Final code review and testing that everything works like it should

Hi @patrickgold  i wana know when this improvements will available for more than 4-row layouts that fixes my big problem fot central kurdish now the 5-row layout is not supported on florisboard I'm still waiting to this bug get fix :(

patrickgold:
> No, it's not that important. the current defaults are sufficient, this is all just in for customization purposes. So, it can wait for that instead. We'll just merge the PR first then expose the customization options in UI for KanaUnicode later

Ok, then I'll add this to the Settings UI rework.

> Hi @patrickgold i wana know when this improvements will available for more than 4-row layouts that fixes my big problem on Central Kurdish 5-row layout is not supported on florisboard I'm still waiting to this bug get fix :(

Before the v0.3.13 release, as it is required for the Japanese layout (from my viewpoint). Probably in the same beta as the Japanese or one beta earlier, will see.

Waelwindows:
@patrickgold could you look into why these icons aren't showing up? [Here](https://github.com/florisboard/florisboard/files/6987169/switcher_icons.zip) the sample SVG file in case. You could preview the char width switcher in the CJK symbols/symbols2 layout

app/src/main/assets/ime/text/symbols/mod/cjk.json
```diff
@@ -0,0 +1,55 @@
...
+          }
+        },
+        "half": { "code": 65380, "label": "､", "popup": {
+            "main": { "code":   44, "label": "," }
+          }
+        }
+      },
+      { "code": -205, "label": "view_numeric_advanced", "type": "system_gui" },
+      { "code": 12288, "label": "空白" },
+      { "code": -9701, "label": "char_width_switcher", "popup": {
```
patrickgold, line 27: It's the key type once again, character keys are not allowed to render symbols, while other types do:

```suggestion
      { "code": -9701, "label": "char_width_switcher", "type": "system_gui", "popup": {
```

This is just one key I suggested the change, although you have to do this for all switcher keys, also when they are in the popup

Waelwindows:
I've changed the the type and added the entries for the 2 char width states to `PopupManager.kt`. However, the icons appear too big. Is there something we could about it?

patrickgold:
>  Squashed commit of the following: 

May I ask what this commit's purpose is? With this commit you include changes on the main branch on your PR manually, which can mess up the main branch history once merged.

Waelwindows:
Oh, it's just a merge of main. If you don't want it, I could just rebase and use a normal merge instead

Waelwindows:
I'm looking into the composer and something must've happened between 7f9b0e46 and master's HEAD because the composer now bypasses composing marks and the small sentinel character. Weird :thinking:

patrickgold:
> Oh, it's just a master merge. If you don't want it, I could just rebase and use a normal merge instead

Thanks for fixing it, looking at the PR changes and now only your changes occur except my master changes and your changes, this makes the final code review easier.

As for the composer bug, I have changed the lowercase logic in TextKey.kt, but other than that I haven't touched anything which should affect the composer in any way.

patrickgold:
> https://github.com/florisboard/florisboard/blob/f6ae545183c7c84c447152d237591214a3be278f/app/src/main/java/dev/patrickgold/florisboard/ime/text/keyboard/TextKey.kt#L58

You merged in master and while fixing the merge conflict forgot to add the workaround for Japanese (if the subtye's language code is "ja" then do not lowercase, else do it).

Replace this line with the following and it should work again:

```kt
    val subtype = evaluator.getActiveSubtype()
    val computedLabel = if (subtype.locale.language == FlorisLocale.from("ja").language) {
        computed.label
    } else {
        computed.label.lowercase(evaluator.getActiveSubtype().locale)
    }
```

patrickgold:
Another question for the squash commit after you are done and after the final code review: Do you want to squash the commit and force push this branch or should I squash everything into 1 commit via the GH UI and merge it this way? For me personally it doesn't matter, just want to clarify this.

Waelwindows:
@patrickgold AFAIK wasn't that workaround for the popups to appear? Since the layouts work fine without the workaround now, it should be fine to keep it the way it is.

As for the merge to master. I believe that using the github UI would be best. It keeps commit history on my end, and it just shows up as a single commit on your end

patrickgold:
> AFAIK wasn't that workaround for the popups to appear? Since the layouts work fine without the workaround now, it should be fine to keep it the way it is.

The workaround isn't needed anymore for the popups? Huh. If it works without it them I would also say let's keep it this way.

As for the composer: is there now an issue or is it resolved?

> As for the merge to master. I believe that using the github UI would be best. It keeps commit history on my end, and it just shows up as a single commit on your end

K, will use that when merging this feature then.

Waelwindows:
I'm not currently on my dev station so i cannot work on it this second. But, yes, there's an issue with KanaUnicode compositor. The compositor wasn't working at all. I managed to narrow it down the a minimal example. The expected behaviour for this test compositor is to add "em" for normal keys and add "emDaku" for the daku (゛) and other composing keys. Current behavior is printing em, but adding the composing keys and sentinel small character without any processing. 

It would help if you could look into it for the mean time, till I get back to my dev station

patrickgold:
Will do! I've just done multiple compares with only my changes and my changes + your changes to see what has changed which could have broken the composer and haven't found anything. Will try it out later on my local machine and debug it.

I will do the beta09 release now, your PR should be included in next weeks beta10 so it can be fully featured in the 0.3.13 stable release.

patrickgold:
> The compositor wasn't working at all. I managed to narrow it down the a minimal example. The expected behaviour for this test compositor is to add "em" for normal keys and add "emDaku" for the daku (゛) and other composing keys. Current behavior is printing em, but adding the composing keys and sentinel small character without any processing.

Ok just tried it out locally and if I press a normal key then "em" is printed, if I press the daku (゛) "emDaku" is added. If I understood correctly this is the correct behavior of your test, isn't it?

Besides the composer thing, I reviewed the symbols key and great job on the design of the full/half-width switcher key! It is so satisfying seeing the A grow and shrink while many characters adapt with them :)

Waelwindows:
Ok, so i tried it again on my device and it didn't work till I completely uninstalled Florisboard DEV and reinstalled it using `./gradlew installDebug`. I double checked the layout definition on my end and it was the same. So, I'm guessing it was an environment bug of some sort on my end. 

Anyways, the compositor works just fine so I reverted the test compositor code. And while checking out the keyboard, I found a bug in the compositor! Turns out I switched over some letters in the small kana portion by accident :sweat_smile:

patrickgold:
> Ok, so i tried it again on my device and it didn't work till I completely uninstalled Florisboard DEV and reinstalled it using ./gradlew installDebug. I double checked the layout definition on my end and it was the same. So, I'm guessing it was an environment bug of some sort on my end.

Ah a classic on-device code cache bug. Happens rarely, is annoying but fixable like you did. Great to hear that it works again!

A small note should you rebase your PRR on master: you need to initialize all submodules correctly, else the app build will always fail. Can be done by issuing `git submodule update --init --recursive`

Is this PR now ready from your side? If so I'd do the final code review/test and then merge it into master.

Waelwindows:
I'd say so, I don't have anything left on my end except maybe rework the kana switcher icon

patrickgold:
Ok, just let me know when you say the code is final. Should be this week though or else it will have trouble making it into 0.3.13, as everything for the stable release feature-wise must also be in the last beta release, which will be done this Sunday.

Waelwindows:
Ok, the PR is done from my end. You can review the code now :tada:

patrickgold:
Thanks for informing me, will do the code review and final test in my time-zone's tomorrow

app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
```diff
@@ -667,6 +673,65 @@ class TextInputManager private constructor() : CoroutineScope by MainScope(), In
         }
     }
 
+    /**
+     * Handles a [KeyCode.KANA_SWITCHER] event
+     */
+    private fun handleKanaSwitch() {
+        activeState.isKanaKata = !activeState.isKanaKata
+        florisboard.dispatchCurrentStateToInputUi()
+    }
```
patrickgold, line 682: What happened to the kana switcher? It does not cycle through the modes anymore. Based on commit 0ac30ce746135a8fedcec2c84f25eaccd95717cc I can see you wanted to remove half_kata from the cycle. The problem is, that you don't set the char width back to full width, so now there are two possible cycles possible:
- Hira -> Kata (What you want)
- Hira -> Half kata (happens if I was previously in the symbols layout and the char width is still on half width.

Should this be unintended, I suggest this change:

```suggestion
    private fun handleKanaSwitch() {
        activeState.isKanaKata = !activeState.isKanaKata
        activeState.isCharHalfWidth = false
        florisboard.dispatchCurrentStateToInputUi()
    }
```

app/src/main/assets/ime/text/characters/jis.json
```diff
@@ -0,0 +1,354 @@
+{
+  "type": "characters",
+  "name": "jis",
+  "label": "JIS",
+  "authors": [ "Waelwindows" ],
```
patrickgold, line 5: Here you specify your user name in upper case, while in most other layout files it is in lower case. Although this is just a cosmetic issue, it should be consistent across all files (especially because in 0.3.14 with the new Settings UI the author tag will be visible). You can also use the upper case variant in all files, just make sure it is the same in all files.

app/src/main/assets/ime/text/numeric/row/cjk.json
```diff
@@ -0,0 +1,108 @@
+{
+  "type": "numeric_row",
+  "name": "cjk",
+  "label": "CJK",
+  "authors": [ "patrickgold" ],
```
patrickgold, line 5: Nice to put me as the author but I think you did the great work here :)

app/src/main/assets/ime/text/symbols/cjk.json
```diff
@@ -0,0 +1,285 @@
+{
+  "type": "symbols",
+  "name": "cjk",
+  "label": "CJK",
+  "authors": [
+    "waelwindows"
+  ],
```
patrickgold, line 7: Cosmetic issue: the authors line should be on one line like in all other asset files too.

patrickgold (changes requested):
Thanks again for your great work here, really appreciate it. I've tested it out locally and it works great!

This review contains one logic bug (maybe is also intended, I am not exactly sure) and a few cosmetic issues. Once these are addressed I can merge your changes into master, so it will be included in tomorrow's beta10 release and Japanese will be in the stable 0.3.13 release at the end of next week.

app/src/main/java/dev/patrickgold/florisboard/ime/text/TextInputManager.kt
Waelwindows: Yeah, i didn't catch that. Previously that function delegates to the other kana type functions which dealt with the character width

app/src/main/assets/ime/text/characters/extended_popups/ja-JP-jis.json
```diff
@@ -0,0 +1,292 @@
...
+      "~kana": {
+        "relevant": [
+          { "code": -9711, "label": "あ" },
+          { "code": -9712, "label": "ア" },
+          { "code": -9713, "label": "ｱ" }
+        ]
+      },
+      "~right": {
+        "main": { "$": "char_width_selector",
+          "full": { "code": 12289, "label": "、" },
```
Waelwindows, line 224: Switch this over to a katakana middle dot ・

app/src/main/assets/ime/text/symbols2/cjk.json
```diff
@@ -0,0 +1,211 @@
...
+        "full": { "code": 65372, "label": "｜", "popup": {
+          "main": { "code":  124, "label": "|" }
+          }
+        },
+        "half": { "code":  124, "label": "|", "popup": {
+          "main": { "code": 65372, "label": "｜" }
+          }
+        }
+      },
+      { "code": 12539, "label": "・", "popup": {
```
Waelwindows, line 40: Use character width selector

patrickgold (approved):
Thanks for fixing the mentioned issues! From my side this PR is ready to be merged, I just don't know if there's still a layout issue because of your two review comments.

patrickgold:
As I have to finish beta10 now, I will squash merge in your changes. Thanks again for your contribution, really appreciate it!

Waelwindows:
Sorry, couldn't add the fixes in time because I was out of the house. In fact those reviews were from my phone :sweat_smile:. I hope you can still add the last commit however.

patrickgold:
The beta10 APK is already built and fully prepared locally, I cannot add these two minor things into beta10 anymore. As this is a small fix though, I can still include it in the final 0.3.13 release (I allow minor bug fixes between the last beta and a stable release), if you open a fix PR.

patrickgold:
As I have to do another minor bug fix and make an additional beta11 release tomorrow, I can even squeeze in your small adjustments in the beta track if you make a patch PR until tomorrow.

RickyM7:
@Waelwindows Sorry for talking about an already merged PR, but why wasn't the wo/o(を) character added to the layout?

Waelwindows:
@RickyM7 it's there as a pop up for わ, just like the layout on PC.

RickyM7:
> it's there as a pop up for わ, just like the layout on PC.

Got it! I've only used the 12-key layout on my phone so far, so I'm slowly adapting to this new layout (and I'm really enjoying it by the way). Thanks for the answer.

PR #1147, opened 2021-08-12
Possibly fix repeating delete key

### LINKED ISSUE #986: Repeating input
#### Short description
Clicked once, repeated until pressed again. Is this a new feature? 

#### Steps to reproduce
1. Type things for a while
2. Catch the bug occur

#### Environment information
- FlorisBoard Version: 0.3.13 beta-3
- Install Source: Fdroid
- Device: Asus Zenfone Max pro m1
- Android: 9 stock

patrickgold:
This PR is a possible fix for the repeating delete/arrow key bug a lot of people reported. Thorough testing needs to be done though before we can say for sure it is fixed.

Closes #986

### DIFF
--- a/app/src/main/java/dev/patrickgold/florisboard/ime/core/InputEventDispatcher.kt
+++ b/app/src/main/java/dev/patrickgold/florisboard/ime/core/InputEventDispatcher.kt
@@ -20,12 +20,12 @@ import android.os.SystemClock
 import android.util.SparseArray
 import androidx.core.util.forEach
 import androidx.core.util.set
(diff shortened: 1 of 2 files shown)


GoRaN909:
I think the bug was fixed in this debug apk now I'm using it the repeating is not happened after  5 hours ago of using till now.

Thank you :)

jeremiah-miller:
Seems fixed to me too

patrickgold:
Thanks for both of you to confirm that it is fixed! I will still let it open for a bit of time and should no report occur that it isn't fixed, then I will merge it.

PR #1153, opened 2021-08-14
Adjust popups for Arabic main forms and remove prebaked numbers.

### LINKED ISSUE #1087: Arabic numerals are hardcoded into the Arabic layout's popups
In most keyboards that support Arabic, Persian, etc.., the default behaviour of long pressing a letter is getting its most frequently used variant first, not the symbol that resides in its place on the symbols layer.. For example,
image
as opposed to,
image
I think maintaining that expectation might be a good idea. As such variants are used much more often than the current symbols.

Waelwindows:
This commit makes the most common popups the main ones which should
allow Arabic sub-layout users to use FlorisBoard's smart popup feature.

Closes #1087

### DIFF
--- a/app/src/main/assets/ime/text/characters/extended_popups/ar.json
+++ b/app/src/main/assets/ime/text/characters/extended_popups/ar.json
@@ -4,59 +4,21 @@
   "authors": [ "HeiWiper" ],
   "mapping": {
     "all": {
-      "ض": {
-        "relevant": [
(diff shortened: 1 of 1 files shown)


patrickgold:
Thanks for handling this issue! I've looked at the mapping file and to me everything looks good. Also when testing it out on a testing device the popups make much more sense now, without the duplicate numerals.

I will merge your changes into master now, will be in tomorrow's beta09 release!

PR #1177, opened 2021-08-23
fix: Add katakana middle dot to JIS popups

Waelwindows:
Just polish to the layout


patrickgold (approved):
Thanks for the fixes. I've locally tested them out and everything works fine. Will add it to the beta11 release.

PR #1212, opened 2021-08-30
Add Igbo character layout

nd500:
Igbo language character layout (Note: also added an update for config.json)

PR #1213, opened 2021-08-30
Update config.json

nd500:
Updated config.json to include Igbo language (also added character layout for Igbo)


patrickgold:
See #1212.

PR #1214, opened 2021-08-30
Add qwerty row symbols for number row users..

Aman9das:


patrickgold:

I see what you are trying to achieve, the problem is you are hard-coding the 1st row into the symbols layout, this is not intended to be this way. I agree that Gboard shows more symbols when the number row is enabled, though this requires some rework of the layout merge logic and the characters you include as an extension row. If you want you can file this as a feature proposal, then I (or someone else) can have a look in the future to implement it.

Aman9das:
I think neo2 does hard code the symbols, and works well.. if we set 'hint is prioritized' option.. the smart prioritize version is a bit off i agree
If possible could you add it for beta..? If it doesnt work you could revert it afterall

patrickgold:
I am not a fan of merging the hard-coded row into the default western symbol layout, but if you copied the default symbols layout, give it a distinct name/label and add the hard-coded row there I have no problem with accepting that. For you to be able to use it you'd just need to select the "other" western symbol layout you added in the subtype editor.

Aman9das:
Yes.. you are right, i've made the changes now..

Could you tell how to add telex to a Layout in the contributing guide, phonetic indic keyboards need it for example ch, and chh are different letters.. it would be easier to add support them..

patrickgold:
Thanks for your changes! I've reviewed them online and fixed the naming of the layout (Was not synced with the json metadata, which causes the layout to fail). Tomorrow I will test it on my local test device and if everything checks out squash merge your changes in!

PR #1236, opened 2021-09-07
Add ñ as popup on catalan layout

hegocre:
Let me explain this. Catalan people like me also speak Spanish, and we usually use both languages at the same time while typing, so it comes very handy to have ñ on the Catalan keyboard (even though it's not on the alphabet). Keyboards like GBoard already do this.

Thanks!


patrickgold:
Thanks for your PR! If you feel a `ñ is a relevant popup for Catalan, then I have nothing against adding it! Have reviewed your changes and will merge them in now.

hegocre:
Thank you! Also for making this great keyboard.

PR #1254, opened 2021-09-15
fix '>' symbol

Aman9das:
Extremely sorry :(

I messed up the symbols-additional layout


patrickgold:
No worries mate, such things happen all the time and are extremely easy to fix. Will merge your fix in, will be available in 0.3.14-beta02!

PR #1274, opened 2021-09-20
Adding Warang Citi layout

Singkiring57:
Hello team,

I wanted to add Warang Citi layout for Ho language. I have added the layout. But I don't really understand how capitalization works for Latin, Cyrillic and Greek, Warang Citi characters also have two cases. Warang Citi characters are mapped for upper and lower cases in Unicode, http://www.unicode.org/charts/PDF/U118A0.pdf but I'm not sure if case switching will work as expected.


tsiflimagas:
You don't need to do anything for capitalization, it's handled internally from a class. You only need to configure the lower characters, as you did (except if there's some special case I'm not aware of).

Singkiring57:
> You don't need to do anything for capitalization, it's handled internally from a class. You only need to configure the lower characters, as you did (except if there's some special case I'm not aware of).

Thank you for the information, capitalization works exactly like in Latin, there are no special cases.

patrickgold:
Thanks for your PR! I've reviewed your changes, made some small fixes (syntax in config.json and naming of the characters layout), but in general great work, really appreciate it!

> but I'm not sure if case switching will work as expected.

I just tested it out on my testing rig and it seems that the automatic capitalization does its job just fine, so no special cases needed.

Will merge your changes in, will be included in 0.3.14-beta03!

Singkiring57:
@patrickgold Thank you for reviewing and merging!!

PR #1288, opened 2021-09-27
Update warang_citi.json for upper case 𑣖

Singkiring57:
Added missing line `"$": "auto_text_key"` for letter 𑣖.


patrickgold:
Thanks for your contribution. I will merge your changes once #1290 is in, which will still be a few days.

I had a quick look over your changes and noticed that you added comments to the JSON file - the JSON syntax does not allow for this at all and the FlorisBoard parser will not be able to parse your layout with these comments there. May I ask you to remove them? Else on first look everything looks fine!

PR #1306, opened 2021-10-06
Remove slurs from glide dictionary

pnppl:
Last thing you want to do is accidentally text someone the n-word.


Glitchy-Tozier:
I think this should be optional because
1. Who decides what words we ban?
2. It may be difficult to maintain for all languages.
3. I dislike the idea of an app stopping me from writing the words I may want to write.

sabzo:
And I’m saying slurs shouldn’t be optional
________________________________

tsiflimagas:
Imo, being polite is not a keyboard's job and it should be optional in its functionality (it already is so for word suggestions), as it is irl. If how the keyboard works is becoming an obstacle for someone wishing to avoid such words, then removing them can probably be a solution, but I regard it as a last resort. Ι see Tozier's first point, but it probably doesn't apply in this case (at least as far as I can see), since this PR pertains to a specific set of words, not only about being insulting, so I believe (and hope) it wouldn't get out of hand and start removing every swear or anything someone might not like. But even in this case, this is problematic: `Gaylord` is a name, `homo` means human in Latin and is used in phrases like `homo erectus`, `dyke` or `dike` means ditch. See, sometimes words are like a knife: you can use it to hurt or to get a job done. Same goes for how you regard them. Sometimes, if one wants to feel insulted, he will anyway. So, what's left to be subjectively justified to remove (as I see it) is the n word, which, though, some groups of people use without some negative implication (I'm not judging whether they're right to do so, I'm mentioning it as a fact).

sabzo:
It's not a keyboard's job to be polite, it's a people's job as keyboards don't make decisions. Someone kindly provided an update to the keyboard to avoid insensitive racial slurs. So why not simply accept it and move on?

tsiflimagas:
I think I made my thoughts clear. Your reply is too simple and short to answer to what I've written, I think. Also, having objections on things also resonates with the ethos of opensource. One might see something like a must-have, while another might see it very differently. What I'm sure about is that no one should be deterred from expressing their thoughts.

Glitchy-Tozier:
> Say what one may,
> 
> The priority of who're should be above whore.
> 
> I don't think anyone would have any complaint against this. They use the same pattern..

Sure ... unless the user uses "whore" in every second sentenc while never using "who're". Then, the keyboard should learn 😉

---

Regarding @tsiflimagas last message: I completely agree. I wouldn't want to be banned from typing certain words, whether it be in latin, in a quote, or just for comedic purposes.

---

Also, I really don't like the thought of the best open-source keyboard suddenly banning users from typing certain words. As an optional add-on to spell-checking, fine, but not as a mandatory filter that targets all users.

tsiflimagas:
I agree @Aman9das

sabzo:
Wonderful, that's the response I hoped for and expected.

tsiflimagas:
Thanks for your response @patrickgold! It seems that there were more to consider since more things are about to change. To me it's not about agreeing or not, it's about being able to discuss freely, without people getting angry with the discussion itself, so thank you for how you responded.

PR #1413, opened 2021-11-29
Add Hebrew niqqud and some punctuation marks

amiryal:
Opening this initially as a draft, to gather feedback while I research how to deal better with the combining characters, which don’t show nicely in the diff and in the source code.


amiryal:
OK, changed the problematic Unicode characters to `\uXXXX` escapes.

patrickgold:
Thanks for your PR!

Just checked the changes and everything is ok, will merge them in now. Will be released later today in 0.3.15-beta01

PR #1445, opened 2021-12-14
Fix typo in strings.xml

Spiryt7:
Found a small typo in strings.xml

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -442,7 +442,7 @@
 
     <string name="setup__finish_up__title">Finish Up</string>
     <string name="setup__finish_up__description_p1">{app_name} is now enabled in the system and ready to be customized by you.</string>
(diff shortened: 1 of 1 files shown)


Glitchy-Tozier:
Honestly, both seem fine to me. "Should" sounds somewhat elegant 🤔

zeGolem:
> Honestly, both seem fine to me. "Should" sounds somewhat elegant thinking

Main issue I can see is that this kind of wording may confuse non-native English speakers, because it's not the "conventional" use of "should"

Glitchy-Tozier:
I see what you mean. However, in my opinion, we can expect people that don't have sufficient knowledge of the English language to have Android set to a language they properly understand.

That being said, let's see what Patrick thinks, once he gets around to looking at this PR. :)

patrickgold:
For me as a non-native English person deciding what is better in this particular instance is extremely difficult. I tried looking this up on the internet and grammatically both are correct, only that "if" is less formal than "should".

But I think going for "if" is the better option here, because it reads a bit better and more fluent (and it is also shorter so better for smaller device screens). Will merge this change in after #1440 is merged, so it will be part of beta08.

PR #1510, opened 2022-01-16
Kotlinx serialization from 1.3.1 to 1.3.2

BFK78:


patrickgold:
Thanks for your PR!

Dependency upgrades are normally only done manually by me, because it requires checking if everything still works as expected and this is mostly done in batches (at the beginning of each release development cycle). As I planned to upgrade serialization soon anyways, I will close your PR and will do that in my batch upgrade. Hope you can understand.

PR #1619, opened 2022-03-06
Update Urdu phonetic layout

blueberry-robin:
2 fixes for wrong layout in the Urdu phonetic keyboard. One was a wrong(duplicate) HTML code causing 2 different keys to output the same character. The other was a wrong label on a key resulting in 1 key label appearing twice and giving 2 different outputs.


patrickgold:
Thanks for your changes! Just checked them and will merge them into master. Will be available in 0.3.14-beta14 later today!

PR #1620, opened 2022-03-06
Fix Kurdish language not change app language

GoRaN909:
Fix Kurdish language not changing the app language manually + some correction to kurdish layout


patrickgold:
@GoRaN909 Does that mean that I used the incorrect language codes for Kurdish? If so, I need to adjust the https://github.com/florisboard/florisboard/blob/master/crowdin.yml file too, else my bot will override your string resource file changes in its next iteration.

patrickgold:
I just added it for you (because I need to release beta14 soon). Everything else checks out fine, will merge your changes in once the automatic check passes. Thanks for your PR!

GoRaN909:
> I just added it for you (because I need to release beta14 soon). Everything else checks out fine, will merge your changes in once the automatic check passes. Thanks for your PR!

Thank you so much for correct language code from yaml file that is the problem of crowdin i contacted them more than a time to correct it but no one replied me the kmr is correct code or kurmanj kurdish and ckb for central kurdish language but the problem is Android OS doesn't recognise it just recognise  KU for kurmanji and ckb for central kurdish defined on AndroidOS.

patrickgold:
> kmr is correct code or kurmanj kurdish and ckb for central kurdish language but the problem is Android OS doesn't recognise it just recognise KU for kurmanji and ckb for central kurdish defined on AndroidOS.

Ok yeah that's what I gathered in below table too:

Code | Name | What Android Says it is
-- |-- | -
ku | Kurdish | Kurdish
kmr | Kurmanji (Kurdish) | (Unknown to Android)
ckb | Sorani (Kurdish) | Central Kurdish

As you are speaking native Kurdish, is there any difference between Kurdish and Kurmanji? Because Crowdin separates these two languages, so there must be some kind of difference (maybe dialect?)

GoRaN909:
> > kmr is correct code or kurmanj kurdish and ckb for central kurdish language but the problem is Android OS doesn't recognise it just recognise KU for kurmanji and ckb for central kurdish defined on AndroidOS.
> 
> Ok yeah that's what I gathered in below table too:
> 
> Language Code	Crowdin Name	Android Name
> ku	Kurdish	Kurdish
> kmr	Kurmanji (Kurdish)	(Unknown to Android)
> ckb	Sorani (Kurdish)	Central Kurdish
> As you are speaking native Kurdish, is there any difference between Kurdish and Kurmanji? Because Crowdin separates these two languages, so there must be some kind of difference (maybe dialect?)

Actually It's same language but two different dialect the KU code is general language code you can use it for all dialects but  on Android it was used for the kurmanji dialect and ignored the kmr for it why google did that idk! 
the ku code is correct and defined on windows/linux or macOs for example Microsoft is used ku for Central Kurdish and kmr for the Kurmanji dialect to the windows language pack and Microsoft translator look [language support](https://docs.microsoft.com/en-us/azure/cognitive-services/Translator/language-support), The difference between Sorani and Kurmanji is writing system the Sorani is write with perso-arabic script (RTL) but the kurmanji is using latin script there is no big different on talking we can talk to each other and understand everything 😅

GoRaN909:
Hi @patrickgold a few minutes ago i just got an update of FlorisBoard 0.3.14-beta14 (70) at Google Play i tried to manually change app language from Advanced settings, It's not changing to any available languages not just for my native language when click on any other languages its just English and doesn't changed to other.

Here is the screen recording:


https://user-images.githubusercontent.com/3160786/157136982-a00b9d47-ce94-46d4-b443-44b96ee1068b.mp4

patrickgold:
Hmm so no language change does occur? Generally it should though, as seen on my device:

https://user-images.githubusercontent.com/19412843/157182242-bb5d6c5b-dad0-4650-bcd4-a72b6eba6b78.mp4

If you set the language to Kurdish, then exit the app (wait 1 to 2 secs before exiting) and restart the app, in which language is the UI then?

patrickgold:
On that note I just noted that for Central Kurdish the UI layout should change to RTL, shouldn't it?

hegocre:
I think the problem might be caused by Google Play split apks. Google Play is only downloading resources for English, so when you select another language, the locale is not installed and it fallbacks to the default locale. Happened to me before, there are two solutions:
* Download the locales on demand by using Google Play APIs (I don't think it's a good idea)
* Add a build gradle flag on the `android` section to not split the locales (might slightly increase app size)
```
bundle {
    language {
        enableSplit = false
    }
}
```
This can only be tested with Google Play releases, so it can be tested either by trying on each beta release or by creating an internal test release.

patrickgold:
Thanks @ hegocre for the hint, Google Play does indeed split up resources, [as seen in the docs](https://developer.android.com/guide/app-bundle/configure-base#disable_config_apks). As I don't want to rely on any runtime Google Services (namely the `SplitManager` in this case), I think disabling the language split is the best option.

hegocre:
> add this line to build.gradle.kts should fix the language not change bug.

@GoRaN909 I think this would only work for the specified locales, not for all of them. The only way of including all of them in the app is disabling the split.

patrickgold:
@GoRaN909 could you please check #1629 and give feedback in that PR if the RTL layout works and if I messed up the RTL layout? Thanks in advance!

> I think this would only work for the specified locales, not for all of them. The only way of including all of them in the app is disabling the split.

Yup you are correct, this is why #1629 disables splits for bundles in the build.gradle.

GoRaN909:
> > add this line to build.gradle.kts should fix the language not change bug.
> 
> @GoRaN909 I think this would only work for the specified locales, not for all of them. The only way of including all of them in the app is disabling the split.

Yes i know but he can add all language codes that already available on FlorisBoard i wrote that five language codes for example :)

GoRaN909:
> @GoRaN909 could you please check #1629 and give feedback in that PR if the RTL layout works and if I messed up the RTL layout? Thanks in advance!
> 
> > I think this would only work for the specified locales, not for all of them. The only way of including all of them in the app is disabling the split.
> 
> Yup you are correct, this is why #1629 disables splits for bundles in the build.gradle.

Thank you so much I tested the debug apk and it's works great the layout direction  now changes to RTL when change app language to the rtl writing systems, But the clipboard manager is not changing when its disabled by default it's just English 😕


https://user-images.githubusercontent.com/3160786/157308939-0e78f12c-b4d1-48c8-b5f1-0bc2112d6779.mp4


Clipboard manager is always English when using other languages.

Screenshot_20220308-221705_Brave

patrickgold:
> I tested the debug apk and it's works great the layout direction now changes to RTL when change app language to the rtl writing systems

That's great to hear!

> But the clipboard manager is not changing when its disabled by default it's just English

The title of the PR is named "Add support for RTL preference UI layout" for a reason :) Adding RTL support in the preference UI is easy, but I can't apply the same easy process directly on the keyboard UI.

What locale should be considered when applying RTL or LTR layout direction to the UI in your opinion? The current subtype, the preference strings language (the one set in Advanced) or the system default language? Personally, the current subtype makes most sense, just because it is the language that also is used for word suggestions in the Smartbar (once word suggestions are implemented).

GoRaN909:
> What locale should be considered when applying RTL or LTR layout direction to the UI in your opinion? The current subtype, the preference strings language (the one set in Advanced) or the system default language? Personally, the current subtype makes most sense, just because it is the language that also is used for word suggestions in the Smartbar (once word suggestions are implemented).

@patrickgold Thank you for you add support RTL languages, I think the most important part that should be change to RTL layout direction is most needed settings preference like language/layout screen, keyboard settings  also the subtype add/edit screen is have a more using other parts that not implemented yet is not so important like suggestions, spell checker or user dictionary.

I hope you all of them get fixed easily just like preference UI :)

patrickgold:
So I extended #1629 and it now also contains support for RTL in the keyboard UI, based on the primary language of the active subtype. The only thing that does not work as intended is the Smartbar, which has a slight measuring issue when in RTL, but I am not sure if that is a bug from me or the UI library itself.

GoRaN909:
WoW you just nailed it 👏 the app and keyboardview UI is fully supported the RTL languages also the clipboard history manager and emoji layout too are now fully translated to it's not just English 😀 

Thank you very much i can confirm that all things that related to RTL works ike a charm the including () {} [] ...etc writing with a correct direction it's not reverse when you are using an RTL layout like my language. 

Thanks again for your great work i wish that someday can i Donate you and sponsor this cool and amazing great project well done bro love and respect for you 🤍💙👏

patrickgold:
Great to hear you like the RTL rework! Will still fine tune the RTL layout based on your feedback in #1629 and my own observations.

> Thank you very much i can confirm that all things that related to RTL works like a charm including () {} [] ...etc is writing with a correct direction it's not reverse when you are using an RTL layout like my language.

Could you elaborate further on that? I got several RTL-language speaking users who complain that the parentheses are inserted in the incorrect direction when in RTL. 

So if I am in RTL mode and I press `(`, should `(` or `)` be inserted? I got several issues which all say something different, and now I am confused what's like the actual correct direction.

GoRaN909:
> So if I am in RTL mode and I press `(`, should `(` or `)` be inserted? I got several issues which all say something different, and now I am confused what's like the actual correct direction.


Oh sorry they were right i tried ( ) when my device language was English but when change language to RTL such ( Central Kurdish, Arabic ) when you press `(` it insert reversed this 👉 `)` it should be insert this 👉 `(`  when you preess ( , And insert `)` when press on ) 👈 this one

This reverse happens when phone language is an RTL languages but it's okay when device lang is English or any LTR languages

patrickgold:
Ok thanks for clarifying!

> This reverse happens when phone language is an RTL languages but it's okay when device lang is English or any LTR languages

Interesting for me it is exactly the opposite, LTR flips it and RTL doesn't. Need to experiment with this further.

PR #1629, opened 2022-03-08
Add support for RTL preference/keyboard UI layout and other minor fixes

### LINKED ISSUE #422: Parentheses flipped in Hebrew
When typing in hebrew the parentheses keys produce flipped characters of how they look in the keyboard. i.e. pressing the left parenthesis inserts a right parenthesis. This is the case for all types of parentheses.

### LINKED ISSUE #1624: Problem with characters
#### Short description

These characters are typed in reverse in Persian, please fix this problem

https://user-images.githubusercontent.com/91434001/157015028-8cc48e77-2d37-45fb-abbd-099dfa870d85.mp4

IMG_20220307_134921

#### Environment information
- FlorisBoard Version: 0.3.14-beta14
- Install Source: GitHub
- Device: xiaomi note 9 pro
- Android: Stock

patrickgold:
This PR addresses issues which came up in the discussion of #1620.

Closes #422 
Closes #1624

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -70,6 +70,15 @@ android {
         }
     }
 
+    bundle {
+        language {
+            // We disable language split because FlorisBoard does not use
+            // runtime Google Play Service APIs and thus cannot dynamically
+            // request to download the language resources for a specific locale.
+            enableSplit = false
+        }
+    }
+
     buildFeatures {
         compose = true
     }
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- Copyright (C) 2020 Patrick Goldinger
+<!-- Copyright (C) 2020-2022 Patrick Goldinger
 
      Licensed under the Apache License, Version 2.0 (the "License");
      you may not use this file except in compliance with the License.
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/phone2/telpad.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/phone2/telpad.json
@@ -1,8 +1,14 @@
 [
   [
-    { "code":   40, "label": "(" },
+    { "$": "layout_direction_selector",
+      "ltr": { "code":   40, "label": "(" },
+      "rtl": { "code":   41, "label": "(" }
+    },
     { "code":   47, "label": "/" },
-    { "code":   41, "label": ")" },
+    { "$": "layout_direction_selector",
+      "ltr": { "code":   41, "label": ")" },
+      "rtl": { "code":   40, "label": ")" }
+    },
     { "code":   45, "label": "-" }
   ],
   [
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/symbols/eastern.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/symbols/eastern.json
@@ -32,22 +32,42 @@
     { "code":   43, "label": "+", "popup": {
       "main": { "code":  177, "label": "±" }
     } },
-    { "code":   40, "label": "(", "popup": {
-      "main": { "code":64830, "label": "﴾" },
-      "relevant": [
-        { "code":   91, "label": "[" },
-        { "code":   60, "label": "<" },
-        { "code":  123, "label": "{" }
-      ]
-    } },
-    { "code":   41, "label": ")", "popup": {
-      "main": { "code":64831, "label": "﴿" },
-      "relevant": [
-        { "code":   93, "label": "]" },
-        { "code":   62, "label": ">" },
-        { "code":  125, "label": "}" }
-      ]
-    } },
+    { "$": "layout_direction_selector",
+      "ltr": { "code":   40, "label": "(", "popup": {
+        "main": { "code":64830, "label": "﴾" },
+        "relevant": [
+          { "code":   91, "label": "[" },
+          { "code":   60, "label": "<" },
+          { "code":  123, "label": "{" }
+        ]
+      } },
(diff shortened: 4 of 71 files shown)


GoRaN909:
Hi @patrickgold I tested the debug apk file many of related issues occur in #1629 was fixed but after this commit 33f23366763c059a8b23ded51bb9ab1b6516060b there are a small bug on theme editor screen when you want pick a colour the direction is reversed move your finger to left side the cursor goes to the right side and move finger to up the cursor goes to down side the navigation action was Opposite move to right is go to left move to left is go right and same thing for up and down action.

patrickgold:
@ GoRaN909 I fixed the theme editor issues, color picking should work normally again.

The Smartbar and emoji screenshots you brought up are actually intended, at least that's how other keyboards also behave in RTL.

GoRaN909:
> I fixed the theme editor issues, color picking should work normally again.

You right thanks it's okay now works without any problem.
 
> The Smartbar and emoji screenshots you brought up are actually intended, at least that's how other keyboards also behave in RTL.

@patrickgold Hmm i agree with you right the emoji palette is change to right direction when use RTL layout but you know in the old version 3.10/3.11 of FlorisBoard the emoji screen was static view that was so great if you makes like before just for Emojis i think is good idea because almost all (RTL) language  users of florisboard include me are okay and like the stable not changeable view when using Emojis :)

patrickgold:
(Hopefully) last thing to ask you on the language/layout improvements of this PR: I fixed the parentheses when typing in RTL mode, could you check if it works for you (in system language / keyboard subtype English and Kurdish)? That would be really helpful!

> Hmm i agree with you right the emoji palette is change to right direction when use RTL layout but you know in the old version 3.10/3.11 of FlorisBoard the emoji screen was static view that was so great if you makes like before just for Emojis i think is good idea because almost all (RTL) language users of florisboard include me are okay and like the stable not changeable view when using Emojis :)

I dunno though, I think there will be bug reports on why the emoji view ignores RTL by some users if I force LTR on it. I can make the bottom bar LTR-only at least (the delete and switch to text button), this would be in accordance to LatinIME based keyboard.

GoRaN909:
> (Hopefully) last thing to ask you on the language/layout improvements of this PR: I fixed the parentheses when typing in RTL mode, could you check if it works for you (in system language / keyboard subtype English and Kurdish)? That would be really helpful!
>

I checked the debug apk but sadly it's same thing parentheses still flipped on RTL layout it have no issue with English! 🤨

https://user-images.githubusercontent.com/3160786/157768789-8e2b985e-0088-4f50-8cd7-4a134a161d5b.mp4



> I dunno though, I think there will be bug reports on why the emoji view ignores RTL by some users if I force LTR on it. I can make the bottom bar LTR-only at least (the delete and switch to text button), this would be in accordance to LatinIME based keyboard.

Because we're used to it like that way LTR 😅

This is RTL look at emoji faces so annoying showing RTL direction 😬

Screenshot_20220310-182351_Samsung Notes


But this one LTR is so much better for emoji positions 🥲

Screenshot_20220311-020351_Samsung Notes

patrickgold:
> I checked the debug apk but sadly it's same thing parentheses still flipped on RTL layout it have no issue with English! 🤨

In an empty box they are flipped, because the text edit app you use assumes LTR before the first strongly typed char. So if you type a Kurdish word and then use the parentheses, the word won't be flipped, correct?

Also if you have Gboard installed you can directly compare the behaviors for RTL and will see that it will also be flipped for an empty text field at first.

> This is RTL look at emoji faces so annoying showing RTL direction 😬

Oh so you mean the emoji itself. Yes I agree, they can be forced LTR.

GoRaN909:
> In an empty box they are flipped, because the text edit app you use assumes LTR before the first strongly typed char. So if you type a Kurdish word and then use the parentheses, the word won't be flipped, correct?
> 

Yes it flipped when no Kurdish word typed but it's okay if you use it with some words like this 👉 (سڵاو چۆنی) anyway that's not a problem for me i don't care about parentheses me or maybe alot of peoples doesn't used usually it just okay thanks.


> Also if you have Gboard installed you can directly compare the behaviors for RTL and will see that it will also be flipped for an empty text field at first.
> 
It's same thing like FlorisBoard on GBoard it also flipped for RTL 😅 but i don't know how Samsung corrected reverse parentheses it's typing correctly on Samsung Keyboard 😳

The reason i don't like GBoard it only have one Kurdish layout that not a Standard form to writing in Kurdish language. 

> Oh so you mean the emoji itself. Yes I agree, they can be forced LTR.

Yup i mean Emojis it self each emoji positions equals to LTR but for icons and delete actions no problem if they RTL expect emoji faces force use LTR please.

patrickgold:
> It's same thing like FlorisBoard on GBoard it also flipped for RTL sweat_smile but i don't know how Samsung corrected reverse parentheses it's typing correctly on Samsung Keyboard flushed

If you type parentheses in an empty text box with Samsung Keyboard and then type something Kurdish, does the parentheses still align correctly?

Anyways I can currently not do much further against this, and the new behavior is at least better then no adjustment.

Also I just fixed the Emoji palette, is now forced RTL.

GoRaN909:
> > It's same thing like FlorisBoard on GBoard it also flipped for RTL sweat_smile but i don't know how Samsung corrected reverse parentheses it's typing correctly on Samsung Keyboard flushed
> 
> If you type parentheses in an empty text box with Samsung Keyboard and then type something Kurdish, does the parentheses still align correctly?
 
Yes it also type correctly on empty box too not just with Kurdish words 🙂


https://user-images.githubusercontent.com/3160786/157909953-ad6c100c-4fb4-4182-869c-17e01009607d.mp4




> Anyways I can currently not do much further against this, and the new behavior is at least better then no adjustment.

Yes it's not from your side i think the problem is parentheses doesn't have a special code at Unicode system for RTL symbols thank you for all off your workd and coding to improve RTL bugs.


> Also I just fixed the Emoji palette, is now forced LTR.

Now it's very nice thanks for this fore layout  corrections.

PR #1681, opened 2022-03-23
Correct layout name for RU&UA (ЙЦУКЕН)

VladWinner:


Underday (commented):
JCUKEN is more universal. Should stay like this.

victorbnl:
I don't think the translitteration is inappropriate as you say, wikipedia's page is also named JCUKEN. 

But I do agree that as russian users are supposed to use the layout, or at least someone who understands cyrillic, it is much more pertinent and correct to use the cyrillic version of the name.

VladWinner:
> wikipedia's page is also named JCUKEN

Basically, name of any article from the English Wikipedia is written in English 🙂
But they are not really sure which option to decide on:
image

patrickgold (approved):
Sorry for the long delay, kinda ignored this PR because I was distracted with a lot of other ongoing development.

So my take on this topic is, that using the native name has priority over using the transliteration, as this layout is supposed to be used by people who know this language and not by ones who only speak English.

`ЙЦУКЕН` applies the convention of using the first 6 characters in the first letter row. As such, I am agreeing with @VladWinner here and I am accepting this layout's label change. The internal layout ID will stay the same (`jcuken_russian` and `jcuken_ukrainian`), to prevent users from losing their subtype config when updating.

This change will be merged in now and released later today in 0.4.0-alpha1.

Issue #1682, opened 2022-03-23
Keyboard layout screen should show hints in the current settings language


SchnWalter:
The UI for adding a new keyboard layout assumes that the user knows the native spelling of the language and that the user has the apropriate keyboard to type those characters. This is not always the case, especially for those who want to learn a new language (or test pull requests like #1681)

I'm not sure what's the correct solution here, but one approach could be to add translations for the layout languages in the various languages used by the keyboard settings application. And then we could have some sort of hint for each option or use the translation as default and have the hint show the native name of the language.

I saw #1681 and I wanted to validate the changes, just to make sure that it's not a `node-ipc`-style vandalism. And while trying to validate the changes I had to use a translation service and send the native language names via an instant messaging application. And at that point I realized that I've ran into the samne issue some time ago, when I was trying to add a new keyboard layout for my language learning app.

patrickgold:
Before I can answer further on your proposal I need some clarification:

Do you mean the language names or the layout names, that are not translated? 

Also is your base for this proposal 0.3.14 or 0.3.13?

VladWinner:
@SchnWalter Is that what you mean?

```
"label": "Russian (ЙЦУКЕН)", <!-- for English-speaking users -->
"label_ru": "Русская", <!-- native name (in Russian) -->
"label_uk": "Російська", <!-- for Ukrainian-speaking users -->


"label": "Ukrainian (ЙЦУКЕН)", <!-- for English-speaking users -->
"label_uk": "Українська",  <!-- native name (in Ukrainian) -->
"label_ru": "Украинская", <!-- for Russian-speaking users -->
```

SchnWalter:
@patrickgold, no need to hurry with these changes, it's just an improvement that only helps in edge cases and we should first collect some feedback.

@VladWinner, yes, but I was thinking of dropping the `label` key in favor of `nativeLabel` or something that would denote the fact that the label MUST be in the language that it's intended to be used.
```js
{
  "id": "jcuken_russian",
  "nativeLabel": "Русская (ЙЦУКЕН)",            // instead of Russian (ЙЦУКЕН)
  "..."
}, {
  "id": "canadian_french",
  "nativeLabel": "Français canadien (QWERTY)",  // instead of Canadian French (QWERTY)",
  "..."
}, {
  "id": "esperanto_with_hx",
  "nativeLabel": "Esperanto kun 'ĥ'",           // instead of Esperanto with 'ĥ'
  "..."
}
```
[...]

PR #1688, opened 2022-03-24
Fix main popup for "ﺍ"

### LINKED ISSUE #1571: Move two characters on a Persian keyboard
In Persian, after the ( ا ) character, the ( آ ) character is the most widely used, but it is difficult to access it on the keyboard.
Please replace character ( ا ) with the specified character to make it easier to access.
IMG_20220213_172130
IMG_20220213_172619

tsiflimagas:
Fixes #1571

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/fa.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/fa.json
@@ -8,12 +8,12 @@
       ]
     },
     "ا": {
+      "main": { "code": 1570, "label": "آ"},
       "relevant": [
         { "code": 1649, "label": "ٱ" },
         { "code": 1569, "label": "ء" },
         { "code": 1571, "label": "أ" },
-        { "code": 1573, "label": "إ" },
-        { "code": 1570, "label": "آ" }
+        { "code": 1573, "label": "إ" }
       ]
     },
     "ه": {


tsiflimagas:
I wonder if the same should be done for Arabic popups. I have no idea about these languages, it's just an easy fix.

patrickgold:
Thanks for your PR!

Reviewed your changes and everything checks out fine.

> I wonder if the same should be done for Arabic popups.

Yeah that's a good question. I didn't add any non-Latin layouts, and by now they are a completely unique mix of different contributors who moved keys and popups all over the place. I think we should leave it? If they are incorrect someone will make an issue about it for sure :)

If you wanna leave it I will merge the changes in, else I will let this PR open for you to add it for Arabic as well.

> I have no idea about these languages, it's just an easy fix.

Same. There's the Unicode CLDR keyboard layout standard though (mentioned in #1682), which I may be integrating into FlorisBoard in the future. Hopefully this CLDR repo fixes a lot of these wild guesses and change requests, because it is a standard and implemented by some keyboards.

tsiflimagas:
> I think we should leave it?

I guess so. Arabic seems to be the same on both FlorisBoard and Gboard, so it should be alright.

> There's the Unicode CLDR keyboard layout standard though (mentioned in #1682), which I may be integrating into FlorisBoard in the future. Hopefully this CLDR repo fixes a lot of these wild guesses and change requests, because it is a standard and implemented by some keyboards.

This seems pretty interesting!

PR #1695, opened 2022-03-25
Add Material You support

Waelwindows:
Continued from matrix.

So far I've managed to get florisboard building on nix with `compose.material3`. Looks like i need to look into `SnyggStylesheet.kt` and maybe theme editor related stuff first. Got any clue for low hanging fruit?

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -171,6 +171,7 @@ dependencies {
     implementation("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.0")
     implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.3.2")
     implementation("androidx.room:room-runtime:2.4.2")
+    implementation("androidx.compose.material3:material3:1.0.0-alpha08")
     kapt("androidx.room:room-compiler:2.4.2")
 
     testImplementation("io.kotest:kotest-runner-junit5:5.1.0")
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,28 @@
+with import <nixpkgs> {};
+
+let
+  cmakeVersion = "3.18.1";
+
+  # Use cmakeVersion when you define androidComposition
+  env = (pkgs.androidenv.override {
+    licenseAccepted=true;
+  });
+
+  androidComposition = env.composeAndroidPackages {
+    platformVersions = [ "31" ];
+    toolsVersion = "26.1.1";
+    buildToolsVersions = [ "31.0.0" ];
+    ndkVersions = ["22.1.7171670"];
+    cmakeVersions = [ cmakeVersion ];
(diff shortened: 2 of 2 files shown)


patrickgold:
There are multiple ways we can implement Material You (=MY) into the keyboard UI. A hot file is definitely [SnyggStylesheet](https://github.com/florisboard/florisboard/blob/master/app/src/main/java/dev/patrickgold/florisboard/snygg/SnyggStylesheet.kt), just because it is the main point where the keyboard UI retrieves its colors.

In the past I had 2 different approaches in mind:
1. Similar to how adaptive themes in 0.3.13 worked, we could inject MY in the `getProperty(...)` method, which decides based on the requested element and metadata which color should be returned. This is a lot of hard coding though, so I don't know if this is the best idea.
2. You could generate a SynggStylesheet dynamically ([like here with the base style](https://github.com/florisboard/florisboard/blob/master/app/src/main/java/dev/patrickgold/florisboard/ime/theme/FlorisImeThemeBaseStyle.kt)), once on startup and then everything the dynamic system color palette changes. Then in [ThemeManager.updateActiveTheme()](https://github.com/florisboard/florisboard/blob/06cfa34a4bf350dbed34d100f8befedcb2876906/app/src/main/java/dev/patrickgold/florisboard/ime/theme/ThemeManager.kt#L117-L121) you select the dynamically generated stylesheet if MY is enabled. The cool thing is that SnyggStylesheet has a DSL builder, so you could write a relatively well looking dynamic stylesheet.

Generally I'd much prefer approach 2, just because the whole process is much cleaner. The problem with both approaches is that the current custom theme has no say in which surfaces should be painted what and atm I don't really have an idea how to integrate it into a sylesheet.json either.

Above info is just a hint, if you have a better idea you are allowed to draft it too :)

patrickgold:
> Just to get a few things clear. How can we integrate method 2 with user themes?

Yeah that's the big question here. Atm approach 2 would ignore the user theme.

What we could do is the following: If Material You is enabled AND `isMaterialYouAware` is set to `true` in the user theme, then we do the following: compile the user stylesheet, but switch out ALL colors to the dynamic colors, BUT respecting the user choices for key shapes, font sizes, shadow, etc. This would be an initial middle way approach to both accept what the user has set and to take the colors from Material You.

I wouldn't compile the stylesheet every time though, only if strictly necessary (user theme has changed or system dynamic color palette has changed)

patrickgold:
@Waelwindows do you still want to work on this feature or should we close this draft and possibly allow others to work on Material You support?

rnayabed:
Is this currently dormant? We would really love to have monet support ☺️

patrickgold:
Closing PR as per resolution on Matrix chat. Material You support will be added either by me or someone else in the future, however there is no ETA for this.

PR #1741, opened 2022-04-03
Fix Armenian layouts duplicate letter

pjtsearch:
The Eastern and Western Armenian layouts had a duplicate letter դ instead of the letter փ.  This PR fixes that issue.


patrickgold:
Thanks for your fix, everything checks out, will be included in the next release 0.3.15-rc01!

Issue #1737, opened 2022-04-02
Wrong place of two characters in Persian layout 2


Mahantor:
#### Short description
Please change the place of these two characters together. This problem was solved in Farsi 1 layout, but it exists in Farsi 2 layout.
![Screenshot_2022-04-02-09-31-57-860-edit_ir ilmili telegraph](https://user-images.githubusercontent.com/91434001/161367367-fc569b2f-112e-495e-98b9-0d0058e83346.jpg)

#### Environment information
- FlorisBoard Version: 
- Install Source: GitHub
- Device: xiaomi redmi note 9 pro
- Android: 11

patrickgold:
Can be fixed together with #1736 probably, placing it on the kbd-layout issue list.

m-koushan:
Now it's fixed. Thanks for reporting @Mahantor !

PR #1745, opened 2022-04-04
Fixing #1737 Issue 

m-koushan:
#1737 fixed

Edit: Also i fixed the order of utility key on Persian2 layout, based on SwiftKey keyboard layout
Edit: #1746 added


m-koushan:
I checked the artifact apk and there is something weird about new German layout 

image

I didn't find anything wrong in my codes, I assume that the problem is with another thing like size calculating or I don't know
I need help on this

patrickgold (changes requested):
@m-koushan thanks for the fixes but what in the holy hell did you do to the branch? Please start from a clean branch off the current master state and cherry-pick your relevant fix commits in there, then force push the new branch on here. The current state is a complete mess to say the least and I won't accept your PR in this state.

PR #1750, opened 2022-04-05
Update persian.json

Mahantor:
kashida and half_space characters are practically useless in Persian, and due to the large space occupied, when typing the spacebar and half_space buttons quickly, they are often mistakenly touched.


patrickgold:
Thanks for your changes. While the changes work, I am unsure if this change aligns with all Persian users. There's always the discussion if the key is useful or not. Is there like an official source if those two keys make sense or not (note that I am completely unfamiliar with Persian)?

Mahantor:
If I create these changes as a new layer for Persian language, will you add it to the project ???

m-koushan:
@Mahantor yes I think its better this way, close this PR and if you like make a new layout for it.

PR #1795, opened 2022-04-16
close unused fd

Sevz17:
dup2 doesn't close old fds, it only duplicates them.

### DIFF
--- a/app/src/main/cpp/utils/log.cpp
+++ b/app/src/main/cpp/utils/log.cpp
@@ -71,6 +71,7 @@ int utils::start_stdout_stderr_logger(const char *app_name) {
     pipe(pfd);
     dup2(pfd[1], 1);
     dup2(pfd[1], 2);
+    close(pfd[1]);
 
     /* spawn the logging thread */
     if (pthread_create(&thr, nullptr, thread_func, nullptr) != 0) {


Sevz17:
Does florisboard read from stdin?, if not, is possible to duplicate pfd[0] to stdin and doing something like this:
```cpp
#include <chrono>
#include <iostream>
#include <thread>
#include <fstream>
#include <unistd.h>

static void thread_func(void) {
	std::string buf;
	std::fstream f("test.txt", std::fstream::out);

	while (std::getline(std::cin, buf)) {
		f << buf << std::endl;
	}

	f.close();
}

int
main()
{
	int piperw[2];

	pipe(piperw);

	dup2(piperw[0], STDIN_FILENO);
	dup2(piperw[1], STDOUT_FILENO);
	dup2(piperw[1], STDERR_FILENO);
	close(piperw[0]);
	close(piperw[1]);

	std::thread thr(thread_func);
	thr.detach();

	for (int i = 1; i <= 10; i++) {
		std::cout << i << std::endl;
		std::cerr << "stderr" << std::endl;
		std::this_thread::sleep_for(std::chrono::milliseconds(500));
	}
}
```

patrickgold:
Sorry for the late review, I was pretty occupied with the input logic and editor caching rework, so had not too much time doing other stuff.

I am not the C++ expert, so I may be completely mistaken, but will closing the file file descriptor not cause any issues? Especially considering it is open to listen to stderr and log it?

> Does florisboard read from stdin?, if not, is possible to duplicate pfd[0] to stdin and doing something like this:

As far as I am concerned and I know no. I really must say that your code snippet looks much more cleaner and also looks more understandable to me! May be worth playing around with it to see if it could benefit the error logging in terms of maintainability and readability of the logging code.

Sevz17:
> Sorry for the late review, I was pretty occupied with the input logic and editor caching rework, so had not too much time doing other stuff.

no problem, sorry for the late reply too.
 
> I am not the C++ expert, so I may be completely mistaken, but will closing the file file descriptor not cause any issues? Especially considering it is open to listen to stderr and log it?

No, actually leaving it open _may_ block florisboard.


> > Does florisboard read from stdin?, if not, is possible to duplicate pfd[0] to stdin and doing something like this:
> 
> As far as I am concerned and I know no. I really must say that your code snippet looks much more cleaner and also looks more understandable to me! May be worth playing around with it to see if it could benefit the error logging in terms of maintainability and readability of the logging code.

In that case, if you want, I can apply a kotlin-like style (and do some refactoring) to the code (just give me the style guide).

patrickgold:
> No, actually leaving it open may block florisboard.

Hmm okay. Still don't 100% understand what happens here, may need to read more into this.

> In that case, if you want, I can apply a kotlin-like style (and do some refactoring) to the code (just give me the style guide).

There's no general code style for C++ in this project, but it mainly follows the Kotlin-defaults for spacing and brackets:

- Indent: 4 spaces
- Brackets open on same line 
- Return type is on the same line as function name

Sevz17:
> Hmm okay. Still don't 100% understand what happens here, may need to read more into this.

A good start is [pipe(7)](https://linux.die.net/man/7/pipe) (read "I/O on pipes and FIFOs" near the end)

patrickgold:
I guess we can close this PR as it has been superseded by #1895

PR #1798, opened 2022-04-18
Update Kurdish layout and overall improvements 

GoRaN909:
Makes some changes in Kurdish layouts also added some symbols that usually using in Kurdish language.


patrickgold:
Thanks for your PR! Had a look at it, everything checks out. Will merge your changes in and will release it with 0.3.16-beta01.

PR #1822, opened 2022-04-28
Major rework of input logic, connection handling and state management

### LINKED ISSUE #524: Double-space period is placing the period in the wrong spot
#### Short description
When entering a double-space for a period, the period is placed in the second space like this .

#### Steps to reproduce
1. Type words into any input field
2. Double-space to place a period
3. See period in wrong place

#### Environment information
- FlorisBoard Version: v0.3.9
- Install Source: F-Droid
- Device: Mi Mix 2
- Android version, ROM: 9, Lineage OS 16

Additionally, see the attached GIF: https://imgur.com/a/NsT1XI2

### LINKED ISSUE #637: Hold delete key to delete last words precisely
also, giving an option for vibration feedback while deleting words.

### LINKED ISSUE #828: Backspace deletes whole word and moves the cursor around
Hey guys. Tried looking through open issues and didnt see this. Sometimes while typing if I hit backspace to correct a typo, it will delete the whole word and the cursor will randomly move to another word so I end up typing in the middle of another word.

### LINKED ISSUE #1036: Option to only delete a single letter after glide
With the delete key swipe making multi-character deletions quicker, it'd be nice to have a toggle so tapping the delete key immediately after gliding could either delete the entire word (as it is now) or delete an individual character.

### LINKED ISSUE #1037: incorrect behaviour send key in Signal app
#### Short description
Shift + send key does not create new line in Signal app, instead it sends the message. Similar to
https://github.com/florisboard/florisboard/issues/33 
Have not noticed issue in other apps

#### Steps to reproduce
1. Open Signal app and click message box to compose message
2. Press shift key + send key
3. See error

#### Environment information
- FlorisBoard Version: 0.3.12
- Install Source: GitHub
- Device: Pixel 4a 5G
- Android: GrapheneOS (Android 11)

### LINKED ISSUE #1541: Delete after glide typing should delete the whole word
#### Short description
Using the latest beta, after glide typing outputs a word, tapping the delete key only deletes one character, it used to deleted the whole word in v0.13.3

#### Steps to reproduce
1. Glide type some word
2. Tap delete key
3. Only the last character is deleted

#### Environment information
- FlorisBoard Version: v0.13.4beta09
- Install Source: Fdroid
- Device: OnePlus 7T
- Android: 11

### LINKED ISSUE #1617: Keyboard preview bug inside the app
#### Short description
Hi in the latest version 0.3.14-beta13 (69) when you want to test a feature inside the FlorisBoard app the keyboard suddenly returns to characters when you press on Delete (backspace) key that's happening on all keyboard screens (symbols, emojis, phonepad.. etc)

This bug happened just when you Test a keyboard inside the florisboard app its okay when you use it in all other apps.

#### Steps to reproduce
1. Go to 'FlorisBoard'
2. Click on 'try out your setup'
3. Scroll down to 'Delete key'

https://user-images.githubusercontent.com/3160786/156819878-01d86ae4-0c4d-4f56-a594-78d4eb0f9e00.mp4

Note: I disabled (spacebar switch back to characters automatically) but the bug still there.

#### Environment information
- FlorisBoard Version: 0.3.14-beta13 (69)
- Install Source: Google Play
- Device: Galaxy S20 Fe 5G
- Android: 12, Stock OneUI 4

### LINKED ISSUE #1712: Error: Double-space period weird behaviour
#### Short description
Pressing space multiple times with double-space period enabled results in characters being deleted

#### Steps to reproduce
1. Go to 'Suggestions & Corrections'
2. Scroll down to 'Corrections'
3. Enable 'Double-space period'
4. Type in text
5. Press space buttom multiple times
6. See error

#### Environment information
- FlorisBoard Version: 0.3.14
- Install Source: F-Droid
- Device: Samsung Galaxy S8+
- Android: 12.1, LineageOS Unofficial

Screen Recording: https://user-images.githubusercontent.com/89562141/160274979-7ad6d2ab-3405-45d7-89bc-480757afcb00.mp4

### LINKED ISSUE #1733: "Delete Words Precisely" gesture deletes many characters/words
#### Short description

This issue is intermittent, and is caused by an inconsistent and sometimes very high sensitivity in the swipe gesture. 

After typing many words into a text field (app independent), deleting 30 words sometimes only takes a gesture large enough to go from the delete key to the "n" key. Other times, I needed to go the "c" or "v" key. 

After the 31 word point, the selection jumps all the way to the top of my long note to delete everything.

Sometimes many words are selected by the smallest gesture that is recognized, meaning that precise selection is impossible.

**Edit:** The point at which the selection jumps up to the top of the text box depends on character count, at around 321 characters. Standard characters and emoji seem to all count the same, but a flag emoji counts for 4 characters towards that limit. This would be because of how flags are represented in Unicode, so that may be a hint. (Although I expected it would count as only [2 characters](https://www.youtube.com/watch?v=sTzp76JXsoY))

I've seen these inconsistencies in the FlorisBoard "try your setup" box, the "simple notes" app, and Signal.

Another oddity is that swiping *right* on the delete key also deletes words to the left, which means I have to be annoyingly precise in order to release at the exact point that doesn't delete anything.
 
#### Steps to reproduce
1. Go to "Gestures & Glide Typing" 
2. Scroll down to "Delete key swipe left" (under the "Other gestures" heading)
3. Select "Delete Words Precisely"
4. Enter many words into text box and swipe left on the delete key.
5. With many attempts you should see huge variability in how large of a gesture is needed to delete several words. 
6. If you do not encounter the bug, maybe try going to a different screen and coming back.

#### Environment information
- FlorisBoard Version: 0.3.14
- Install Source: F-Droid
- Device: Pixel 5
- Android: GrapheneOS (Android 12) SP2A.220305.012.2022033013

### LINKED ISSUE #1792: Spacebar deletes some characters when tapped rapidly.
#### Short description
The spacebar sometimes deletes the characters if it is tapped several times very rapidly in the typing space present at the very bottom of Florisboard app.

#### Steps to reproduce
1. Go to 'Florisboard app'
2. Click on 'the down-most typing space for checking current keyboard setup.
3. The spacebar sometimes deletes the characters if it is tapped several times very rapidly.
4. It is only seen on the Typing space present at the very bottom of the app itself. It is not reproducible so far in any other field while typing. 

#### Environment information
- FlorisBoard Version: 0.3.15(80)
- Install Source: F-Droid
- Device: Redmi note 7s (lavender)
- Android: 10, MIUI Global 12.5.1 stable

https://user-images.githubusercontent.com/88554859/163524647-cff033dd-0a19-4390-aa66-057ad9dba080.mp4

patrickgold:
This PR aims to improve the way internal key events are processed, how the text of the editor is cached and how the state of the input logic is managed. A lot of testing and re-evaluation has gone into this rework, and there's still some things to do to fix up bugs and other issues.

### Minimum API level raised from 23 to 24
(Android 6 --> Android 7)

This step was hard to decide because I had to carefully weigh in the effect of discontinuing support for devices running Android 6. I only have limited data (bug reports and GPlay) for device distribution, but going from that only a tiny tiny fraction of my user base is still on Android 6, so I decided that the benefits outweigh the drawbacks of raising the minimum API level by far.

**Why was there even the need to consider this step?**

Well simple answer: FlorisBoard now makes use (and will even more in the future) of the [`android.icu.text`](https://developer.android.com/reference/android/icu/text/package-summary) platform package, which exposes the system installed ICU module to the app via Java APIs. This allows me to easily analyze complex Unicode strings and find word and character boundaries, something which is really important for correct input behavior and for the upcoming word suggestions. While with most APIs there's a simple workaround by using `if..else` statements for lower API levels, for `android.icu.text` however there's sadly no easy way to substitute its feature set on devices prior to Android 7.

### Editor caching

To improve input performance, FlorisBoard now caches the input for rich input connections locally (currently 256 characters around the cursor + any selected text) and predicts the new cache state for many user actions. If the selection update callback metadata matches with the prediction it means we do not have to re-fetch the editor text and thus save a lot of IPC-roundtrip time, which, depending on the app, can be massive.

### Input event dispatcher changes

The input event dispatcher has been reworked to remove the async channel logic entirely. This logic was a common source of state bugs and other weird state mismatches and has been replaced by a concurrent model for managing active keys and their repeating/long press actions.

### Other changes / fixes

Closes #524 
Closes #637 
Closes #828 
Closes #1036 
Closes #1037 
Closes #1541 
Closes #1617 
Closes #1712 
Closes #1733 
Closes #1792


tsiflimagas:
Wow, that lag in input seems to have been fixed! Even on debug builds, it works fast even right after installation or force close! That was a pretty major issue, and it feels great that it's now fixed.
Two things to mention about the PR, which you may already be aware of, are:
* Glide typing should check whether it's the start of the line to not prepend a space.
* When long pressing the shift key for caps lock, it doesn't get colored until you press another key.

patrickgold:
Thanks for testing it out! Great to hear that it works more smoothly now, that's one of the major goals of this rework.

Regarding the issues you mentioned:

The last pushed commits have fixed the glide phantom space behavior, on a new line there's now no additional space appended.

However the shift key logic in general is a bit buggy and needs some tweaking, I think though that I will fix this in beta02.

PR #1859, opened 2022-05-19
Update hu.json

ptrg:
Addition character assignments and changed TLD strings for Hungarian keyboard layouts.

These are important changes because Hungarian users mostly search accented characters where they belong.
For example, the u ú ü ű are similar shapes but u ú and ü ű are the short and long symbol pairs for two different sounds (the same is true for o ó and ö ő).
So the changes are intended to show up as follows:
- assignments for o and u is kept for compatibility with the old style and non-hungarian layouts but with hungarian language (as I use florisboard)
- assignments added for ö and ü to found their long versions (ő and ű) in the right place too (as others use florisboard)

TLD strings are changed because .hu and .eu is widely used in Hungary and .gov.hu is the official ending for government sites.

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/hu.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/hu.json
@@ -22,6 +22,11 @@
         { "$": "auto_text_key", "code":  337, "label": "ő" }
       ]
     },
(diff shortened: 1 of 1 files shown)


patrickgold:

I tested out the changes, and everything checks out fine. Will merge the changes now into master and release them with 0.3.16-beta04.

PR #1884, opened 2022-05-30
Add DvorakES layout

### LINKED ISSUE #1853: Dvorak-ES layout request
Hello, 
If possible it would be very nice to add a **Dvorak-es** layout to florisboard.

Something similar to this one:

Thanks for such a nice app!

tsiflimagas:
Not that I speak Spanish, it was just easy to add. Please let me know if I should label it differently. Gboard for example names it "Dvorak (ÑÇ)".

Closes #1853

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
@@ -112,6 +112,13 @@
         "modifier": "org.florisboard.layouts:dvorak"
       },
       {
+        "id": "dvorak_es",
+        "label": "Dvorak (ÑÇ)",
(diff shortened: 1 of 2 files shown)


patrickgold:
Thanks a lot for adding the Dvorak ES layout!

I tested it locally and everything checks out fine. As for the name, if Gboard names it "Dvorak (ÑÇ)" we should probably name it that way too, so new users can more easily recognize this layout (although I am myself not entirely sure).

tsiflimagas:
> Thanks a lot for adding the Dvorak ES layout!

No problem😁

> if Gboard names it "Dvorak (ÑÇ)" we should probably name it that way too, so new users can more easily recognize this layout

I agree, this makes sense. Also changed the commit name accordingly.

Issue #196, opened 2021-01-16
Feature request:(easy to implement) full layout customization


ShiroiKuma0:
I'd like to request full layout customization — this might seem like a tall order, it's however implemented nicely, via a simple interface in Multiling O Keyboard. 

There you can add via text layouts full new keyboard layouts which are then used. 

This is great — because I use different layouts, different number of columns, rows, characters displayed per app user scenario. 

The big problem with it is the keyboard is not FLOSS, so can any time become obsolete. 

So this would be a great opportunity here.

patrickgold:
Thanks for your feature request! I am currently unsure if and how exactly I will implement dynamic keyboard creation within the Settings UI. I will add it to the maybe list for v0.5.0 though!

patrickgold:
Just noticed that this issue is still on "maybe" status, updated it to "accepted" for v0.5.0.

HiGein:
That would be awesome to see this feature in 0.5!

SilverGreen93:
Looking forward to this! I want to create my custom keyboard layout.

dky4:
Would be great to see personally!
Is it getting added soon? I can't seem to find it right now.

ghost:
## custom keyboard layout
> Users can drag to adjust each letter

<img width="1344" height="768" alt="Image" src="https://github.com/user-attachments/assets/4881c92b-62e3-4dde-b6ef-3a01bbb5c704" />

![Image](https://github.com/user-attachments/assets/8dbd53e2-9929-4f16-8920-171d2e278a9a)

provisionally:
> I am currently unsure if and how exactly I will implement dynamic keyboard creation within the Settings UI.

@patrickgold With regards to the how, #3022 from Samsung is likely the best and most widely used implementation there is.

lm41:
We are currently implementing a parser for the keyboard3 standard fron unicode and we are going to add the possibility to add and modify keyboard3 layouts.

caco3:
I also would like to design my own keyboard like I could do in Multiling O Keyboard.
What is the state of the work on this? Could I help working on it?

What do you refer to with "keyboard3 standard"?

lm41:
Unicode cldr github: https://github.com/unicode-org/cldr/tree/main/keyboards
Unicode ldml keyboard3 stabdard: https://www.unicode.org/reports/tr35/tr35-74/tr35-keyboards.html

[...]

PR #1890, opened 2022-06-02
Adding new layout Persian4 & some modifications on Persian2 layout

m-koushan:
Added new Persian4 layout based on Samsung keyboard layout. Here is the picture:
image

Removed some Arabic characters in popups and reordered popups in Persian2 layout.

Edit: checked the artifact, everything works fine.

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
@@ -283,6 +283,13 @@
         "modifier": "org.florisboard.layouts:persian3"
       },
       {
+        "id": "persian4",
+        "label": "Persian4",
+        "authors": [ "M-Koushan" ],
+        "direction": "rtl",
+        "modifier": "org.florisboard.layouts:persian2"
+      },
+      {
         "id": "qwerty",
         "label": "QWERTY",
         "authors": [ "patrickgold" ],
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/characters/persian4.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/characters/persian4.json
@@ -1,0 +1,36 @@
+[
+  [
+    { "code":  1589, "label": "ص" },
+    { "code":  1579, "label": "ث" },
+    { "code":  1602, "label": "ق" },
+    { "code":  1601, "label": "ف" },
+    { "code":  1593, "label": "ع" },
+    { "code":  1607, "label": "ه" },
+    { "code":  1582, "label": "خ" },
+    { "code":  1581, "label": "ح" },
+    { "code":  1580, "label": "ج" },
+    { "code":  1670, "label": "چ" }
+  ],
+  [
+    { "code":  1588, "label": "ش" },
+    { "code":  1587, "label": "س" },
+    { "code":  1740, "label": "ی" },
+    { "code":  1576, "label": "ب" },
+    { "code":  1604, "label": "ل" },
+    { "code":  1575, "label": "ا" },
+    { "code":  1578, "label": "ت" },
+    { "code":  1606, "label": "ن" },
+    { "code":  1605, "label": "م" },
+    { "code":  1662, "label": "پ" },
+    { "code":  1705, "label": "ک" }
+  ],
+  [
+    { "code":  8204, "label": "half_space" },
+    { "code":  1591, "label": "ط" },
+    { "code":  1586, "label": "ز" },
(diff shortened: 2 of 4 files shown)


m-koushan:
Isn't possible downloading artifacts right after build? or if it is and I'm doing something wrong can you tell me what should I do?
I'm not so advance in git stuff by the way.

patrickgold:
Ok before I review this PR, the number of different Persian layouts starts to get out of hand. This is a forth version of Persian (+ modifications of an existing version), is there not a kind of standard?

I don't want to discredit your efforts at all and I really appreciate any input, but the amount of changes within the Persian layouts within FlorisBoard is really high and I don't understand why the Persian layouts are so volatile. As you seem to be a native speaker of Persian you may be able to answer this question: Is it because there just isn't a standard and these changes are just personal preferences or is there another reason like regional differences, dialects, etc?

If we continue at this pace by the end of the year we have 10 different Persian layouts and 20 different PRs overriding each others contribution, that is something I want to avoid if possible.

m-koushan:
The variation of Persian layouts is because as you said, we don't have an standard version. Also every single keyboard had their own layout for Persian language and for this reason, when somebody wants to change from whatever they keyboard was to something new, one thing they think about it most is if the new one has the layout which they were used for several months or even years. 
I made 2 keyboard layouts based on two popular keyboard apps (SwiftKey Persian2, Samsung Persian4) and also I never saw Persian 1 layout before the FlorisBoard.
Anyway, its based on your choice. I appreciate your hard work and I agree with whatever you do about this.
But in my opinion, the app is better to have more layouts for every language for people comfort.

m-koushan:
I searched a little bit and I found this
https://persian-computing.org/references/ISIRI/ISIRI-2901.html
This is standard layout but I doubt that its popular between Persian people (at least in smart phones keyboard)

patrickgold:
> The variation of Persian layouts is because as you said, we don't have an standard version

Ah I see, what I thought the reason was.

> But in my opinion, the app is better to have more layouts for every language for people comfort.

I agree with that and you are correct. Though maybe it would be better to give the Persian layouts sub-names instead of just appending numbers? For instance the pattern "Persian (variations)", where `variations` are 2-3 characters which characterize that specific Persian layout the most? I just don't like the numbering, which means nothing to a new FlorisBoard user.

> Edit: Sorry I forgot to tell that most foreign people think Persian language alphabet is as same as Arabic language alphabet, in fact they're not. They have similarity but they are not the same. The reason I modified the Persian2 layout was that most of popups were for Arabic language and we don't use them at all.

That's what I thought too, but for me I can only trust what contributors think should be part of the popups, I don't speak Arabic nor Persian so I don't really know.

> This is standard layout but I doubt that its popular between Persian people (at least in smart phones keyboard)

Hmm, yeah if its not popular nor known within the smartphone market then it may have a reason, maybe usability or just the way it went.

GoRaN909:
> If we continue at this pace by the end of the year we have 10 different Persian layouts and 20 different PRs overriding each others contribution, that is something I want to avoid if possible.

@patrickgold You right, Maybe he will add more than 50 layout which is just make a small changes on letters places I think the best way to solve this issues is add new option to the FlorisBoard and let users manually change the place of characters or edit exiting layouts like this one requested to add/edit custom layouts in #196

m-koushan:
>I searched a little bit and I found this
>https://persian-computing.org/references/ISIRI/ISIRI-2901.html
>This is standard layout but I doubt that its popular between Persian people (at least in smart phones keyboard)

I read this and I agree its very handy and popular(in PC keyboards). Also it has standard from Iran government and I'm pretty sure we can use it for an standard keyboard layout.

>EDIT: the only standard I found was the CLDR fa layout, although I don't know if this is used by many people.

I checked this too. It has some similarities with standard version (QWERTY) that I mentioned above, but its not good for these reasons:
1. Its not standard version(its based on usability in big companies such as Microsoft)
2. It has Arabic characters that we don't have them.
examples: (ي, ء , ...)
3. Its not in the handy order that the standard version is.

>That's what I thought too, but for me I can only trust what contributors think should be part of the popups, I don't speak Arabic nor Persian so I don't really know.

I'm from Iran and we speak Persian. I'll do my best to help you for this great project.

>@patrickgold You right, Maybe he will add more than 50 layout which is just make a small changes on letters places

I understand your concern. Unfortunately now I'm really busy and I can't do something that's really big for you and the only thing which keeps me up is programming, I find that easy so as a habit I do this and I promise you I won't make more layouts(Persian at least).
Also I agree with your suggestion. However it may cost a long time to do. I suggest adding this until you find a way.

PR #1895, opened 2022-06-04
Cpp refactor

Sevz17:
Supersedes: #1795


app/src/main/cpp/dev_patrickgold_florisboard_FlorisApplication.cpp
```diff
@@ -42,12 +41,14 @@ Java_dev_patrickgold_florisboard_FlorisApplication_00024Companion_nativeInitICUD
...
     if (!in_file) {
+        delete[] icu_data;
         in_file.close();
         return U_FILE_ACCESS_ERROR;
     }
     icu_data[size] = 0;
     in_file.close();
     UErrorCode status = U_ZERO_ERROR;
     udata_setCommonData(reinterpret_cast<void *>(icu_data), &status);
+    delete[] icu_data;
```
patrickgold, line 52: This line right here should be removed, as `udata_setCommonData` does not make a copy of the `icu_data` array, and requires that our array stays in memory. Actually if this line is existent FlorisBoard instantly crashes for the first word being spell-checked.

```suggestion
```

app/src/main/cpp/utils/log.cpp
```diff
@@ -15,67 +15,60 @@
...
-
-void utils::log_error(const std::string &tag, const std::string &msg) {
-    __android_log_print(ANDROID_LOG_ERROR, tag.c_str(), "%s", msg.c_str());
-}
-
-void utils::log_wtf(const std::string &tag, const std::string &msg) {
-    __android_log_print(ANDROID_LOG_FATAL, tag.c_str(), "%s", msg.c_str());
+void utils::log(int log_priority, const std::string &tag, const std::string &msg) {
+    __android_log_print(log_priority, tag.c_str(), "%s", msg.c_str());
 }
```
patrickgold, line 28: The reason why I had several different methods for logging is to have some sort of consistency with the Kotlin counterpart, where each level also has a different method for better readability. I guess we can keep your version too, but this change didn't necessarily improve the C++ code base.

app/src/main/cpp/utils/log.cpp
```diff
@@ -15,67 +15,60 @@
  */
 
 #include <android/log.h>
+#include <errno.h>
+#include <string.h>
```
patrickgold, line 19: Is there a particular reason why you don't use the modern C++ header names for errno.h and string.h?

```suggestion
#include <cerrno>
#include <cstring>
```

patrickgold (changes requested):
Thanks a lot for your C++ refactor, really appreciate it!

I have reviewed your changes and found some issues or have questions regarding them. These need to be addressed or resolved first, so I can merge your changes in.

app/src/main/cpp/utils/log.cpp
Sevz17: What error?

tsiflimagas: @GithinjiHans you probably didn't initialize git submodules. It's done by running `git submodule update --init --recursive`.  Btw, that's not the most suitable place for this question. Next time better use the issues or the discussions page.

patrickgold (approved):
Sorry for the long delay in re-review, I ha d a lot of work todo lately. I have rebased your branch to the latest state of master to solve all merge conflicts.

As for the Cpp refactor, everything looks good from me now, will merge in your changes now and release them with 0.4.0-alpha1. Thanks again for your efforts!

Issue #1888, opened 2022-05-31
Add search to clipboard


MHKalantarian:
I wanted to add search to clipboard, I have added the filter logic and implemented a textfield to get query from, but problem is keyboard doesnt focus on the textfield inside the smartbar and only types in the targeted app textfield, is there anyway to redirect the keyboard to type inside its own layout?

patrickgold:
That's also a way of achieving this task, haven't thought of doing it this way yet. Although it is still a bit hacky, it is far less hacky than my proposed solution, so if you want to go for it I'd accept a PR with this solution :)

MHKalantarian:
@patrickgold Please take a look at my PR, I believe you can solve the small problems that I've encountered:
Lose focus on touch outside keyboard layout
Backspace not working
#1897

tutacat:
Probably the best way to do this (like other keyboards for example) is to use the app's textbox or a temporary buffer as the input field for the search. Then replace the whole value or matching text when it is selected.

Another method is implementing it as a suggestion, but that could cause a runaway process (especially if you didn't limit the characters for a search, or didn't wait for typing).

While you could disable sending input method to have an internal textbox, it is not designed for that, and can cause confusion or bugs.

DeanLemans:
Bump

PR #1897, opened 2022-06-06
Internal text field

MHKalantarian:
I have extracted a widget to type inside the keyboard layout
Problems are:
Backspace button not deleting anything
Focus not lost when touched outside keyboard

#1888

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Tue May 31 12:14:01 IRDT 2022
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
+distributionSha256Sum=29e49b10984e585d8118b7d0bc452f944e386458df27371b49b4ac1dec4b7fda
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
+#Tue May 31 12:14:01 IRDT 2022
 distributionBase=GRADLE_USER_HOME
-distributionPath=wrapper/dists
-distributionSha256Sum=29e49b10984e585d8118b7d0bc452f944e386458df27371b49b4ac1dec4b7fda
 distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
-zipStoreBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
 zipStorePath=wrapper/dists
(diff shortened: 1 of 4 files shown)


patrickgold:
Thanks a lot for your PR, really appreciate it!

I've had a quick look at the changes and the base idea is good!

The main problem with focus on Android is that the concept is completely stupid and does not work in anyway that would make sense, JC makes that even worse. I will have a look later today and play around with your code to address your mentioned problems (have some idea but am not sure if it is the cause), if I have results then I will post the patch commit here.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
+#Tue May 31 12:14:01 IRDT 2022
 distributionBase=GRADLE_USER_HOME
-distributionPath=wrapper/dists
-distributionSha256Sum=29e49b10984e585d8118b7d0bc452f944e386458df27371b49b4ac1dec4b7fda
 distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
-zipStoreBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
 zipStorePath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
```
patrickgold, line 6: In the meantime, could you please revert the changes of the gradle-wrapper.properties file please? This file is auto-generated by my Gradle setup and thus should not be hand-edited. Thanks!

patrickgold (changes requested):

patrickgold:
Sorry for getting back this late, I tried out some changes a week ago but did not succeed and then got distracted. I will currently leave your PR in this state and when I find the time provide a mechanism for the problems you mentioned. I hope that this is okay for you.

MHKalantarian:
Could you point me to the version before you changed everything to compose? I want to try somethings on legacy

patrickgold:
Basically it is the 0.3.13 release tag: https://github.com/florisboard/florisboard/tree/v0.3.13

After that I almost immediately began adding Compose.

MHKalantarian:
I tested out various situations, found a workaround to work with a button instead of handling focus. But the problem with cursor still remains, when typing in internal edit text using it input connection, the cursor does not move forward, therefore backspace won't work and words are backward. Can you suggest anything for this part?

MHKalantarian:
Problem fixed. The only problem that remains is to detect touch gesture outside of keyboard bounds, I have not found any documentation for this. Do you know anything about this?

atikattar1104:
Hi!
Currently Typing With Florisboard.
I Have Tested It. Results As Follows :- 
1) Backspace Is Working In All Edittext Views.
2) Keyboard Is Closing When Touched Outside Keyboard On Webpages And When Touching Clickable Views In Apps (Like, Buttons, Switches, Etc.). But, When You Click On Blank Area (i.e. - Non-Clickable Views) In Apps, Keyboard Doesn't Closes. Same Behaviour Is With Other Keyboards As Well.
Thanks.

patrickgold:
I will close this PR as it is stalled for a long time and not really compatible with the current code base anymore. Adding a text box within the keyboard UI is definitely possible but not easy to get right and then maintain, so for now it is best to just not have it in the keyboard UI. Once the keyboard code base nears 1.0 we can then try adding it in.

PR #1963, opened 2022-07-21
Add circumflex popups to Turkish layout (#1962)

### LINKED ISSUE #1962: Can you add circumflex letters to Turkish keyboard (â, û, î)
In Turkish; Ââ, Ûû and Îî characters are used sometimes. It would be nice if they were in the keyboard when you hold the key, just like how it is in the attachments.

Ââ in the pop-up menu

Ûû in the pop-up menu

Îî in the pop-up menu

If you are able to add the other characters shown in the pop-up, it would be so great if you could do that! But if you are going to do it, can you please make "ı" as the first suggestion, and "î" as the next to that one, just like how the order is in the screenshots? (the same for â, and û) Because ü and ı are often used whilst û and î are rarely used in Turkish

tsiflimagas:
So, I tried to address this, but i couldn't get "î" to also show with caps lock. So I'm hoping for a little tip😃

Closes #1962

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/tr.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/tr.json
@@ -1,5 +1,8 @@
 {
   "all": {
+    "a": {
+      "main": { "$": "auto_text_key", "code":  226, "label": "â" }
+    },
     "c": {
       "main": { "$": "auto_text_key", "code":  231, "label": "ç" }
     },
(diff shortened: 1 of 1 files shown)


tsiflimagas:
Ah, and i was wondering why the `"ı"` section was there...really tricky indeed😆 Thanks for the hint!

patrickgold (approved):
Thanks for implementing the fix, merging your changes in now! Will be parts of the 0.4.0-alpha02 release.

Issue #1840, opened 2022-05-10
Capitilized i in Turkish outputs "I"


Atlas-19:
#### Short description
Whenever I press shift and press i the keyboard outputs "I" instead of "İ".

A video demonstration:
https://user-images.githubusercontent.com/71735202/167711788-b5504a94-ea5f-4fb8-b56e-2f8bf29df0f3.mp4

#### Steps to reproduce
1. Go to "Languages & Layouts"
2. Create a Turkish keyboard subtype
3. Press shift and type "i" using the aforementioned subtype

#### Environment information
- FlorisBoard Version: 0.3.15 (80)
- Install Source: F-Droid
- Device: Lenovo TB-X606F & Samsung SM-J730F
- Android: Android 10 & Android 9

demirdegerli:
Fixed this 🇹🇷 
#1966

patrickgold:
@demirdegerli Please see #1966 why I think your PR does not address the issue, while my PR (#1967) does address the issue.

Both of you: could you try out #1967's debug artifact (installs separately and does not alter your main/beta installation) and test out the behavior of the I in QWERTY for both a Turkish and English subtype? The Turkish subtype should cause the capital I to have a dot, while the English one shouldn't have one. Thanks in advance!

demirdegerli:
Ok, this commit (#1967) dynamically solves the issue for all languages. Exactly corrects me, thanks!

PR #1966, opened 2022-07-25
fix upper and lower characters for "i" and "ı"

demirdegerli:
see issue #1840

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/tr.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/tr.json
@@ -8,14 +8,14 @@
     },
     "i": {
       "main": { "$": "case_selector",
-        "lower": { "code":  305, "label": "ı" },
-        "upper": { "code":   73, "label": "I" }
+        "lower": { "code":  305, "label": "i" },
+        "upper": { "code":   73, "label": "İ" }
       }
     },
     "ı": {
       "main": { "$": "case_selector",
-        "lower": { "code":  105, "label": "i" },
-        "upper": { "code":  304, "label": "İ" }
+        "lower": { "code":  105, "label": "ı" },
+        "upper": { "code":  304, "label": "I" }
       }
     },
     "o": {


patrickgold:
Thanks for your PR!

I think you misunderstood the issue: the issue was that the base character did show a capital I with dot above, however a normal I was written. As such I created a new PR for this issue: #1967 and I think this PR is obsolete. Please also see #1840.

demirdegerli:
correct pr is #1967

patrickgold:
@demirdegerli Is your device language by chance Turkish? If so it explains why the base char issue is not apparent at first, because then every i gets capitalized to İ and it seems correct.

> Here is the capital and lower versions of both letters:

I agree that these are the correct capitalization rules for the i and ı, however those are already implemented correctly (using #1967 as the code base for this statement), or am I mistaken here?

See implementation of the popups for the i and ı: https://github.com/florisboard/florisboard/blob/b576cafaa4a66b128bdb558e5916482c697d3c4e/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/tr.json#L9-L20

demirdegerli:
Ok, my device language is Turkish and in the QWERTY layout the capital i character "I" showing up "İ" to me because the capitalize method capitalizes the letter for my language. Now I understood, right?

demirdegerli:
@patrickgold Ok, my device language is Turkish and in the QWERTY layout the capital i character "I" showing up "İ" to me because the capitalize method capitalizes the letter for my language. Now I understood, right?

patrickgold:
> Ok, my device language is Turkish and in the QWERTY layout the capital i character "I" showing up "İ" to me because the capitalize method capitalizes the letter for my language. Now I understood, right?

Yup :)

> And with your PR, Turkish users also can use the QWERTY layout?

Yes, with my #1967 fix Turkish users can use QWERTY for both Turkish and English and it will properly adapt thanks to `state.recomputeIfNecessary(evaluator.activeSubtype().primaryLocale)`, for English it is i->I and for Turkish i->İ. Of course there are also the Turkish-Q and Turkish-F layouts, which already have a custom config and should work flawlessly.

demirdegerli:
Ok, thanks!

PR #1984, opened 2022-07-29
Fix uppercase Greek vowels popups

### LINKED ISSUE #1981: Broken uppercase accented characters in Greek layout
#### Short description
I just came to notice that popups for vowels in Greek don't have any accents, in current HEAD. The offending commit seems to be https://github.com/florisboard/florisboard/commit/b04f8d75f36659beb046fe7ab55d1a8ae1600d4c.
Btw I'm not sure if this should be labeled as `kbd-logic` instead.

#### Steps to reproduce
1. Go to the keyboard, using the Greek layout
2. Hold on a vowel
3. See that popups don't have accents

#### Environment information
- FlorisBoard Version: 0.4.0-alpha01
- Install Source: locally compiled
- Device: Xiaomi Redmi note 8
- Android: Android 12.1, custom rom

tsiflimagas:
So, I gave this a shot, but it turned out I didn't know how to handle more than one popup with `case_selector`. Thus, I'm once again looking for an advice😁

Closes #1981

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/el.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/el.json
@@ -2,41 +2,74 @@
   "all": {
     "α": {
       "relevant": [
-        { "$": "auto_text_key", "code":  940, "label": "ά" }
+        { "$": "case_selector",
+          "lower": { "code":  940, "label": "ά" },
+          "upper": { "code":  902, "label": "Ά" }
+        }
       ]
     },
     "ε": {
       "relevant": [
-        { "$": "auto_text_key", "code":  941, "label": "έ" }
+        { "$": "case_selector",
+          "lower": { "code":  941, "label": "έ" },
+          "upper": { "code":  904, "label": "Έ" }
+        }
       ]
     },
     "η": {
       "relevant": [
-        { "$": "auto_text_key", "code":  942, "label": "ή" }
+        { "$": "case_selector",
+          "lower": { "code":  942, "label": "ή" },
+          "upper": { "code":  905, "label": "Ή" }
+        }
       ]
     },
     "ι": {
-      "main": { "$": "auto_text_key", "code":  943, "label": "ί" },
+      "main": { "$": "case_selector",
+        "lower": { "code":  943, "label": "ί" },
+        "upper": { "code":  906, "label": "Ί" }
+      },
       "relevant": [
-        { "$": "auto_text_key", "code":  912, "label": "ΐ" },
-        { "$": "auto_text_key", "code":  970, "label": "ϊ" }
+        { "$": "case_selector",
+          "lower": { "code":  912, "label": "ΐ" },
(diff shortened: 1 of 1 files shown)


patrickgold:
You can put case selectors in the relevant section too, basically you can put a case selector in every spot there an auto text key also fits :)

tsiflimagas:
Thanks, it now works fine! I initially tried some weird stuff and didn't think of the simplest.
One last thing which I just came to notice. It seems like I found out why Gboard doesn't insert `Ϊ́` and `Ϋ́` but only their lowercase forms. These characters don't actually exist standalone in Unicode, instead they're a combination of the plain letter+dieresis+accent. Till now in FlorisBoard I had been using the correct label, but the code was actually just the plain letter. What's weird is that they handle this in AOSP keyboard, but not in Gboard. Anyway, is there any way at the moment to have a combination of letters in the layout, or should I go with what Gboard does with those two characters?

patrickgold:
> Anyway, is there any way at the moment to have a combination of letters in the layout, or should I go with what Gboard does with those two characters?

You can use a [`MultiTextKey`](https://github.com/florisboard/florisboard/blob/79d9e73608c76ed7645e4830caf7e33e78052358/app/src/main/kotlin/dev/patrickgold/florisboard/ime/text/keyboard/TextKeyData.kt#L557-L565) for this. Actually this would be the first proper use for it. Example (code points are exemplar, you need to insert the correct ones which compose the letter):

```json
{ "$": "multi_text_key", "codePoints": [1, 2, 3], "label": "Ϋ́" }
```

Note that multi text keys do not have an auto case mapper, so you need to use a case selector and put above snippet in for the `upper` variation.

tsiflimagas:
Oh, great! I tried but it seems I'm doing something wrong, as it just outputs the letter without the other two characters, so I pushed it in order for you to spot the mistake easier. I can't quite spot it myself.

patrickgold:
> I tried but it seems I'm doing something wrong, as it just outputs the letter without the other two characters

You are doing everything correct, the issue is that the editor instance and the composer don't support multiple code points... I now adjusted the logic and created PR #1989 to fix this circumstance. I tested the behavior and it inserted the `Ϋ́` as intended for me. If you want you can test this out locally (check out my PR's branch and manually apply the commit you pushed above).

I still need to do some testing if I didn't break something else with this change in the composer, but I am too tired for it now so I will do this tomorrow, then I will merge in the change into master.

patrickgold:
#1984 is now merged, so if you rebase your branch onto the latest master it should work flawlessly now.

tsiflimagas:
Yes, it works great! Thank you 😁

patrickgold (commented):
Thanks for the rebase! I did check the changes again and they look good, all the Greek popups are now properly capitalized and inserted correctly.

One question though: For letters with only one popup (`α`, `ε`, `η`, `ο` and `ω`) they previously only had a `relevant` popup, now they have a `main` popup. Is this change intended?

tsiflimagas:
> Is this change intended?

Lol, I actually don't know why I did that, nor noticed there was no reason to. Changed it back.

patrickgold (approved):
Thanks for the fix, now everything checks out fine. Will merge in the changes now.

PR #2037, opened 2022-08-30
Add spanish fastlane metadata

delthia:
Hi,
I've seen in the `CONTRIBUTING.md` file that translations must be done through crowdin, but as I couldn't find the metadata there, I decide to translate it this way. Then I saw #750.
So I made the changes, hoping that they can be used this way. If they can be merged, great, but if you prefer for them to be added in the translation platform, I don't mind doing so either.

Thanks


patrickgold (changes requested):
Hey delthia,

thanks for your PR, really appreciate it! It is true that the app store texts are currently not available for translation on Crowdin, as such your PR will get the same exception as the linked one. I will probably move the translation of the store entries to Crwodin once I do the proper app store branding, however that's still a awhile to go and a topic for the 0.5.0 dev cycle.

As for the changes, they look good. The only thing I require to be changed are that translating changelogs of already released versions is useless, as the metadata for each specific version is fetched at release time and never re-fetched again by the build systems and GPlay. As such your 86.txt files should be removed. Also thanks a lot for providing a localized feature graphic, this is a very neat addition!

delthia:
Hi,
I've just removed the changelog. I translated it thinking that it was the latest one, and therefore the one shown in the store listings, but I didn't check to see if it actually was, I just went from the assumption that the highest numbered one was the latest.

Providing the localized graphic didn't cost me anything! I just took an screenshot on my phone and added the background on gimp. That is why the keyboard is ever so slightly narrower. I just maintained the overall dimensions, so that it isn't a problem.
If you want, I don't mind making some screenshots of the spanish keyboard, but if you are thinking of changing the branding, I don't mind helping to create some graphics on gimp or inkscape, or something like that. Anyways, that is a topic for another issue.

Thanks for the amazing project! I'll keep helping where possible

patrickgold (approved):
Sorry for now reviewing earlier, have a lot ongoing in the background.

Your changes now look good. I will let the automated CI do its job and after it passes I will merge in your changes. Thanks again for the translation!

delthia:
> Sorry for now reviewing earlier, have a lot ongoing in the background.
> 
> Your changes now look good. I will let the automated CI do its job and after it passes I will merge in your changes. Thanks again for the translation!

No problem, I just requested the review in case it was necessary for you to be notified or something like that (this is the first time I make a pull request).

Your welcome. I think that translating the app helps everyone in general, so having the store listings translated is a good way to give a better first impression for people how might not know read the english listing, and it was a quick job that didn't require more than a bit of time.

Issue #2080, opened 2022-10-03
Florisboard crashed on device startup


snaik20:
#### Short description
Got an app crash notification when started the device.

#### Steps to reproduce
1. Shutdown the phone and start it.
2. Once phone is started, observe there is a system notification about the crash
3. Did not open any app after the start, not even florisboard.

#### Environment information
- FlorisBoard [0.4.0-alpha04](https://github.com/florisboard/florisboard/releases/v0.4.0-alpha04) (90)
- Device: Motorola Moto G Play (lineage_harpia)
- Android: 10 (cn=Q sdk=29)

#### Attached logs and stacktrace files
<details>
<summary>Detailed info (Debug log header)</summary>

```
======= SYSTEM INFO =======
Time                : 2022-10-03T10:23:00.824Z
Manufacturer        : Motorola
Model               : Moto G Play
Product             : lineage_harpia
Android             : 10 (cn=Q sdk=29) [e_harpia-userdebug 10 QQ3A.200805.001 eng.root.20220923.195713 dev-keys,dev-release]
ABIs                : [armeabi-v7a, armeabi]
Memory              : 931.43 MiB (48.82% used, 1.86 GiB max)
Font scale          : 1.0
Locales             : en-IN

======= APP INFO =======
Package             : dev.patrickgold.florisboard.beta
Name                : FlorisBoard Beta
Version             : 0.4.0-alpha04 (90)
Build type          : beta
Build commit hash   : 31a01c8d39ba425e4e1e424e9671ffbb96c24fc0
Java heap memory    : 12.54 MiB (6.53% used, 192.00 MiB max)
Native heap memory  : 22.96 MiB (90.23% used, 25.45 MiB max)

======= FEATURE CONFIG =======
Smartbar enabled            : true
Suggestions enabled         : true
Inline autofill enabled     : true
Glide enabled               : false
Internal clipboard enabled  : false

```
</details>

<details>
<summary>1664792356201.stacktrace</summary>

```
java.lang.RuntimeException: Exception while computing database live data.
	at androidx.room.RoomTrackingLiveData$1.run(RoomTrackingLiveData.java:9)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:919)
[...]

PR #2085, opened 2022-10-08
Fix app crash on startup

snaik20:
Fixes - https://github.com/florisboard/florisboard/issues/2080

- The app crashes during initialization while quering from the `clipboard_history` table.
- The CursorWindow is unable to hold the entire row size from the table.
- This could happen because the table had previously inserted a large entry such as image.
- Made changes to prevent inserting an entry if the CursorWindow cannot be initailized with item size.
- Fixed a lint warning to make the method `insertClip` private.
- Note: The crash is not consistently reproducible, the fix discussed above is based on the stacktrace and possible cause related to the exception.

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/clipboard/ClipboardManager.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/clipboard/ClipboardManager.kt
@@ -19,6 +19,8 @@ package dev.patrickgold.florisboard.ime.clipboard
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Context
+import android.database.CursorWindow
+import android.os.Build
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import dev.patrickgold.florisboard.app.florisPreferenceModel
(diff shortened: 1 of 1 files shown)


BHydden:
`Fixes - #2080`

Just a heads up, I don't think it works for edits or comments, but if you had written this line without the hyphen `-` GH would automatically close the issue if the PR is merged. [source](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

snaik20:
> `Fixes - #2080`
> 
> Just a heads up, I don't think it works for edits or comments, but if you had written this line without the hyphen `-` GH would automatically close the issue if the PR is merged. [source](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

Thanks @BHydden, I have updated the description per the docs.

snaik20:
@patrickgold Please review when you get a chance.

patrickgold (commented):
Thanks a lot for your PR, really appreciate it.

I haven't seen the crash from #2080 before yet. Did you insert an abnormally long text-snippet to get this crash?

As for the fix itself, wouldn't it be better if the limit is computed once and then cached? It seems to me as performing this check for every insert is a waste of resources and pretty costly. Although I am not too sure yet what the exact cause of this problem is, so this might be necessary.

> This could happen because the table had previously inserted a large entry such as image.

Images/Videos are stored on the app-private file system, only a path to the image is stored in the database. (At least for the new database system, if you still have a relict database from an old version this might be different)

snaik20:
> Did you insert an abnormally long text-snippet to get this crash?

I have not been able to reproduce the crash with definite steps. It occurred when I reboot my device. The fix is only based on the stack trace that I received in the crash.

> As for the fix itself, wouldn't it be better if the limit is computed once and then cached?

For now, I'm catching the exception that is thrown when input size is too large. I'm in favor of caching the size once and reusing it, but not quite sure how we could calculate it accurately and efficiently. Any suggestions?

PR #2054, opened 2022-09-06
Add Chinese Shape Based Layouts

Waelwindows:
This PR adds Chinese Hanzi shape based input to FlorisBoard, This should be a step towards #149 

Currently, this PR only implements the Zhengma layout (鄭碼輸入法) inside of the `ime/dict/han.sqlite3` database. The data was taken from [this repo](https://github.com/acevery/ibus-table-zhengma) although the data is now in the public domain.

The plan is to add the rest of the other shape based layouts inside of this database and select the appropriate table to suggest from using the secondary locales of the subtypes. e.g. `zh-zhengma`, `zh-cangjie`, etc...

One thing to note is that spaces are disabled in the PR, as CJK suggestions do not use spaces. However, the default keyboard layouts add spaces after committing suggestions via the smart menu or the spacebar.

So far, word frequency is not implemented, although it should be easy to add. Other than the above, this PR is usable as is for Chinese input, ~~although it bricks everything else because of spaces~~

Here's a video of it in action

https://user-images.githubusercontent.com/5228243/188711969-f28dd84e-ec78-4daa-a7ae-cccc4254c4bb.mp4

PR #2093, opened 2022-10-24
Fixes to Chinese shape-based layout: mixed language support

moonbeamcelery:
This PR is based on #2054 and adds some quality-of-life updates for using the new subtype to type a mixture of Chinese and English text.

- Allowing committing raw input using Enter with CJK, and allowing the committed text to persist instead of being included in the composing range again. ~See below for caveat.~
- Reset suggestions when changing keyboard subtypes.
- ~Delegate NlpProvider to determine composing range when starting IME or moving cursor. Latin behavior remains considering the last word, and HanShapeBased only consider viable characters (lowercase letters for now, but may need to add support a different set for each input method).~ (already merged)
- Update: tracks the last commit position, and limits `HanShapeBasedLanguageProvider` to not compose beyond that position.
- Force suggestion on if a subtype uses HanShapeBased, so when people first use the IME they would not be confused to think the IME is broken if suggestion is off.

Some extra fixes:
- Full-width comma for CJK keyboard, which seems to be missing.
- Allow minimal handling of physical keyboard input (space and enter). Full-width characters, tab to select candidate etc. are still missing.

~Caveat: these work now, but I'm currently using the previous composing range as a reference for changing the composing range. For some use cases, this breaks the composing range preservation due to how Android handles an zero-length composing range when calling onUpdateSelection (e.g. deleting the composing range forces re-evaluation of composing range. So if you type abc and hit Enter, abc gets commited, and then you type def and three backspaces, abc becomes the composing range again despite them being committed). I am not exactly sure if this is the best way to go about it.~

~Maybe I should just maintain an internal record of where the last committed text ended? But the character limit of context may make this hard.~

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt
@@ -146,6 +146,7 @@ class KeyboardManager(context: Context) : InputKeyEventReceiver {
             subtypeManager.activeSubtypeFlow.collectLatestIn(scope) {
                 reevaluateInputShiftState()
                 updateActiveEvaluators()
+                editorInstance.refreshComposing()
                 resetSuggestions(editorInstance.activeContent)
             }
             clipboardManager.primaryClipFlow.collectLatestIn(scope) {
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt
@@ -57,6 +57,7 @@ import dev.patrickgold.florisboard.ime.text.keyboard.TextKeyboardCache
 import dev.patrickgold.florisboard.lib.android.showLongToast
 import dev.patrickgold.florisboard.lib.android.showShortToast
 import dev.patrickgold.florisboard.lib.devtools.LogTopic
+import dev.patrickgold.florisboard.lib.devtools.flogDebug
 import dev.patrickgold.florisboard.lib.devtools.flogError
 import dev.patrickgold.florisboard.lib.ext.ExtensionComponentName
 import dev.patrickgold.florisboard.lib.kotlin.collectIn
@@ -507,6 +508,21 @@ class KeyboardManager(context: Context) : InputKeyEventReceiver {
     }
 
     /**
+     * Handles a hardware [KeyEvent.KEYCODE_SPACE] event. Same as [handleSpace],
+     * but skips handling changing to characters keyboard and double space periods.
+     */
+    fun handleHardwareKeyboardSpace() {
+        val candidate = nlpManager.getAutoCommitCandidate()
+        candidate?.let { commitCandidate(it) }
+        // Skip handling changing to characters keyboard and double space periods
+        // TODO: this is whether we commit space after selecting candidate. Should be determined by SuggestionProvider
+        if (!subtypeManager.activeSubtype.primaryLocale.supportsAutoSpace &&
+                candidate != null) { /* Do nothing */ } else {
+            editorInstance.commitText(KeyCode.SPACE.toChar().toString())
+        }
+    }
+
+    /**
      * Handles a [KeyCode.SPACE] event. Also handles the auto-correction of two space taps if
      * enabled by the user.
      */
@@ -533,6 +549,7 @@ class KeyboardManager(context: Context) : InputKeyEventReceiver {
                 }
             }
         }
+        // TODO: this is whether we commit space after selecting candidate. Should be determined by SuggestionProvider
         if (!subtypeManager.activeSubtype.primaryLocale.supportsAutoSpace &&
                 candidate != null) { /* Do nothing */ } else {
             editorInstance.commitText(KeyCode.SPACE.toChar().toString())
@@ -803,6 +820,21 @@ class KeyboardManager(context: Context) : InputKeyEventReceiver {
         val devtoolsEnabled = prefs.devtools.enabled.get()
         activeState.batchEdit {
             activeState.debugShowDragAndDropHelpers = devtoolsEnabled && prefs.devtools.showDragAndDropHelpers.get()
+        }
+    }
+
+    fun onHardwareKeyDown(keyCode: Int, event: KeyEvent?): Boolean {
+        flogDebug { "Calling from FlorisImeService.onKeyDown with $keyCode" }
+        when (keyCode) {
+            KeyEvent.KEYCODE_SPACE -> {
+                handleHardwareKeyboardSpace()
+                return true
+            }
+            KeyEvent.KEYCODE_ENTER -> {
+                handleEnter()
+                return true
+            }
+            else -> return false
         }
     }
 
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt
@@ -35,6 +35,7 @@ import dev.patrickgold.florisboard.ime.ImeUiMode
 import dev.patrickgold.florisboard.ime.core.DisplayLanguageNamesIn
 import dev.patrickgold.florisboard.ime.core.Subtype
 import dev.patrickgold.florisboard.ime.core.SubtypePreset
+import dev.patrickgold.florisboard.ime.editor.EditorContent
 import dev.patrickgold.florisboard.ime.editor.FlorisEditorInfo
 import dev.patrickgold.florisboard.ime.editor.ImeOptions
 import dev.patrickgold.florisboard.ime.editor.InputAttributes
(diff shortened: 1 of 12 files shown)


patrickgold:
Thanks a lot for your PR and the fixes!

So if I see correctly this PR incorporates all changes from #2054 and your fixes? (So all fixes from your sub-PRs to Waelwindows' branch?). I am currently not aware of the status from @Waelwindows on this. If he has time it would be better to first add your changes to his branch and then use his PR for the actual merging into FlorisBoard. Else if he's ok with it we can proceed with this PR as well, as it preserves the original commit history and author identity.

> Caveat: these work now, but I'm currently using the previous composing range as a reference...

On a system level there can only be 1 active composing range at a time, and as the active composing range also tells Android which range to underline the current word must be the current composing range. If you need the previous composing range we need to have an internal history.

moonbeamcelery:
Thanks for the feedback Patrick! @Waelwindows pulled some of my fixes until [22671e7](https://github.com/florisboard/florisboard/pull/2093/commits/22671e7756a2868ba2ea94cb68ee12774c7f0e23), but for the later ones that are more complicated and less crucial, [it is suggested that I put the PR here to get feedback](https://github.com/Waelwindows/florisboard/pull/3).

So I think the plan is to first help getting his PR done, and then this one can be rebased and reviewed for real.

> If you need the previous composing range we need to have an internal history.
I see. In this draft, Android would tell me the current composing range, which will be used to determine the next composing range. It mostly works. But when I tried to tell Android that I wanted the next composing range to be zero-length at a specific location, next time when `onUpdateSelection` is called, Android just tells me the composing range is set to (-1, -1). 

I'll just try to record a internal history then.

moonbeamcelery:
Update: drafted a version that tracks the last commit position, and limits `HanShapeBasedLanguageProvider` to not compose beyond that position.

patrickgold:
> So I think the plan is to first help getting https://github.com/florisboard/florisboard/pull/2054 done, and then this one can be rebased and reviewed for real.

Ok this sounds like a plan. Would it be ok for you to assist me in the review of the original PR? You seem to have tested it quite thoroughly and know what might be a problem (excluding the things that this PR addresses).

> Update: drafted a version that tracks the last commit position, and limits HanShapeBasedLanguageProvider to not compose beyond that position.

Had a quick look at your code and the implementation looks pretty clean I must say! Although I have to test it more thoroughly to ensure the history does not interfere with the normal logic.

moonbeamcelery:
> Would it be ok for you to assist me in the review of the original PR? You seem to have tested it quite thoroughly and know what might be a problem (excluding the things that this PR addresses).

Yes, I can provide feedback on what I think is missing or problematic on that PR, if that is what you are talking about. Or if there is anything else you need help with regarding these PRs, let me know.

I have been using a version with some other fixes. I can change to using this version instead and test it.

patrickgold (commented):
As #2054 is now finally merged in I will now work towards getting this PR to also be mergeable. Before I do the code review and testing though, is this PR up-to-date or are some of the improvements already in #2054?

moonbeamcelery:
Thanks a lot for the merge! A few changes were included into #2054 , but others not. I'll rebase this and my other PRs soon.

app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt
```diff
@@ -146,6 +146,7 @@ class KeyboardManager(context: Context) : InputKeyEventReceiver {
             subtypeManager.activeSubtypeFlow.collectLatestIn(scope) {
                 reevaluateInputShiftState()
                 updateActiveEvaluators()
+                editorInstance.refreshComposing()
                 resetSuggestions(editorInstance.activeContent)
```
moonbeamcelery, line 150: might compute `.suggest()` of suggestion provider twice?

moonbeamcelery:
There seems to be a giant lag when switching to a HanShapeBased subtype, especially if its table is from an extension. Need to resolve.

Waelwindows:
That could be due to `getListOfWords` and co. Could you try stubbing them and checking the performance again?

moonbeamcelery:
Seems to be caused by that, thanks for the pointer. I'll stub it in this PR.

app/src/main/kotlin/dev/patrickgold/florisboard/ime/nlp/NlpManager.kt
```diff
@@ -174,12 +175,26 @@ class NlpManager(context: Context) {
...
+        textBeforeSelection: CharSequence, breakIterators: BreakIteratorGroup, localLastCommitPosition: Int
+    ): EditorRange {
         return getSuggestionProvider(subtypeManager.activeSubtype).determineLocalComposing(
-            subtypeManager.activeSubtype, textBeforeSelection, breakIterators
+            subtypeManager.activeSubtype, textBeforeSelection, breakIterators, localLastCommitPosition
         )
     }
 
+    fun providerForcesSuggestionOn(subtype: Subtype): Boolean {
+        // Using a cache because I have no idea how fast the runBlocking is
```
patrickgold, line 187: runBlocking itself has a very minimal overhead, its actual speed is highly dependend on the function(s) you call from within the block, so in this case `getSuggestionProvider()`

patrickgold (changes requested):
Thanks a lot for adapting this PR, really appreciate it!

I now thoroughly reviewed your code and have tested it out on my testing rig, and as far as I can see almost everything checks out fine.

As for the LANGUAGEPACKS.md file, I think the Chinese packs-related info which is on there should be moved either to a separate file (like CHINESE_INPUT_SUPPORT.md) or to an external repo, as this repo's language is English only throughout all documentary files, and the file looks a bit out-of-place in its current form.

Also one thing I noticed is that backspace does not rerun the suggestion cycle once the user has inputted something that triggered zero suggestions. For instance input "hh", which provides some suggestions. Then add another "h", and there are correctly 0 suggestions. However pressing backspace to delete the additional "h" does not give me any suggestions again. Is this intended behavior or a bug? (Note that this behavior was already there in the original PR, however I only actively noticed it in this PR).

> Force suggestion on if a subtype uses HanShapeBased, so when people first use the IME they would not be confused to think the IME is broken if suggestion is off.

Really like this change tbh, nice idea of yours :)

moonbeamcelery:
Hi Patrick, thanks for reviewing the code, much appreciated!

> I think the Chinese packs-related info which is on there should be moved either to a separate file (like CHINESE_INPUT_SUPPORT.md) or to an external repo

What would you like to do? Do you want to host those extensions in another repo that you control, or do you wish I host one for Chinese? Note that there is no guarantee that I will be actively maintaining things I host as time goes on.

> backspace does not rerun the suggestion cycle once the user has inputted something that triggered zero suggestions

I took a closer look today, and it turns out this issue will affect Latin too in the future. The suggestion provider does give suggestions, but if you have the smartbar with suggestions & actions shared, and auto-expand/collapse on, then the empty suggestion will auto-collapse the smartbar. [Because delete is a "repeatable code", deleting never re-expands it.](https://github.com/florisboard/florisboard/blob/master/app/src/main/kotlin/dev/patrickgold/florisboard/ime/nlp/NlpManager.kt#L314) If you manually expand it after deleting, you will see the suggestions again.

What do you say I [remove `KeyCode.DELETE` and `KeyCode.FORWARD_DELETE` from the `repeatableKeyCodes`](https://github.com/florisboard/florisboard/blob/master/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt#L113-L114)?

patrickgold (changes requested):
> What would you like to do? Do you want to host those extensions in another repo that you control, or do you wish I host one for Chinese? Note that there is no guarantee that I will be actively maintaining things I host as time goes on.

I think it is best to just put it into a different file in the root of this repo for now - this way everyone interested can contribute to it. If I find a better place for it I can always move it later on.

> What do you say I [remove KeyCode.DELETE and KeyCode.FORWARD_DELETE from the repeatableKeyCodes](https://github.com/florisboard/florisboard/blob/master/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/KeyboardManager.kt#L113-L114)?

Nope that has unintended side effects that severely interfere with the gesture logic. I think you are right, this is a deeper problem. When I am integrating the Latin NLP core I will take a closer look on it and try to come up with a solution for it.

So once the Chinese related info is moved into a separate file I think this PR is merge-ready :)

moonbeamcelery:
> I think it is best to just put it into a different file in the root of this repo for now - this way everyone interested can contribute to it. If I find a better place for it I can always move it later on.

Done. But note that the full Fcitx language pack file needs to be hosted somewhere due to licensing, hopefully not in this repo. ~Please don't use my file share~

> When I am integrating the Latin NLP core I will take a closer look on it and try to come up with a solution for it.

A hacky workaround for now could be: for these scenarios to return an one-item suggestion list with an empty string, with "--" as the secondary text. Would you like that?

patrickgold (approved):
Thanks for the changes!

Now everything checks out fine, will merge the changes in. Thanks again for your work!

> Done. But note that the full Fcitx language pack file needs to be hosted somewhere due to licensing, hopefully not in this repo. 

Yep the planned official community repo in the future would probably the best place for hosting external, differently-licensed content.

> A hacky workaround for now could be: for these scenarios to return an one-item suggestion list with an empty string, with "--" as the secondary text. Would you like that?

Tbh I really have to take a closer look on why it behaves that way and then decide what to do. Like I said I'll do it when I integrate the NLP core into the main repo.

moonbeamcelery:
There's a regression that the "force suggestion on" change made suggestion available for password fields for both Latin and HanShapeBased. Need to fix.

PR #2108, opened 2022-11-05
Update workflows dependencies

tsiflimagas:
This should fix some ci warnings. Also added gradle/wrapper-validation-action, I guess it's good to have. I don't know if you prefer another java distribution, if so, let me know.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -20,20 +20,23 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


patrickgold (approved):
Thanks a lot for your changes!

I've reviewed your changes and everything checks out fine as far as I can see.

I really like your idea of adding the Gradle wrapper validation action, that's a good step in ensuring that only official Gradle wrapper jars land in this repo. Not sure why I haven't thought of this myself, considering I added the sha256 sum to the wrapper properties a long time ago.

As for the Setup Java version, I personally do not have a specific preference for the distribution, but based on my quick research your choice is the best and its license is compatible with this project.

As such I am approving your changes and will merge them in, thanks again :)

PR #2132, opened 2022-12-12
Support user dictionary in suggestions

moonbeamcelery:
This PR adds custom-defined dictionary into the suggestions for both Latin and HanShapeBased providers, and the fallback one.

This is ~again based on #2054 and includes changes from #2093. For now, only the last commit is specific to this PR~ rebased on to master.

~TODO:~
- ~Add default language values to the entries based on which screen the user is on.~ :heavy_check_mark: done
- ~Fix bugs in Latin that (1) auto-commits, and (2) matches empty text and non-starting positions (feature or a bug?).~ :heavy_check_mark: done
- ~The added quick action does not show up unless you delete the app's data and start over. Need to fix.~ :heavy_check_mark: done

Update: fixed these issue and added a quick action button to add things to the internal user dictionary.

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/app/AppPrefs.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/app/AppPrefs.kt
@@ -684,21 +684,32 @@ class AppPrefs : PreferenceModel("florisboard-app-prefs") {
                 val oldArrangement = QuickActionArrangement.Serializer.deserialize(entry.rawValue)
                 val newDefault = QuickActionArrangement.Default
 
-                val newStickyAction = oldArrangement.stickyAction?.takeIf { newDefault.contains(it) }
-                val newDynamicActions = oldArrangement.dynamicActions.filter { newDefault.contains(it) }.toMutableList()
-                val newHiddenActions = oldArrangement.hiddenActions.filter { newDefault.contains(it) }.toMutableList()
+                // FIXME: what is the canonical equality for two quick actions? Using .keyData().code for now.
+                val oldActionKeys = buildSet {
+                    oldArrangement.stickyAction?.let { add(it.keyData().code) }
+                    oldArrangement.dynamicActions.forEach { add(it.keyData().code) }
+                    oldArrangement.hiddenActions.forEach { add(it.keyData().code) }
+                }
+                val newActionKeys = buildSet {
+                    newDefault.stickyAction?.let { add(it.keyData().code) }
+                    newDefault.dynamicActions.forEach { add(it.keyData().code) }
+                    newDefault.hiddenActions.forEach { add(it.keyData().code) }
+                }
+                val newStickyAction = oldArrangement.stickyAction?.takeIf { it.keyData().code in newActionKeys }
+                val newDynamicActions = oldArrangement.dynamicActions.filter { it.keyData().code in newActionKeys }.toMutableList()
+                val newHiddenActions = oldArrangement.hiddenActions.filter { it.keyData().code in newActionKeys }.toMutableList()
                 newDefault.stickyAction?.let {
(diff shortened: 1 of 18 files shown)


moonbeamcelery:
TODO: fix bugs in Latin that (1) auto-commits, and (2) matches empty text and non-starting positions (feature or a bug?).

Turhvjbufv:
Does it have an estimated finish date?

moonbeamcelery:
Not really, this heavily interacts with the autocorrect and suggestion implementation that Patrick is working on. When that is done, this can be rewritten and become ready.

patrickgold:
> Any suggestions on what to do with the current behavior of matching non-starting positions? e.g. when tiramisu is in the user dictionary, an input of mis would fetch the tiramisu suggestion. Do you want this behavior to stay or can I change it to only match the beginning of an entry?

Tbh if the user enters "mis" he/she probably didn't meant to enter "tiramisu", so it would make sense to match the beginning only. In my currently unrelased NLP core this problem has been addressed differently through proximity matching (but in your example it would not suggest "tiramisu" for "mis" as input either because it is not even close to the beginning of the word), however for the scope of the feature you implement I think it is more than enough to do exact matches on the beginning only.

Turhvjbufv:
> > Any suggestions on what to do with the current behavior of matching non-starting positions? e.g. when tiramisu is in the user dictionary, an input of mis would fetch the tiramisu suggestion. Do you want this behavior to stay or can I change it to only match the beginning of an entry?
> 
> Tbh if the user enters "mis" he/she probably didn't meant to enter "tiramisu", so it would make sense to match the beginning only. In my currently unrelased NLP core this problem has been addressed differently through proximity matching (but in your example it would not suggest "tiramisu" for "mis" as input either because it is not even close to the beginning of the word), however for the scope of the feature you implement I think it is more than enough to do exact matches on the beginning only.

Maybe make it an option to toggle it on/off

moonbeamcelery:
Update: fixed these issue and added a quick action button to add things to the internal user dictionary. The added quick action does not show up unless you delete the app's data and start over. Need to fix.

app/src/main/kotlin/dev/patrickgold/florisboard/app/AppPrefs.kt
```diff
@@ -679,6 +680,48 @@ class AppPrefs : PreferenceModel("florisboard-app-prefs") {
             -> {
                 entry.transform(rawValue = entry.rawValue.uppercase())
             }
+            "smartbar__action_arrangement" -> {
+                val oldArrangement = QuickActionArrangement.Serializer.deserialize(entry.rawValue)
+                val newDefault = QuickActionArrangement.Default
+
+                // FIXME: what is the canonical equality for two quick actions? Using .keyData().code for now.
```
moonbeamcelery, line 687: Note that for QuickActionArrangement setting migration, we add newly available action tiles. But the correspondence is determined by `.keyData().code`, as `QuickActionArrangement.contains()` seems to compare by object reference or something

moonbeamcelery:
This PR technically is ready to merge if I do not find any issue in the next few days. But it may be more convenient to merge after the suggestion engine is ready and after adapting this PR to that version.

daakru:
@moonbeamcelery is there any plan to take suggestions from the system user dictionary as well?

moonbeamcelery:
> @moonbeamcelery is there any plan to take suggestions from the system user dictionary as well?

FlorisBoard supported that before, and this PR enables that as well. I just didn't add a quick action button for it.

patrickgold:
As this PR is stale for a long time now and a lot has changed in the time since then, this PR will be closed. With the v0.5 NLP user dictionary support will be natively supported in the new algorithm. Still thanks for your contribution!

PR #2163, opened 2023-02-24
Add New Kabyle (Taqbaylit) Popup Mapping

yanis867:
Added support for Kabyle (Latin script) with the default qwerty layout


yanis867:
I copied the Gboard "Tamazight (Latin) QWERTY" layout, tested it and looks pretty close. Maybe I'll add Tifinagh layout later.

PR #2200, opened 2023-04-25
Add neo keyboard layout variant: mine

manuelmessner:
Thank you very much for your great work on this project!

This patch adds the current version of neo's `mine` layout. It is based on the the `bone.json` and the values are adapted according to https://neo-layout.org/Layouts/mine/.

Notes:

1. This layout misses the last key on the middle row, as that's an accent key in mine's original version.
   How are cases like this handled in other layouts?
1. There is a `neobone` localization (extension). 
   As `mine` also belongs to the neo family do I also need to create an `neomine` extension?


patrickgold:
Hi there!

PR #2213, opened 2023-05-10
Implement support for Material You color selection in theme editor

### LINKED ISSUE #1371: Dynamic color [a.k.a Material You] support
Add an option to theme the keyboard with Android 12's dynamic colors. GBoard already has this feature, but I would rather want to use FlorisBoard. This is optional, but you could also implement Material Design 3 elements too [Pill shapes & circles instead of rounded rectangles].

Here's a screenshot of GBoard's dynamic color theme:
gboard

Thanks!

### LINKED ISSUE #2046: Use MonetCompat  to implement Material You Support
Hi!  I am not really a dev so i dont know a lot but i just found this monetcompat developed by Kieron Quinn https://github.com/KieronQuinn/MonetCompat
and he implements this to his apps to have a material you style interface that automatically matches the wallpaper on Android version lower than A12 that does not implement material you theming and its very good cause my phone is Android 10. I would like to suggest using this to generate automatic material you theming on florisboard themes to support monet on themes on Android version lower than A12.

Hoping for a kind response.  Thank you!

1fexd:
This PR introduces support for Material You colors in the theme editor. This means all colors typically available to Material You applications generated from the device wallpaper are available as values to select in the them editor. 

This works by creating a new `SnyggValue` which stores the name of the color, which is then dynamically evaluated when rendered in the theme editor as well as in the keyboard.

Closes #1371 
Closes #2046


patrickgold (commented):
Thanks a lot for your contribution, really appreciate it!

As I am currently actively busy implementing the NLP core into the main project I wasn't able to fully review and test your changes on a real device yesterday. However I had a quick look at the code changes and I really like your approach of you adding a new SnyggValue and also how you dynamically compute the color.

I will fully review your changes this weekend, but it already looks very promising and it will probably just need minor fixes before we can merge in your changes :)

app/src/main/kotlin/dev/patrickgold/florisboard/lib/snygg/value/SnyggMaterialYouValue.kt
```diff
@@ -0,0 +1,209 @@
+/*
+ * Copyright (C) 2021 Patrick Goldinger
```
patrickgold, line 2: For new files the copyright year should be the current, so 2023:

```suggestion
 * Copyright (C) 2023 Patrick Goldinger
```

app/src/main/kotlin/dev/patrickgold/florisboard/lib/snygg/value/SnyggValue.kt
```diff
@@ -105,8 +107,12 @@ object SnyggImplicitInheritValue : SnyggValue, SnyggValueEncoder {
     override fun encoder() = this
 }
 
-val SnyggVarValueEncoders = listOf(
+val isSOrAbove = Build.VERSION.SDK_INT >= Build.VERSION_CODES.S
+
+val SnyggVarValueEncoders = listOfNotNull(
     SnyggSolidColorValue,
+    if (isSOrAbove) SnyggMaterialYouLightColorValue else null,
+    if (isSOrAbove) SnyggMaterialYouDarkColorValue else null,
```
patrickgold, line 115: For version checking we can use AndroidVersion:

```suggestion
val SnyggVarValueEncoders = listOfNotNull(
    SnyggSolidColorValue,
    if (AndroidVersion.ATLEAST_API31_S) SnyggMaterialYouLightColorValue else null,
    if (AndroidVersion.ATLEAST_API31_S) SnyggMaterialYouDarkColorValue else null,
```

gradle/libs.versions.toml
```diff
@@ -6,6 +6,7 @@ androidx-activity = "1.5.1"
 androidx-autofill = "1.1.0"
 androidx-collection = "1.2.0"
 androidx-compose = "1.4.3"
+androidx-compose-material3 = "1.1.0-rc01"
```
patrickgold, line 9: Latest stable release is `1.1.0`:

```suggestion
androidx-compose-material3 = "1.1.0"
```

patrickgold (changes requested):
I finally had the time to properly review it and I am really impressed how well it works, so again thanks a lot!

However, during my testing and review of the code I found some issues and bugs that need to be addressed before merging, some issues I will mention in this general review comment and for some issues I will directly comment in-place.

Something that I think is pretty annoying is the fact that we now need to pass `context` around so much. I've looked into it and also played around with removing the context attribute, however I wasn't able to come up with a clever solution to get around this. The only thing that I could imagine is that we call `LocalContext.current` directly within the functions, however I don't know if this is a good solution. What's your opinion about this?

Also in the color name dropdown within the color edit dialog the attribute name `background` is incorrectly displayed as `Background`, however after searching your changes I wasn't able to determine why (especially because all other color names correctly start with lowercase).

patrickgold:
Some additional thoughts I just had: What do we do with themes that contain dynamic color references and which are then imported on devices that run Android 11 or earlier? I haven't tested this out yet but does your implementation fall back to a default color or does it crash the app?

1fexd:
> Something that I think is pretty annoying is the fact that we now need to pass `context` around so much. I've looked into it and also played around with removing the context attribute, however I wasn't able to come up with a clever solution to get around this. The only thing that I could imagine is that we call `LocalContext.current` directly within the functions, however I don't know if this is a good solution. What's your opinion about this?

I think passing around `context` is just something that just has to be done, I don't think it should be coupled (that would also require `@Composable` annotations)

1fexd:
> Some additional thoughts I just had: What do we do with themes that contain dynamic color references and which are then imported on devices that run Android 11 or earlier? I haven't tested this out yet but does your implementation fall back to a default color or does it crash the app?

I just checked, and your error handling appears to work very well, it appears to be falling back to an empty/transparent color? However, in the "Variables" section of the theme editor the property value is replaced by its package (https://github.com/1fexd/florisboard/blob/feature/material-you-theme-colors/app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/theme/ThemeTranslations.kt#L173).

How should we handle this? Maybe just display something like "Material You properties are only supported on Android 12+" for each Material You property in the theme editor and leave the empty/transparent color it appears to be falling back as is?

patrickgold (approved):
Thanks a lot for applying the suggested changes and fixes! I've retested your changes and now the color is serialized correctly and it generally works really well now.

> I think passing around context is just something that just has to be done, I don't think it should be coupled (that would also require @Composable annotations)

Yeah I think we just have to live with that fact.

> I just checked, and your error handling appears to work very well, it appears to be falling back to an empty/transparent color? However, in the "Variables" section of the theme editor the property value is replaced by its package

There's a fallback property yep, for anything that isn't detected as a valid variable/value/function. As you wrapped the Material You encoders in AndroidVersion checks these `dynamic-dark-color()` statements are seen as invalid on Android 11 devices and earlier, so it just uses that fallback color.

> How should we handle this? Maybe just display something like "Material You properties are only supported on Android 12+" for each Material You property in the theme editor and leave the empty/transparent color it appears to be falling back as is?

We could do that, however for this to work we need to allow Material You values to be loaded internally on Android 11- devices (as opposed to be seen as invalid functions) and then within the Material You color logic prevent the statement from doing anything to avoid crashes.

I think the easier approach would be to just let the theme editor overwrite it to the fallback color, and themes that use dynamic system colors must explicitly state that its only designed for Android 12 or newer. Or maybe I could add an Android version constraint to the extension meta for this, so users cannot import themes designed for newer Android versions on older devices.

Another general idea I had is that we can add 2 theme presets, one for day and one for night theme mode, which automatically has all Material You values pre-used. I don't mind adding them in a separate PR later on, except if you want to add them by yourself.

Apart from that I think I don't have anything to add to this pull request, so if there's nothing to add from your side I think we can merge this PR in.

tsiflimagas:
Hi, thanks for this awesome feature! I've already played around with it and created some gorgeous themes! One thing I wanted to note before this gets merged is that for the colours to be updated on the change of wallpaper or selected system colour scheme, FlorisBoard has to be force stopped, they're not updated on the fly. I'm not sure if that's intended or something that would need too much work thus was left out of this PR, but it felt like it should be noted.

patrickgold (changes requested):
Thanks for pointing out this issue, I just tested it out and it indeed requires a force stop, whereas Gboard for instance immediately changes its colors. I've investigated the changes and made a simple patch:

[0001-Fix-MaterialYou-color-scheme-not-updated-correctly.patch](https://github.com/florisboard/florisboard/files/11521168/0001-Fix-MaterialYou-color-scheme-not-updated-correctly.patch)

I tested out the patch and the colors should now update every time the keyboard window is shown again (meaning if you switch out the wallpaper you will see the changes once you hide the keyboard and tap into any input field again), however I think this is more than enough and minimizes extra resources needed for creating the dynamic color scheme.

@ 1fexd: if possible please test out the patch locally on your device and then push the patch onto the branch so it gets included in this PR, thanks!

1fexd:
> I think the easier approach would be to just let the theme editor overwrite it to the fallback color, and themes that use dynamic system colors must explicitly state that its only designed for Android 12 or newer. Or maybe I could add an Android version constraint to the extension meta for this, so users cannot import themes designed for newer Android versions on older devices.

Maybe we could add a flag for Material You themes like the one that already exists for night themes? 

> Another general idea I had is that we can add 2 theme presets, one for day and one for night theme mode, which automatically has all Material You values pre-used. I don't mind adding them in a separate PR later on, except if you want to add them by yourself.

That sounds like a good idea, but how we'd have to device on a concrete mapping from Florisboard color properties to Material You colors.

patrickgold (commented):
Thanks a lot for applying the patch!

> Maybe we could add a flag for Material You themes like the one that already exists for night themes?

I was gonna say that this is a brilliant idea until I remembered that such flag already (kinda) exists: https://github.com/florisboard/florisboard/blob/deaf3295fd413f9bddd2d11a1ace5d4f00973f80/app/src/main/kotlin/dev/patrickgold/florisboard/ime/theme/ThemeExtensionComponent.kt#L43

The flag is missing from the theme editor atm and I am unsure if it is even used anywhere, however we could use that flag to block theme imports (or at least warn users) when importing a theme which has this flag set. This is still not 100% bullet-proof but it would save us from adding an Android version constraint to the theme json file.

> That sounds like a good idea, but how we'd have to device on a concrete mapping from Florisboard color properties to Material You colors.

That's true, I'd probably "take inspirations" from the external appearance of Gboard's dynamic color usage, although deviating when needed to match the FlorisBoard style.

1fexd:
> The flag is missing from the theme editor atm and I am unsure if it is even used anywhere, however we could use that flag to block theme imports (or at least warn users) when importing a theme which has this flag set. This is still not 100% bullet-proof but it would save us from adding an Android version constraint to the theme json file.

I should I try implementing this?



> That's true, I'd probably "take inspirations" from the external appearance of Gboard's dynamic color usage, although deviating when needed to match the FlorisBoard style.

I built this real the other day, and I kinda like it

image

patrickgold:
> I should I try implementing this?

If you want yes.

> I built this real the other day, and I kinda like it

Doesn't look bad I must say. I think though that we can add these dynamic color preset themes in a separate PR and keep the focus of this PR on the actual implementation of dynamic coloring.

1fexd:
Hm, how should we implement the `isMaterialYouAware`? Automatically set it to `true` if the user uses a Material You color in their theme? Or should we allow it to be set in the "Edit theme component" screen (using the metadata edit dialog at the top) and only display Material You color values if it is set to `true`?

Turhvjbufv:
I think it would be nice if the user could turn it off manually so second option

patrickgold:
Tbh automatically deriving it from the rules is the better option, else the flag is kinda useless, as it could indicate something completely different than what is present in the stylesheet itself.

TeaEndsAcronyms:
I really like this keyboard and want to see material you implemented!

patrickgold:
> Hey, I have been kinda busy with my own projects lately. Just wanted to let you know I am not sure when I will have time/motivation to implement this, so maybe you could go ahead and implement it yourself so this can finally be merged.

No worries, thanks for letting me know. I think for the first few alphas I will just try to see what happens if we don't have the flag at all, and if it causes issues implement a mechanism for the older devices to prevent issues.

I will merge in your changes as is, so thanks a lot again for your effort, this is a really valuable contribution :) Note that as for merging I need to coordinate the merge with my big PR so to try to minimize conflicts it might still be a bit of time until I finally do the merge, but it will definitely be before the 0.4.0-alpha05 release.

> I really like this keyboard and want to see material you implemented!

Next alpha release and you can try it out yourself :)

TeaEndsAcronyms:
Excited! :)

TheHawaiianKoala:
I have created a dark theme and light theme following the design scheme of the Gboard. 
Dark
Light

TheHawaiianKoala:
Sorry for the notification again, I have updated the light theme to look more like the new gboard theme and changed some colors here and there

Attached are multiple screenshots, but while I was taking the screenshots, I encountered a bug. Whenever you change the wallpaper, the suggestion row and navbar don't change their color

Bug

<details markdown='1'><summary>Click to Expand/Collapse</summary>

Dark - 1
Light - 1
Dark - 2
Light - 2
Dark - 3
Light - 3
Dark - 4
Light - 4

</details>

1fexd:
Wow thoese look amazing

TeaEndsAcronyms:
## I've also made one for a pastel pink color! (as well as a borderless version!)
### [Github link](https://github.com/TeaEndsAcronyms/Gboard-for-Florisboard-theme)

## Bordered:
Bordered

---
## Borderless:
Bordered

TheHawaiianKoala:
I forgot about the borderless variants, but after seeing your theme, I updated the dynamic theme with borderless variants aswell.

1fexd:
> ## I've also made one for a pastel pink color! (as well as a borderless version!)
> ### [Github link](https://github.com/TeaEndsAcronyms/Gboard-for-Florisboard-theme)
> ## Bordered:
> Bordered
> ## Borderless:
> Bordered

Always so inspiring to see what people are able to come up with by building on the stuff you kind of laid the "groundwork" for, this is going to be my new theme - thank you so much.

TeaEndsAcronyms:
@1fexd Wow, this made my day!

patrickgold (approved):
Finally get around to merging in this PR. There's currently a bug in the base repo where sometimes themes aren't properly calculated, however this bug was only introduced after the latest commit was done at the time, thus I am approving the changes fully and will merge in the changes.

This feature will be publicly available in v0.4.0-alpha05

D3SOX:
Hm some icons have a different color for me. Any idea why this might happen?

image

D3SOX:
Ok nevermind I think that's correct

patrickgold:
@TeaOwO Wow your themes look really amazing, that's really cool! You are a primary example to me why a community extension store is really needed for this project, simply because some of you are so creative in generating amazing user content and this store would massively help in discoverability :)

Also big thanks again to @1fexd for your amazing work and generally everyone here who helped testing the debug artifacts!

itsmartashub:
Really, @TeaOwO?

patrickgold:
@itsmartashub Ah I just see that you are the original author so thanks to you for the amazing themes! To TeaOwO's defense he's declaring it as a fork and I over-read it tbh, but yeah. Will edit the original comment to reflect the correct attribution.

itsmartashub:
Mostly, oh well...
Okay, it seems I don't understand how this works. I thought that when you come across a repo/project that you like and think it can be improved, you open an issue with a suggestion or create a pull request (contributing), rather than completely copying (aka fork) the repo to your own profile, making a few changes to the readme file, and creating releases... And on top of that "self promoting", or whatever.

Anyway, immense thanks to @patrickgold and all the Florisboard contributors, especially @1fexd for the great work you've done together. I can't express how much I look forward to the next update. 🤗

patrickgold:
Well I agree that it can hurt a lot if such a thing happens and honestly I don't know why TeaOwO did it, but maybe he will explain.

Generally it must be said that forking is valid and quite common in the open-source world for a lot of reasons, however for small changes like this one a PR to your repo would have been the better path.

itsmartashub:
Hi there, @TeaOwO. If you genuinely wanted to help or share improvement suggestions rather than taking over everything, simply opening an issue (or few) and listing suggestions you have on your mind would have been enough. You'd still be mentioned in releases if that was your goal, and even if it wasn't, I'd never take someone else's credit, even if it's just a change in the shade of the `enter` button.

In this case (keyboard themes/`flex` file) I think even a PR would be inadequate and a bit strange to review, do you agree?

itsmartashub:
@patrickgold I agree with every word you've written. Thanks to the powerful ability to fork, many beautiful and significant things have been created and improved in the FOSS world and beyond. But in this case and in this way, I truly believe there was no need for it.
For such a small change, you wouldn't even need a PR. An issue with the `enhancement` label would suffice.

Anyway, I really wouldn't want you to spend more of your free and precious time on this themes-author saga. I believe you have a lot on your plate, and it's probably better to redirect and invest your energy into that, or into whatever makes you "flourish" 🤗 Happy FlorisBoarding to all.

PR #2229, opened 2023-05-28
Add samsung board inspired layout

nettnikl:
The layout i'm used to.
Shout out to https://onlineutf8tools.com/convert-utf8-to-code-points


patrickgold (changes requested):
Thanks a lot for your changes, really appreciate it!

The layout itself looks good, just the naming is not the best yet. Please change the id of the layout to "western_samsung" and its label to "Western (Samsung)" respectively, then I see no problem merging the symbol layouts in.

nettnikl:
Thanks for your quick review!
Just one question that came up during usage: 
Would you recommend to use in the part with the game-card-symbols:
Medium Lozenge „⬧“ (U+2B27) + „⬨“ (U+2B28) - which are used by Samsung, or
Diamond Suit „♦“ (U+2666) + „♢“ (U+2662) - which are the actual card-game-symbols (also, U+2666 is also shown as colored emojy in messengers)?

app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
```diff
@@ -679,6 +679,12 @@
         "label": "Western (Additional symbols)",
         "authors": [ "patrickgold" ],
         "direction": "ltr"
+      },
+      {
+        "id": "sboard",
+        "label": "Samsung",
```
patrickgold, line 685: You forgot to apply the renaming here :)

patrickgold (changes requested):
Thanks for the changes, we are almost there (see specific line comment)

> Would you recommend to use in the part with the game-card-symbols:
Medium Lozenge „⬧“ (U+2B27) + „⬨“ (U+2B28) - which are used by Samsung, or

I mean this layout is basically a replica of Samsung's so it would make sense to use the symbol Samsung uses, but I will leave the final decision to this to you.

app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
nettnikl: How did i even miss that.. Corrected

nettnikl:
I'd keep the samsung original nonsense then and just locally customize the better fitting symbol in. For anyone who is bothered, they can find the better suited symbols in this thread, and do the same.

PR #2253, opened 2023-06-29
Add Arabic Kaf to the Persian popup

sadegh-msa:
The Arabic Kaf (code: 1603) is part of the standard Persian
keyboard (ISIRI 9147), but it was missing in the current layout.

This commit adds the Arabic Kaf (code: 1603) to the popup of the Persian Kaf
(code: 1705) key, following the convention of other keys that have
both Arabic and Persian variants.
This improves the backward compatibility and accessibility of
the layout for users who need to type both scripts.


patrickgold (approved):
Thanks a lot for the changes, they look good, merging into master.

sadegh-msa:
Thank you.

PR #2267, opened 2023-07-14
Fix uppercase behaviour with telex

### LINKED ISSUE #2260: [Vietnamese Telex] Uppercase error in Vietnamese Telex
#### Short description
When add tone to Uppercase letter, the whole word become the uppercase letter:

+ "Ans" -> "ÁN", should be: "Án"
+ "Ais" -> "ÁI", should be: "Ái"
+ "Aif" -> "ÀI", should be: "Ài"

#### Steps to reproduce
1. Chose Vietnamese Keyboard
2. Click "Try out your setup"

#### Environment information
- FlorisBoard Version: 0.4.0-alpha04 (90)
- Install Source: IzzyOnDroid
- Device: Nokia 6.1 plus
- Android: 13 - LineageOS 12

error with Vietnamese Keyboard

Hayleia:
Basically in the current master code, when typing for example Anhs, 'A' is uppercase and this decides that the whole result of the rule will be uppercased (ÁNH), which is most likely not what the user intended. We only want the first character to be uppercased, so we can replace "value.uppercase()" with "value.replaceFirstChar(Char::titlecase)".
However, this fix would also mean that typing ANHS gives Ánh as a result, which is also most likely not what the user intended.

So the proper fix is a better case detection and management, such as
- If every character was uppercase, then uppercase the whole result (ANHS→ÁNH).
- Else if only the first character was uppercase, then only uppercase the first character in the result (Anhs→Ánh).
- Else if everything was lowercase, keep everything lowercase (anhs→ánh).
- Else, well this is a weird case and I guess anything is acceptable for such a rare case? (aNhs→ánh for example).

Closes #2260.

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/text/composing/Composer.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/text/composing/Composer.kt
@@ -58,8 +58,13 @@ class WithRules(
         for (key in ruleOrder) {
             if (str.lowercase().endsWith(key)) {
                 val value = rules.getValue(key)
-                val firstOfKey = str.takeLast(key.length).take(1)
-                return (key.length - 1) to (if (firstOfKey.uppercase() == firstOfKey) value.uppercase() else value)
+                val matchingPart = str.takeLast(key.length) // the part of the string that matches the key
+                val firstOfKey = matchingPart.take(1)
+                return (key.length - 1) to when {
+                    matchingPart.uppercase() == matchingPart -> value.uppercase()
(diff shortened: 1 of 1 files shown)


patrickgold:
Hi there!

We're closing this PR because the v0.5 development cycle is about to begin, and it will bring significant changes to the text logic, including a lot of new behavior. I'd recommend after the text logic has stabilized in the v0.5 betas and telex behavior is still broken that you open a new PR and we then handle the fix in a timely manner. Let me know if you have any input what we should specifically take care of for the new text logic for telex.

thanhhocse96:
After 0.5 beta this issue still on

PR #2275, opened 2023-08-02
Add pink heart

z3roco01:
Adds the pink heart to the emojis after the red heart


patrickgold (commented):
Thanks for your PR. The emoji list is an automatic export of the Unicode CLDR data using the [emojicon](https://github.com/florisboard/emojicon) tool, as such manual edits to this file should not be done. I will close this PR thus, hope you can understand.

Issue #1125, opened 2021-08-02
Halmak and engram layout request


germe-fur:
i want to use halmak in florisboard, but is not implemented yet.
engram is not too
here is my explantation, i have a mockup here:
https://github.com/dslul/openboard/issues/416

jeremiah-miller:
Halmak is in there, at least in the beta. But keep in mind at least for halmak (and maybe engram), it's optimized for typing on a physical keyboard not a phone screen. I'd recommend using one optimized for phone input like NALMY or Sangaline, but if you still want to add engram you can see how I did it in #1015, it's pretty easy to add a layout

germefur:
hi, i'm the same germe-fur, i lost my account, so i created a new one...

Im in florisboard beta, and the implementation of halmak is OK but a bit ugly
![halmak floris](https://user-images.githubusercontent.com/88954325/129492694-d75c8e07-bc3f-4cb5-adf8-0209d0203802.png)

the last line is compressed, and is very bad to use.

can you change it to a dvorak like implementation?
![dvorak floris](https://user-images.githubusercontent.com/88954325/129492724-94dc87b6-4e95-47bc-8523-cc12e66666b7.png)

the [ , ] and [ . ] is moved, and they dont appear two times, and the first and last letters are moved down

jeremiah-miller:
I'm sure it's possible, idk how to do it though and i'm not sure what the point in it would be since halmak is designed for physical keyboards and wouldn't add much for typing on a phone

germefur:
uhh yes... well qwerty is not designed for mobiles too
but is there...
the F and Y keys need to be swapped, at least in s sub-layout, like "halmak(F Y swap)"

MisterY:
https://github.com/MadRabbit/halmak

Anysoft has a very functional implementation. With the most-common special characters included.

It feels quite spacious although it's the same area of the screen as in Florisboard.

germe-deb:
But... move the [F] and [Y] keys down... XD

that's the solution for the problem

PR #2300, opened 2023-09-25
Add Engram layouts

nothingnesses:
Addresses #1125. I didn't include the symbols in the middle columns since they make the layout too crowded for use on mobile and can be accessed in the symbols layer anyway.

Edit: I added the original layout, in addition to the layout sans the symbols.


bb010g:
Could the primary layout including symbols also be provided as its own layout, for use on devices that are wide enough?

nothingnesses:
@bb010g Hey, thanks for the suggestion! I pushed [a change](https://github.com/florisboard/florisboard/commit/669109f2c341d04448e810457cfe3dcaf2e1701f) where I made separate files for the original full layout and my reduced layout.

patrickgold:
Hi there!

PR #2335, opened 2023-11-25
Add german variant of the dvorak layout

msrd0:
Screenshot_2023-11-25-16-28-18


patrickgold (approved):
Thanks a lot for the changes, they look good, merging into master.

PR #2339, opened 2023-12-09
Upgrade traget SDK to 34 and cleanup code base

patrickgold:
This PR upgrades the target SDK to 34 to comply with Google's guidelines. Additionally we migrate to using non-transitive R-classes, other dependencies are upgraded and some minor cleanup is done.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -29,11 +29,20 @@ plugins {
     alias(libs.plugins.mikepenz.aboutlibraries)
 }
 
+val projectMinSdk: String by project
+val projectTargetSdk: String by project
+val projectCompileSdk: String by project
+val projectBuildToolsVersion: String by project
+val projectNdkVersion: String by project
+val projectVersionCode: String by project
+val projectVersionName: String by project
+val projectVersionNameSuffix: String by project
+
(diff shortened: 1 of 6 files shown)


app/build.gradle.kts
```diff
@@ -29,11 +29,20 @@ plugins {
...
+val projectVersionName: String by project
+val projectVersionNameSuffix: String by project
+
 android {
     namespace = "dev.patrickgold.florisboard"
-    compileSdk = 33
-    buildToolsVersion = "33.0.2"
-    ndkVersion = "26.1.10909125"
+    compileSdk = projectCompileSdk.toInt()
+    buildToolsVersion = projectBuildToolsVersion
```
tsiflimagas, line 44: I remember having seen [this comment](https://github.com/TeamNewPipe/NewPipe/pull/8986#discussion_r988658580), suggesting setting a specific version for build tools isn't really needed.

patrickgold: Hmm difficult to say, AGP will probably auto-resolve it. However the discussions also mentions that omitting the build tools version might make problems with F-Droid builds so I think just keeping it in is the best thing to do.

Issue #2173, opened 2023-03-15
Expand one-handed mode arrow button to fill the entire space


opusforlife2:
Currently, the arrow to swap between right and left sides is a small button below the 'exit one-handed mode' button. This makes it very difficult to tap the button, especially on larger phones.

Google Keyboard's arrow fills the entire height of the keyboard, which makes it perfect for tapping. Any reason why this wasn't done for Floris?

PR #2340, opened 2023-12-09
Implement #2173

lm41:
Closes: #2173 The arrow-button now extends to the full width and height of the free space in one-handed mode.


patrickgold (approved):
Thanks a lot for the quick and effective fix. Tested it locally and it works flawlessly. Merging into master.

Just a suggestion because we discussed this yesterday: the commit name could have been better like "Fix one-handed mode not extending to bottom" or something in that direction, but like discussed yesterday it isn't enforced by me.

PR #2354, opened 2023-12-31
Add ЈЦУКЕН interslavic keyboard layout

### LINKED ISSUE #1633: Add an interslavic keyboard layout
There are already some auxiliary languages in FlorisBoard, like Esperanto and Interlingua. I'd also like to have the [Interslavic](https://en.m.wikipedia.org/wiki/Interslavic)  layout.

Here's an example of a community-made interslavic layout for gboard:

IMG_20220311_144056_943

All of those letters already exist in either russian, ukraininan or serbian layouts, but using three different for one purpose is obvioulsy not a solution.

victorbnl:
Closes #1633

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/characters/jcuken_interslavic.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/characters/jcuken_interslavic.json
@@ -1,0 +1,36 @@
+[
+  [
+    { "$": "auto_text_key", "code":  1112, "label": "ј" },
+    { "$": "auto_text_key", "code":  1094, "label": "ц" },
+    { "$": "auto_text_key", "code":  1091, "label": "у" },
+    { "$": "auto_text_key", "code":  1082, "label": "к" },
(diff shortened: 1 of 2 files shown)


victorbnl:
I couldn’t figure out where to add the layout in `extension.json` because the first half of the list seems to be sorted alphabetically, but then the order seems random. So I ended up adding it to the end of the list. But feel free to tell me if I should move it.

kuroya2mouse:
I think that there is no need for a layout for Interslavic language. It would really be better to add the Interslavic language itself so that there are suggestions on top of the letters like with Czech language + QWERTY/Colemak layout.

screen

patrickgold (approved):

app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/characters/jcuken_interslavic.json
```diff
@@ -0,0 +1,36 @@
+[
```
eqoffical, line 1: the "Д" (or "D") letter is missing

eqoffical (changes requested):

PR #2356, opened 2023-12-31
Sort keyboard layouts extensions by label alphabetically

victorbnl:
The current order of keyboard layouts extensions (characters, charactersMod, numeric…) in the `extension.json` file seems random. Some of it is ordered alphabetically, but there are a lot of inconsistencies. And names in different scripts are among those in latin, which does not make much sense when it comes to sorting alphabetically.

This PR sorts each list of the `layouts` section by their children’s label, according to Python’s `list.sort` function. Consequently, this brings all layouts with non-latin names at the end of the list. Note that I strived to keep formatting as it is (e.g. authors lists are still on one line).

I only edited the file to sort it once here, so it won’t be kept sorted, which means manual checks on PRs would still have to be performed to preserve the order. However if you’re interested I can make a GitHub Action that checks on PRs if the order has been altered and either fail or fix it automatically if it has been. The issue with fixing it automatically is it would be hard to keep the current format (authors on one line for instance), but this does not seem too important to me.

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
@@ -11,18 +11,17 @@
   "layouts": {
     "characters": [
       {
+        "id": "azerty",
+        "label": "AZERTY",
+        "authors": [ "patrickgold" ],
+        "direction": "ltr"
+      },
+      {
         "id": "arabic",
         "label": "Arabic",
         "authors": [ "HeiWiper" ],
         "direction": "rtl",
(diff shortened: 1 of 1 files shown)


patrickgold:
Thanks for your efforts!

As per feature map keyboard layouts will be completely re-implemented with v0.5, thus also the way layouts are defined and dynamically sorted (with the new keyboard engine the Settings UI will contain an order based on device settings first and remaining layouts will be sorted alphabetically). A manual reorder is thus not necessary for upcoming versions and introduces unnecessary merge conflicts with open PRs, as such I will reject your PR.

victorbnl:
Nice!

PR #2364, opened 2024-01-10
Add the monochrome icon variants

### LINKED ISSUE #2159: Add monochrome icon
Please support the Themed Icon.
https://developer.android.com/develop/ui/views/launch/icon_design_adaptive#user-theming

Thanks.

lm41:
ic_app_icon_beta_monochrome
ic_app_icon_debug_monochrome
ic_app_icon_stable_monochrome

Closes #2159


patrickgold (commented):
The changes look good, however I don't have an Android 14 device and my emulator refuses to start without crashing so I currently cannot test it out if the icon config works properly.

patrickgold (approved):
Users have confirmed that the monochrome version works on their real Android device. Will let the PR open until tomorrow to see if someone still finds an issue with it (which I don't think) and then merge it into main.

Thanks for the icons, really appreciate it!

lm41:
image
This is the image from my emulator. I've also tested this with my Pixel (only the debug one)

snaik20 (commented):
Changes looks and works as expected. Just one minor comment that the path data looks slighly off when comparing the exisitng foreground icon with the monochrome icon. For example you can compare `ic_app_icon_debug_foreground` and `ic_app_icon_debug_monochrome` to see the difference in pathData.

CC: @patrickgold 
 
Screenshot on Pixel 4a:
Screenshot_20240111-090434

lm41:
> Changes looks and works as expected. Just one minor comment that the path data looks slighly off when comparing the exisitng foreground icon with the monochrome icon. For example you can compare `ic_app_icon_debug_foreground` and `ic_app_icon_debug_monochrome` to see the difference in pathData.


This is caused by the new text for the debug and beta icons.

patrickgold:
Yep the diff in path data is indeed caused by the texts and fully expected. As I've not heard any issues I will now merge the changes in.

Issue #1869, opened 2022-05-21
Use form issue template


Glitchy-Tozier:
- [Example](https://github.com/ventoy/Ventoy/issues/new?assignees=octocat&labels=&template=issue_template.yml&title=%5Bissue%5D%3A+)
- [Docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms)

1. This may help with making people to actually fill out their **Environment information**.
2. Additionally, it might make reporting issues (especially on phones) easier, since it doesn't lead to tons of comments inside your issue-report.

That being said, if this is implemented, we should wait until it is available in the GitHub-app. Currently, issue-templates containing a form may only be edited in the browser.

patrickgold:
Yep I've seen those issue forms before and I think they would be really useful. But as you also said, I want to wait until they are stable and available in the official GitHub app too.

lm41:
#2366

PR #2366, opened 2024-01-11
[Meta]: Switch to issue forms

lm41:
This pull request changes the issue templates to issue forms to avoid "unnecessary" issues and to make it easier for the user to create an issue. I know that the issue forms are still in beta, but I think they have been around long enough to be considered stable.

I don't think that filling out those forms in the github app will be anything implemented soon™ish. 

Closes: #1869


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,23 @@
+name: Feature request / Suggestion
+description: Suggest an idea for this project
+title: "[FEAT]: "
```
patrickgold, line 3: To keep the bracket tagging consistent with Bug and Crash we should titlecase Feat instead of uppercasing it.

```suggestion
title: "[Feat]: "
```

patrickgold (changes requested):
Thanks for the issue templates!

I completely forgot that GitHub had issue forms, also really strange that the app still not supports them. But yeah in general I think using them is the way to go. Had a look at your templates and they look good.

As for the bracket tags in the title: do you think they add good use? I fear that they may clutter the titles, as we already have the labels like bug and proposal. Not sure yet how I should feel about them yet, although in general I don't have anything against them.

.github/ISSUE_TEMPLATE/feature_request.yml
lm41: It's totally up to you if you want the brackets in the title. But you are right that we have labels for that ¯\_(ツ)_/¯

patrickgold: I just had some thoughts about this and think that omitting them is the cleaner way to go, especially as everything else mostly didn't carry bracket tags either and it worked quite well

patrickgold (approved):
Looks good to me now. Merging in.

patrickgold:
Ok for the GitHub app I just tried it out it doesn't support them  but at least it redirects to a browser then, so there's no option that users can create issues without the template anymore.

Issue #2363, opened 2024-01-10
Copied images are randomly rotated when pasting them


realpixelcode:
#### Short description

When copying an image and pasting it, it's rotated after being sent. The issue happened both in Threema and Threema Libre, but I couldn't reproduce it with either other messengers or other keyboards. I have already reported this to the Threema [support](https://threema.ch/en/support-request?i_have_read_the_faq=1) ([*SUPPORT](https://threema.id/*SUPPORT)), but they said it's not a bug in Threema but in FlorisBoard instead.

https://github.com/florisboard/florisboard/assets/52963327/e3b6eff0-7a3f-45c8-bbbc-8f84d5a735c8

#### Steps to reproduce

1. Copy any image to the clipboard.
2. Open a Threema chat and paste the image from FlorisBoard.
3. The image is rotated.

#### Environment information

- FlorisBoard Beta Version: 0.4.0-alpha06
- Install Source: F-Droid
- Device: Fairphone 3
- Android: Android 12, /e/OS

Kasi07:
I can reproduce this in Version 0.3.16 too.

Environment information
-     FlorisBoard Beta Version: 0.3.16
-     Install Source: Github
-     Device: Pixel 7
-     Android: Android 14, GrapheneOS

lm41:
This is really weird. 
I can reproduce this in Version 0.4.0-alpha04 too.
Environment information
-     FlorisBoard Version: 0.4.0-alpha04
-     Install Source: Own Build
-     Device: Pixel 7 Pro
-     Android: Android 14, GrapheneOS

However, the image is rotated correctly in Signal, in MonoclesChat, in SchildiChat (Element) as well as when you click on the image in Threema and go to the image viewer.
In my opinion, this is most likely a Threema bug.

chop-suey:
I can also reproduce this.

- FlorisBoard Version: 0.4.0-alpha04
- Install Source: IzzyOnDroid
- Device: Samsung S20+
- Android: Android 13

As this does not happen with any other keybord i have tested my conclusion is that this is most likely both: a bug in floorisboard _and_ in threema. In some cases threema seems to only apply the rotation to the generated thumbnails (e.g. when the image is sent as a file). But there most be something odd with the images pasted by floorisboard.

patrickgold:
That really seems to be a Threema issue, as you can see in the message send preview the image is correctly formatted, from that point on FlorisBoard does not influence the image anymore, only the target app, in that case Threema

[...]

PR #2369, opened 2024-01-12
Fix random rotated images in threema

lm41:
This PR tries to fix the random image orientation bug in Threema. 

Closes: #2363


patrickgold (commented):
I've had a look at your code and currently am not sure about it. When the clipboard query() gets a request with the projection set to ORIENTATION jet you just return the orientation? I need to take a closer look at the Android docs here, atm I can't say if this behavior is correct or not.

lm41:
I think it is the correct behavior: https://developer.android.com/guide/topics/providers/content-provider-basics#kotlin
In https://github.com/florisboard/florisboard/pull/2369/files#diff-76c7c47d41f9a377918cf21eaf428807fd8e42a1f09ae29f8d7b17d087e47cd4R95-R97 I check if the requested cols are existing in the db and then return the cursor with the requested cols.

patrickgold:
Ok you are correct the projection array is the array of requested cols to return. However you simply join the array to a string and then compare it, you should probably say something more like `if projection contains MediaStore.Images.Media.ORIENTATION then do X`. Also then you need to check if using a joined version of projection can be used for the SELECT list statement

patrickgold (commented):
Thanks a lot for the fixes. I have reviewed your approach and this version is now much cleaner and better imo. I don't have threema so I can't really test out this fix, should we merge it in like this or try finding some testers first testing the debug artifact?

lm41:
https://github.com/florisboard/florisboard/issues/2363#issuecomment-1974898162

lm41:
I think you can merge it in like this now.

patrickgold:
Perfect doing that now

PR #2370, opened 2024-01-13
fix pinned items getting unpinned when copying the same content again

### LINKED ISSUE #2318: Copying to clipboard unpins pinned items
#### After pinning a word, if you copy the same word sometimes it unpins the pinned entry
In my case I pinned some tracking numbers.
Sometimes it is easier to just copy and paste than to go to history and select pinned item.
However, copying the same tracking number in some cases unpins the previously pinned entry, which is not what I want.
Pinned entries should not be unpinned unless user specifically unpins or deletes entry.

#### Steps to reproduce
1. Pin an item/word
2. After using keyboard for a while, copy the same entry that was recently pinned
3. If you check clipboard history, sometimes that entry gets unpinned
4. Pinned items, should not be removed automatically if you copy them again

#### Environment information
- FlorisBoard Version: 0.3.16 
- Install Source: Github 
- Device: Mix 2s 
- Android: 13

lm41:
Closes #2318


patrickgold (approved):
thanks for the fix, tested it out and everything works, merging in

PR #2371, opened 2024-01-13
fix primary clipboard entry accessible while phone is locked

### LINKED ISSUE #2247: Primary clipboard entry accessible while phone is locked
#### Short description
The primary clipboard entry is still accessible while the phone is locked. Originally reported in Matrix chat.

#### Steps to reproduce
1. Copy something while the phone is unlocked
2. Lock the device
3. While locked, the primary clip can still be pasted into the password field and shown via the eye icon

This affects any paste buttons. The suggestion bar correctly detects the locked state and does not show the primary clip suggestions.

#### Environment information
- FlorisBoard Version: 0.4.0-alpha04 (and latest debug version)
- Install Source: Custom-compiled
- Device: not relevant
- Android: not relevant

(dev notes)
- Use same locking mechanism as for clipboard history

lm41:
Closes #2247


patrickgold (approved):
thanks for the fix, tested out and the paste action is not accessible anymore in the locked state. merging in

PR #2383, opened 2024-01-21
Add Gujarati language Layout

Abcd1234-dot:
This is my first pull request. I have no coding background so make sure there's no syntax error. There are two characters in gujarati language which are made up of combination of other characters but they are still one of the alphabets, so they do not have html code, GBoard still have them ("ક્ષ" and "જ્ઞ" respectively), is there any way to add them too? for now I excluded them.


patrickgold:
Hi there!

Issue #2378, opened 2024-01-14
Crash report screen not working on A14


tsiflimagas:
### Short description

It seems that crash report screen doesn't work on Android 14, possibly because notifications permission needs to be declared. 

### Steps to reproduce

1. Go to FlorisBoard settings>Devtools
2. Click on "Test crash report screen"
3. See error

### FlorisBoard Version

0.4.0-alpha06

### Install Source

GitHub

### Device

Poco F2 pro

### Android

Android 14, CrDroid ROM

Issue #2134, opened 2022-12-19
Smartbar vibrations


KaKi87:
#### Short description
Smartbar actions used to vibrate on press but no longer do from an alpha version to another.

#### Steps to reproduce
1. Install older alpha version
2. Notice smartbar haptic feedback working
3. Install newer alpha version
4. Notice smartbar haptic feedback no longer working

#### Environment information
- FlorisBoard Version: 0.4.0-alpha04 (90)
- Install Source: F-Droid
- Device: Xiaomi Mi 8
- Android: 12 (crDroid 8)

---

Thanks

Issue #2112, opened 2022-11-14
Add option to share image to florisboard


rancidfrog:
Most apps do not provide copy image,
![](https://i.imgur.com/IRcBOcj.png)

Even gallery apps and image editors such as simple gallery, aves, pocketpaint, do not have a copy image.
Only bromite does, in my apps list.

However, all apps have share, so we can extend the share to florisboard which in turn copies image to clipboard.

Currently, share mechanism does not have a copy for images themselves:
![](https://i.imgur.com/anblqdz.png)

![](https://i.imgur.com/6Fn8u7L.png)

It either shows the image or blank, but no option to copy,

patrickgold:
That's a creative idea I must say, really like it :)

I don't know for sure if it is 100% possible but normally this should be absolutely implementable. Will assign it to the list of clipboard issues for now, cannot give any ETA when I (or someone else) will work on this.

alensiljak:
What would be the point of sharing an image to a keyboard app?

There are already several "share to clipboard" apps one can use for the scenario mentioned here. Share is a common operation on mobile devices. Not to mention that clipboard is inherently unsafe (can be monitored by other apps).

rancidfrog:
Fairemail supports pasting images inline.
I use it when I need to add an image inline in an email quickly without any hassle.
Florisboard already supports copying and pasting images.
As many other apps have yet to support copying images to clipboard, sharing to florisboard solves the limitation of other apps. So you can use copying and pasting images as user and keyboard already allows

PR #2388, opened 2024-01-24
Fix #2309, #2134,  #2112, #2378 and some deprecation warnings

### LINKED ISSUE #2309: Recent emoji show up in lock screen
Recent emoji are displayed while device is locked.

# Steps to reproduce
1. Lock device
2. Focus on password field
3. Open emoji menu
4. Recents are shown

# Environment information
- FlorisBoard Version: 0.3.16 (86)
- Install Source: F-Droid
- Device: Pixel 4a 5G
- Android: 14

lm41:
Closes: #2309 
Closes: #2134 
Closes: #2112
Closes: #2378


app/src/main/AndroidManifest.xml
```diff
@@ -19,6 +19,9 @@
     <!-- Permission needed to vibrate if the user has key press vibration enabled -->
     <uses-permission android:name="android.permission.VIBRATE"/>
 
+    <!-- Permission needed to create notifications on devices running android 13+-->
```
patrickgold, line 22:
```suggestion
    <!-- Permission needed to create notifications on devices running Android 13+ -->
```

(formatting nitpick)

app/src/main/kotlin/dev/patrickgold/florisboard/app/FlorisAppActivity.kt
```diff
@@ -96,6 +97,14 @@ class FlorisAppActivity : ComponentActivity() {
...
         }
 
+        //Check if android 13+ is running and the NotificationPermission is not set
+        if (AndroidVersion.ATLEAST_API33_T &&
+            prefs.internal.notificationPermissionState.get() == NotificationPermissionState.NOT_SET
+        ) {
+            // update pref value to show the setup screen again again
+            prefs.internal.isImeSetUp.set(false)
+        }
+
```
patrickgold, line 107: That check does not work or the redirect fails, either way if I disable notifications after setup the settings never complains about that anymore

patrickgold (changes requested):
Thanks a lot for the changes, really appreciate it!

I've reviewed the code and tested out each issue if it is fixed. My feedback:

1) #2309 is fixed, recent emojis aren't showing up anymore when the device is locked. Code for that also looks clean
2) #2134 is fixed but introduced a new crash, see the specific review comment for details
3) #2112 I don't like the UX of the import, can you add at least a toast message with "Added to clipboard history" or "Failed to add because XYZ"? Also the import itself is extremely buggy, e.g. from the OnePlus camera import fails, from ente photos it works flawlessly (same image btw).
4) #2378 is confirmed fixed, however you allow passage in the setup screen even if the user denies the permission, is this intended?

Also the usual formatting nitpicks as single comments :)

app/src/main/kotlin/dev/patrickgold/florisboard/app/FlorisAppActivity.kt
lm41: This is intended behavior. We can no longer request permission after it has been rejected anyway
https://developer.android.com/training/permissions/requesting#handle-denial
If the user decides to denie a permission he has to take the consequences. He can always grant the permission in the android app settings anyway. 
```
At the same time, your app should respect the user's decision to deny a permission. Starting in Android 11 (API level 30), if the user taps Deny for a specific permission more than once during your app's lifetime of installation on a device, the user doesn't see the system permissions dialog if your app requests that permission again. The user's action implies "don't ask again." On previous versions, users saw the system permissions dialog each time your app requested a permission, unless they had previously selected a "don't ask again" checkbox or option.
```

Now to the check. This is primarily there to bring the permission / the new grant notification step back into the setup screen for devices that are android 13+ and have not yet seen it. So for people who have already installed the app in a beta, who have already completed the setup, but have not yet granted the permission because it was not implemented before.

lm41:
I cannot reproduce the buggy behavior of the share to florisboard feature. It works flawlessly on my Pixel (GrapheneOS Android 14).

app/src/main/kotlin/dev/patrickgold/florisboard/app/FlorisAppActivity.kt
patrickgold: Ah I see thanks a lot for the explanation, didn't know that specific detail about permissions. Also now that check makes sense.

patrickgold:
> I cannot reproduce the buggy behavior of the share to florisboard feature. It works flawlessly on my Pixel (GrapheneOS Android 14).

It could also be buggy behavior of the OnePlus camera app, will check if I find other apps to reproduce this on.

patrickgold:
An idea for the copy to clipboard activity UX: instead of auto closing it you could also make the activity transparent and slide a small dialog bottom sheet in, which shows the copied image, a text like "added to FlorisBoard clipboard" (or in an error case error details) and a button to close (additionally to swipe down). This way we can prevent visual artifacts, because if the system is quite busy the activity can cause a short black screen flash.

lm41:
> An idea for the copy to clipboard activity UX: instead of auto closing it you could also make the activity transparent and slide a small dialog bottom sheet in, which shows the copied image

I like this idea. I will try something like that. 
I hadn't even thought about the possibility of a blakcscreen, as I don't think this occurs with my hardware. I really need an older device on which I can test something like this :/

patrickgold:
> I really need an older device on which I can test something like this :/

an emulator with arm architecture on x86 hardware is enough :)

app/src/main/res/values/themes.xml
```diff
@@ -31,4 +31,21 @@
...
+        <item name="android:windowBackground">@android:color/transparent</item>
+        <item name="android:colorBackground">@android:color/transparent</item>
+        <item name="android:windowIsTranslucent">true</item>
+
+        <!--<item name="android:forceDarkAllowed" tools:targetApi="q">false</item>
+        <item name="android:windowIsTranslucent">true</item>
+        <item name="android:windowContentOverlay">@null</item>
+        <item name="android:windowNoTitle">true</item>
+        <item name="android:windowIsFloating">true</item>
+        <item name="android:backgroundDimEnabled">false</item>-->
```
patrickgold, line 48: Those can be removed, was an artifact of me debugging your theme

```suggestion
```

app/src/main/kotlin/dev/patrickgold/florisboard/FlorisCopyToClipboardActivity.kt
```diff
@@ -0,0 +1,245 @@
...
+import dev.patrickgold.florisboard.lib.compose.ProvideLocalizedResources
+import dev.patrickgold.florisboard.lib.compose.stringRes
+import dev.patrickgold.florisboard.lib.snygg.ui.snyggBackground
+import dev.patrickgold.florisboard.lib.snygg.ui.snyggClip
+import dev.patrickgold.florisboard.lib.snygg.ui.solidColor
+import dev.patrickgold.florisboard.lib.snygg.ui.spSize
+import kotlin.math.roundToInt
+
+class FlorisCopyToClipboardActivity : ComponentActivity() {
+
```
patrickgold, line 58:
```suggestion
```

(formatting nitpicking)

patrickgold (changes requested):
Thanks a lot for the changes! Your code looks really good now, just some small formatting nitpickings and then we can merge in the changes!

patrickgold:
Also should I fix the merge conflicts upon merging or do you do it?

patrickgold (approved):
Perfect thanks for fixing, will merge in the changes once the pipeline is finished!

PR #2391, opened 2024-01-29
Update year in Copyright

inson1:


patrickgold (approved):
Thanks for the quick change, merging in!

PR #2432, opened 2024-03-29
Switch to ImageVector icons

lm41:


app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/ComputingEvaluator.kt
```diff
@@ -228,55 +261,55 @@ fun ComputingEvaluator.computeIconResId(data: KeyData): Int? {
...
             }
         }
         KeyCode.TOGGLE_AUTOCORRECT -> {
-            R.drawable.ic_font_download
+            Icons.Default.FontDownload
         }
         KeyCode.KANA_SWITCHER -> {
             if (evaluator.state.isKanaKata) {
-                R.drawable.ic_keyboard_kana_switcher_kata
+                ImageVector.vectorResource(theme = null, resId = R.drawable.ic_keyboard_kana_switcher_kata, res = this.context()?.resources!!)
```
patrickgold, line 290: if we can construct an ImageVector from a drawable res id that easily could we not obsolete jetIcon in jetpref and make a shorthand function?

E.g.
```kotlin
fun Context.vectorResource(@DrawableRes id: Int): ImageVector {
    val theme = null // could be inferred from context receiver probably
    return ImageVector.vectorResource(theme = theme, resId = id, res = this.resources)
}

@Composable
inline fun vectorResource(@DrawableRes id: Int): ImageVector {
    val context = LocalContext.current
    return context.vectorResource(id)
}
```

Then in all preference composables just accept an ImageVector. If you like that idea I would suggest completely removing drawable id support in `jetpref` (just accept an `icon: ImageVector?` param) and adding this utility to the `lib.compose` module.

lm41: We could, but I would include the shorthand function in the jetpref library. (So that if others want to use jetpref they don't always have to implement it themselves)

patrickgold: Ok yeah then lets do it this way

gradle/libs.versions.toml
```diff
@@ -23,7 +24,7 @@ ksp = "1.9.22-1.0.17"
 mannodermaus-android-junit5 = "1.10.0.0"
 mikepenz-aboutlibraries = "10.10.0"
 patrickgold-compose-tooltip = "0.1.0"
-patrickgold-jetpref = "0.1.0-rc01"
+patrickgold-jetpref = "20240328T224800Z-SNAPSHOT"
```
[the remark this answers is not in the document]
lm41: i have now added the review suggestions and am now waiting for a release of jetpref.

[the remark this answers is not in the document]
patrickgold: I made a new [release](https://github.com/patrickgold/jetpref/releases/tag/0.1.0-rc02)

settings.gradle.kts
```diff
@@ -29,7 +29,7 @@ dependencyResolutionManagement {
         google()
         mavenCentral()
         // Uncomment the following if testing snapshots from Maven Central
-        // maven("https://s01.oss.sonatype.org/content/repositories/snapshots")
+        maven("https://s01.oss.sonatype.org/content/repositories/snapshots")
```
patrickgold, line 32: needs to be recommented again
```suggestion
        // maven("https://s01.oss.sonatype.org/content/repositories/snapshots")
```

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/about/AboutScreen.kt
```diff
@@ -72,8 +78,9 @@ fun AboutScreen() = FlorisScreen {
                 modifier = Modifier.padding(top = 16.dp),
             )
         }
+        //TODO: Wait until https://github.com/patrickgold/jetpref/pull/2 is merged to update the preferences
```
patrickgold, line 81:
```suggestion
```

patrickgold (changes requested):
Thanks a lot for the changes! Tested them and everything checks out fine, only 2 small artifacts from testing need to be changed and then it is ready for merging into master.

patrickgold (approved):

PR #2450, opened 2024-04-20
fix rust beta build

lm41:


patrickgold (approved):
Thanks for the quick fix, merging in

Issue #1944, opened 2022-07-08
Suggestion: Add Clipboard History in Backup/Restore Data category


09kkthxbb:
![Screenshot_20220708-170228_Pixel Launcher](https://user-images.githubusercontent.com/70685560/177959047-90e02513-0906-4b94-9c17-810ba4bab7c9.png)
Hi. I've been using your keyboard for several months and I have no issues so far. I like its Backup/Restore feature because it doesn't need to reconfigure everything manually when I want to reinstall or use it to my other devices from scratch.

And I think it would be better if you add Clipboard History as well (text-based clipboard) to be a part of backup/restore checkbox in Advanced settings, like for the Pinned and Other categories if possible. By adding that, we don't need to recopy or readd text everything to clipboard and re-pin what I want in particular order.

Thank you very much for this keyboard development!

patrickgold:
Thanks for your suggestion!

As the clipboard data is stored inside a database, it isn't as easy to backup and restore it. There was already a discussion about this here on GitHub (but I can't find it for some reason), but basically what we concluded there is that we can write a converter to a json file upon backup, and on restore reading this json and creating new database entries.

Will add this to the list of Backup&Restore issues, don't know though when exactly I will implement this, as it requires a bit more work to do this properly.

PR #2458, opened 2024-04-24
Add clipboard history to backup

lm41:
Closes: #1944 

- [x] Working proof of concept.
- [x] Add strings
- [x] Fix SQLiteConstraintException in some cases
- [x] Fix duplicate clipboard items when merging the same archive multiple times (related to the above fix)
- [x] Move file to lib.android and remove lib.compose todo
- [x] Update UI
- [x] Add image media backup
- [x] Add video media backup
- [x] Add image media restore
- [x] Add video media restore
- [x] ~~Investigate/fix strange application status after 'Erase and overwrite current data'~~

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/BackupScreen.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/BackupScreen.kt
@@ -16,6 +16,7 @@
 
 package dev.patrickgold.florisboard.app.settings.advanced
 
+import android.content.ContentUris
 import android.content.Intent
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.result.contract.ActivityResultContracts
@@ -45,6 +46,9 @@ import dev.patrickgold.florisboard.R
 import dev.patrickgold.florisboard.app.LocalNavController
 import dev.patrickgold.florisboard.cacheManager
 import dev.patrickgold.florisboard.clipboardManager
+import dev.patrickgold.florisboard.ime.clipboard.provider.ClipboardFileStorage
+import dev.patrickgold.florisboard.ime.clipboard.provider.ClipboardMediaProvider
+import dev.patrickgold.florisboard.ime.clipboard.provider.ItemType
 import dev.patrickgold.florisboard.lib.android.showLongToast
 import dev.patrickgold.florisboard.lib.android.writeFromFile
 import dev.patrickgold.florisboard.lib.cache.CacheManager
@@ -70,7 +74,9 @@ import kotlinx.serialization.Serializable
 object Backup {
     const val FILE_PROVIDER_AUTHORITY = "${BuildConfig.APPLICATION_ID}.provider.file"
     const val METADATA_JSON_NAME = "backup_metadata.json"
-    const val CLIPBOARD_JSON_NAME = "clipboard.json"
+    const val CLIPBOARD_TEXT_ITEMS_JSON_NAME = "clipboard_text_items.json"
+    const val CLIPBOARD_IMAGES_JSON_NAME = "clipboard_images.json"
+    const val CLIPBOARD_VIDEO_JSON_NAME = "clipboard_video.json"
 
     fun defaultFileName(metadata: Metadata): String {
         return "backup_${metadata.packageName}_${metadata.versionCode}_${metadata.timestamp}.zip"
@@ -92,6 +98,10 @@ object Backup {
 
         fun validateClipboardCheckbox(): Boolean {
             return clipboardTextItems && clipboardImageItems && clipboardVideoItems
+        }
+
+        fun provideClipboardItems(): Boolean {
+            return clipboardTextItems || clipboardImageItems || clipboardVideoItems
         }
 
         fun atLeastOneSelected(): Boolean {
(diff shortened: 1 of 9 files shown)


app/src/main/kotlin/dev/patrickgold/florisboard/lib/util/UriSerializer.kt
patrickgold, line 1: I'd personally opt for this serializer to be placed in lib.compose, as it is a direct "extension" of functionality of a system class

patrickgold (changes requested):
Thanks a lot for the changes!

In general a good addition, however if we allow clipboard backup we should also include media items, else I can already see users complaining that this is missing. In fact the backup of media items and its difficulty was the primary reason why I didn't touch this feature yet.

But other than that the backup of text items is cleanly implemented!

patrickgold:
In terms of the weird app state I'll investigate later on if I can reproduce this and if I find a fix for it.

app/src/main/kotlin/dev/patrickgold/florisboard/ime/clipboard/provider/ClipboardMediaProvider.kt
```diff
@@ -30,6 +30,7 @@ import android.provider.OpenableColumns
 import androidx.exifinterface.media.ExifInterface
 import dev.patrickgold.florisboard.BuildConfig
 import dev.patrickgold.florisboard.lib.devtools.flogError
+import dev.patrickgold.florisboard.lib.io.FsDir
```
patrickgold, line 33: that import is probably an artifact right?

app/src/main/kotlin/dev/patrickgold/florisboard/ime/clipboard/provider/ClipboardFileStorage.kt
```diff
@@ -28,6 +28,9 @@ import dev.patrickgold.florisboard.lib.io.subFile
  * Backend helper object which is used by [ClipboardMediaProvider] to serve content.
  */
 object ClipboardFileStorage {
+
```
patrickgold, line 31:
```suggestion
```

(format nitpick)

app/src/main/kotlin/dev/patrickgold/florisboard/ime/clipboard/provider/ClipboardDatabase.kt
```diff
@@ -304,6 +308,7 @@ abstract class ClipboardHistoryDatabase : RoomDatabase() {
 }
 
 @Entity(tableName = CLIPBOARD_FILES_TABLE)
+@Serializable
```
patrickgold, line 311: I was about to nitpick that the order of annotations vary for ClipboardItem and ClipboardFilesInfo until I did a quick search and noticed there is 0 consistency from my added code either... I'd just suggest that in the same file we have consistent ordering.

app/src/main/kotlin/dev/patrickgold/florisboard/ime/clipboard/ClipboardManager.kt
```diff
@@ -296,6 +296,34 @@ class ClipboardManager(
...
     }
 
+
+    /**
+     * Restore the clipboard history from a [List]
+     *
+     * @param shouldReset if the history should be reset
+     * @param items the [ClipboardItem] list with the new items
+     */
+    fun restoreHistory(shouldReset: Boolean, items: List<ClipboardItem>) {
```
patrickgold, line 306: Flags of functions are usually placed after the main argument, in that case I'd suggest having `items` as the first arg and `shouldReset` as the second

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/BackupScreen.kt
```diff
@@ -72,9 +91,21 @@ object Backup {
         var jetprefDatastore by mutableStateOf(true)
         var imeKeyboard by mutableStateOf(true)
         var imeTheme by mutableStateOf(true)
+        var clipboardTextItems by mutableStateOf(true)
+        var clipboardImageItems by mutableStateOf(true)
+        var clipboardVideoItems by mutableStateOf(true)
+        var clipboardData by mutableStateOf(true)
```
patrickgold, line 97: Here the question arises should clipboard data automatically be marked for backup? Clipboard data is the first section in this screen that backs up sensitive information, so if the user doesn't notice that it is checked and backs up the zip to an unencrypted location they could accidentally leak private information. However I am not sure and open to other opinions here.

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/BackupScreen.kt
```diff
@@ -253,6 +316,53 @@ internal fun BackupFilesSelector(
...
+                } else {
+                    filesSelector.clipboardTextItems = false
+                    filesSelector.clipboardImageItems = false
+                    filesSelector.clipboardVideoItems = false
+                }
+                filesSelector.clipboardData = filesSelector.validateClipboardCheckbox()
+            },
+            checked = filesSelector.clipboardTextItems && filesSelector.clipboardImageItems && filesSelector.clipboardVideoItems,
+            text = stringRes(R.string.backup_and_restore__back_up__files_clipboard_history)
+        )
```
patrickgold, line 335: Technically from UI design perspective this should be a [TriStateCheckbox](https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#TriStateCheckbox(androidx.compose.ui.state.ToggleableState,kotlin.Function0,androidx.compose.ui.Modifier,kotlin.Boolean,androidx.compose.foundation.interaction.MutableInteractionSource,androidx.compose.material.CheckboxColors)), because if only some children are selected the checkbox should reflect that the clipboard is only partially backed up. However I don't mind if we keep it the current way.

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/BackupScreen.kt
```diff
@@ -253,6 +316,53 @@ internal fun BackupFilesSelector(
...
+            text = stringRes(R.string.backup_and_restore__back_up__files_clipboard_history__clipboard_text_items),
+            isSecondaryListItem = true,
+        )
+        CheckboxListItem(
+            onClick = {
+                filesSelector.clipboardImageItems = !filesSelector.clipboardImageItems
+                filesSelector.clipboardData = filesSelector.validateClipboardCheckbox()
+            },
+            checked = filesSelector.clipboardImageItems,
+            text = stringRes(R.string.backup_and_restore__back_up__files_clipboard_history__clipboard_text_items),
```
patrickgold, line 353:
```suggestion
            text = stringRes(R.string.backup_and_restore__back_up__files_clipboard_history__clipboard_image_items),
```

(you used the incorrect string, probably due to copy/paste)

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/BackupScreen.kt
```diff
@@ -253,6 +316,53 @@ internal fun BackupFilesSelector(
...
+            isSecondaryListItem = true,
+        )
+        CheckboxListItem(
+            onClick = {
+                filesSelector.clipboardVideoItems = !filesSelector.clipboardVideoItems
+                filesSelector.clipboardData = filesSelector.validateClipboardCheckbox()
+            },
+            checked = filesSelector.clipboardVideoItems,
+            text = stringRes(R.string.backup_and_restore__back_up__files_clipboard_history__clipboard_text_items),
```
patrickgold, line 362:
```suggestion
            text = stringRes(R.string.backup_and_restore__back_up__files_clipboard_history__clipboard_video_items),
```

(you used the incorrect string, probably due to copy/paste)

app/src/main/res/values/strings.xml
```diff
@@ -502,6 +502,10 @@
     <string name="backup_and_restore__back_up__files_ime_keyboard">Keyboard extensions</string>
     <string name="backup_and_restore__back_up__files_ime_spelling">Spelling extensions / dictionaries</string>
     <string name="backup_and_restore__back_up__files_ime_theme">Theme extensions</string>
+    <string name="backup_and_restore__back_up__files_clipboard_history">Clipboard History</string>
```
patrickgold, line 505:
```suggestion
    <string name="backup_and_restore__back_up__files_clipboard_history">Clipboard history</string>
```

(consistency with other parts of the UI)

app/src/main/res/values/strings.xml
```diff
@@ -502,6 +502,10 @@
...
+    <string name="backup_and_restore__back_up__files_clipboard_history__clipboard_text_items">Text Items</string>
```
patrickgold, line 506:
```suggestion
    <string name="backup_and_restore__back_up__files_clipboard_history__clipboard_text_items">Text items</string>
```

(consistency with other parts of the UI)

patrickgold (changes requested):
Thanks a lot for the changes!

I reviewed your code and tested the functionality, please see the comments on what I found.

A general thing: Should the clipboard history be exportable when it is disabled in the prefs? I am currently torn apart if it should or shouldn't...

In terms of weird app state could you elaborate more on what you mean by that or how to reproduce it? On my quick testing I couldn't really see it.

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/BackupScreen.kt
lm41: You're right, it's more from testing that I didn't always have to check it manually

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/BackupScreen.kt
lm41: Oops

Issue #2032, opened 2022-08-29
Poor switch divider visibility in settings ui


tsiflimagas:
#### Short description
I just noticed the switch divider in the settings and it's a quite nice addition, but at the moment it's a little hard to discern. Well, it could be looking somewhat different on another screen, so you may want to check for yourself. For me, it looks okayish on white and dark themes, though it could still be more distinct, but on AMOLED dark I can't see it almost at all. I'm not sure if it's the color or the width that should be changed. By the way, I do realize it has to look discreet, so I'm proposing just a subtle change, nothing exaggerated. 

#### Steps to reproduce
1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

#### Environment information
- FlorisBoard Version: 0.4.0-alpha03
- Install Source: GitHub
- Device: Xiaomi Redmi note 8
- Android: 12.1, custom

tsiflimagas:
I just wondered whether I should have opened this issue in the Jetpref repo🤔

patrickgold:
Thanks for your bug report!

I have noticed this too that the divider is not that much noticeable in some conditions, don't really know why this happens because in general I just assign the color from the MaterialTheme as per Material specs. Either the FlorisBoard app theme is misconfigured or something else is wrong.

> I just wondered whether I should have opened this issue in the Jetpref repothinking

No this repo is okay, as the JetPref lib doesn't actually set the color but derives it from the app's MaterialTheme, in this case the FlorisAppTheme, and that's where the issue originates from I think.

lm41:
Fixed with https://github.com/patrickgold/jetpref/pull/8

PR #2467, opened 2024-05-04
Material3 migration

lm41:
Fix: #2032

- [x] Investigate wrong colors on some cards (info card in 'Languages & Layouts' and in 'Typing')
- [x] Change background
- [x] Release Jetpref 0.2.0-beta01

### For the testers:
This PR addresses the Settings UI and not the Keyboard. 
Please note the following points in particular:
1. Strange UI bugs that weren't there before
If you notice any unexpected behavior or something that doesn't look right, please take the time to document it. Any unusual glitch could be an indication of a problem that needs to be fixed.
2. Inconsistent theming
We want to make sure that the theme is consistent and appealing. If you come across any inconsistencies or inconsistent designs, please let us know so we can take care of it.

Please note that comments such as "Material 2 was better than Material 3" are not particularly helpful. We are focusing on moving from Material 2 to Material3/Material You and your constructive feedback will help us to do this

Link to the debug build: https://github.com/florisboard/florisboard/actions/runs/8983708234/artifacts/1479764046


app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/about/ThirdPartyLicensesScreen.kt
```diff
@@ -39,6 +41,13 @@ fun ThirdPartyLicensesScreen() = FlorisScreen {
             modifier = Modifier
                 .fillMaxSize()
                 .florisScrollbar(lazyListState, isVertical = true),
+            colors = LibraryDefaults.libraryColors(
+                backgroundColor = MaterialTheme.colorScheme.background,
+                badgeBackgroundColor = MaterialTheme.colorScheme.primaryContainer,
+                badgeContentColor = MaterialTheme.colorScheme.onPrimaryContainer,
+                contentColor = MaterialTheme.colorScheme.onBackground,
+                dialogConfirmButtonColor = MaterialTheme.colorScheme.primary
```
patrickgold, line 49:
```suggestion
                dialogConfirmButtonColor = MaterialTheme.colorScheme.primary,
```

(nitpick)

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/AdvancedScreen.kt
```diff
@@ -77,6 +78,14 @@ fun AdvancedScreen() = FlorisScreen {
                 )
             },
         )
+        SwitchPreference(
+            pref = prefs.advanced.useMaterialYou,
+            icon = Icons.Default.FormatPaint,
+            title = "Use Material You",
+            visibleIf = {
+                AndroidVersion.ATLEAST_API31_S
+            }
```
patrickgold, line 87:
```suggestion
            },
```

(nitpick)

app/src/main/kotlin/dev/patrickgold/florisboard/lib/compose/FlorisCards.kt
```diff
@@ -70,24 +67,30 @@ object BoxDefaults {
...
         onClick = onClick ?: { },
         enabled = onClick != null,
         modifier = modifier.fillMaxWidth(),
-        backgroundColor = backgroundColor,
-        contentColor = contentColor,
+        colors = CardDefaults.cardColors(
+            contentColor = contentColor,
+            containerColor = backgroundColor,
+            disabledContainerColor = backgroundColor,
+            disabledContentColor = contentColor
```
patrickgold, line 90:
```suggestion
            disabledContentColor = contentColor,
```

(nitpick)

app/src/main/kotlin/dev/patrickgold/florisboard/lib/compose/FlorisChip.kt
```diff
@@ -16,44 +16,83 @@
...
+                            .size(16.dp),
+                        imageVector = trailingIcon,
+                        contentDescription = null,
+                    )
+                }
+            }
+        }
+
+
+    )
```
patrickgold, line 93:
```suggestion
        }
    )
```

gradle/libs.versions.toml
```diff
@@ -24,7 +24,7 @@ ksp = "1.9.22-1.0.17"
 mannodermaus-android-junit5 = "1.10.0.0"
 mikepenz-aboutlibraries = "10.10.0"
 patrickgold-compose-tooltip = "0.1.0"
-patrickgold-jetpref = "0.1.0-rc02"
+patrickgold-jetpref = "20240504T230200Z-SNAPSHOT"
```
patrickgold, line 27: We need a proper jetpref release before merging in

settings.gradle.kts
```diff
@@ -29,7 +29,7 @@ dependencyResolutionManagement {
         google()
         mavenCentral()
         // Uncomment the following if testing snapshots from Maven Central
-        // maven("https://s01.oss.sonatype.org/content/repositories/snapshots")
+        maven("https://s01.oss.sonatype.org/content/repositories/snapshots")
```
patrickgold, line 32: before merge comment in again

app/src/main/kotlin/dev/patrickgold/florisboard/lib/compose/FlorisScreen.kt
patrickgold, line 1: I have identified the issue regarding the scrollbar calculating too much, it was applied to early, which was always an error but somehow still worked in the old UI...

Please replace the Scaffold code in Render() with this:
```kotlin
        Scaffold(
            topBar = { FlorisAppBar(title, navigationIcon.takeIf { navigationIconVisible }, actions) },
            bottomBar = bottomBar,
            floatingActionButton = fab,
        ) { innerPadding ->
            val scrollModifier = if (scrollable) {
                Modifier.florisVerticalScroll()
            } else {
                Modifier
            }
            PreferenceLayout(
                florisPreferenceModel(),
                modifier = Modifier
                    .padding(innerPadding)
                    .fillMaxWidth()
                    .then(scrollModifier),
                iconSpaceReserved = iconSpaceReserved,
                content = content,
            )
        }
```

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/advanced/AdvancedScreen.kt
```diff
@@ -77,6 +78,14 @@ fun AdvancedScreen() = FlorisScreen {
                 )
             },
         )
+        SwitchPreference(
+            pref = prefs.advanced.useMaterialYou,
+            icon = Icons.Default.FormatPaint,
+            title = "Use Material You",
```
patrickgold, line 84: Please add a string resource for this to it is translatable

patrickgold (changes requested):
Thanks a lot for the changes! Generally the new UI looks pretty solid, we may still need some minor color scheme adjustments but apart from that everything looks good.

I've reviewed the code changes and have some requests and nitpicks.

patrickgold:
I've just noticed that issue #2032 is still open, could we fix this with this PR as well if we are already completely reworking the color scheme?

lm41:
> I've just noticed that issue #2032 is still open, could we fix this with this PR as well if we are already completely reworking the color scheme?

Yes, but this has to be done in the JetPref repo. I'll open a PR

tsiflimagas:
I've seen that some apps also use `com.mikepenz:aboutlibraries-compose-m3`. It seems that the only change it brings in this case is to make the dependency dialog card and the ok button pill shaped.
Also, it seems that in the theme edit section, the edit and delete buttons are still rectangle shaped. Is this because of the rectangle outline they're in or should they also be pill shaped? 
Screenshot_20240506-135015_FlorisBoard Debug (1) (1)
Another thing which seems to me could be different is the color and height of the topbar when scrolling down in a section. Just empirically, it seems to me that the text should be below the back button when at the top of the page, and while scrolling, the topbar collapses to a smaller height, the text moves next to the back button, and the topbar's tone changes to a slightly lighter one.
I guess you material3 conventions very well already, so you can assess this. Thank you for your work!

lm41:
Thank you for your suggestions :) 
However, the TopAppBar remains a "normal" TopAppBar, as otherwise there is too much empty space for my taste. Only the colors I have now adapted to spec.

tsiflimagas:
Thank you for implementing the indicated points! 

> However, the TopAppBar remains a "normal" TopAppBar, as otherwise there is too much empty space for my taste

Fair enough, that's reasonable 😀

There are a couple more things that came to my mind:

Do you think those backgrounded "pressed" "focus" "disabled" selectors in theme editor could be made into pill shaped, or would it be misleading resembling pressable buttons? I mean, I think I've seen in other apps such non functional snippets of text being in pill shaped background (like tags maybe), but I don't know if you'd agree with such a design option.
Screenshot_20240506-232419_FlorisBoard Debug (1)


Also, the splashscreen is currently not material You themed. Could this be also taken care of here? 

And the navbar, could it also be coloured? Now it's black. 

Lastly, do you think it would be a good idea to make material You default for A≥12? Many apps which have fully implemented the design, have set it as default, and I think it's reasonable. On a system which entirely follows this design, why not look uniform right away?

lm41:
> Do you think those backgrounded "pressed" "focus" "disabled" selectors in theme editor could be made into pill shaped, or would it be misleading resembling pressable buttons?

I have now given the text background some rounder corners. 
> And the navbar, could it also be coloured? Now it's black.

Navbar and status bar now have the same color.
> Lastly, do you think it would be a good idea to make material You default for A≥12? Many apps which have fully implemented the design, have set it as default, and I think it's reasonable. On a system which entirely follows this design, why not look uniform right away?

Good point. I would like to have more opinions on this.
> Edit: also the crash report screen hasn't been converted to m3, but you probably won't think it'a worthwhile adapting it?

The crash report screen has not been converted to JC. It still uses the old layout xml files. I will have a look at this.

patrickgold (approved):

Issue #1338, opened 2021-10-23
More control to "adapt colors to a app" theming


Pun-e:
Adapting colors automatically is awesome, but the lack of control is a bit underwhelming.

I enjoy the default color scheme immensely, with its small pops of color, but when I turn on adaptive colors, the new colors completely ignore my current theme.

I propose that in the theme maker, that in the ui for selecting a color swatch underneath the transparency slider is a check box simply labeled "adaptive"

Ticking this box means that if you subsequently enable adaptive colors that the program will know which colors to change

Obviously there's bigger things to tackle, but hopefully this is something that can be squeezed in at some stage

Edit: i've just discovered the advanced attributes for theming, so a custom attribute to access the adaptive color palette could also work - but will make setting the default color a bit confusing

klaurence:
Should this issue be closed? `Adapt colors to app` isn't a feature of Florisboard anymore but this can be the issue for implementing specific "adaptive" options in the theme maker.

patrickgold:
This feature is not supported by FlorisBoard anymore, thus I am closing this issue as not-planned.

Issue #1313, opened 2021-10-10
Adapt colours to app not working after restart


ghost:
#### Short description
Having adapt colours to app enabled and then restarting your device makes the keyboard not adapt to the app until you change to another keyboard then back

#### Steps to reproduce
1. Go to theme settings
2. Enable adapt colours to app
3. Restart the app
4. Open the keyboard on any app
5. See the keyboard using the deafult theme

#### Environment information
- FlorisBoard Version: 0.3.14-beta03
- Install Source: F-Droid
- Device: Redmi Note 9 Pro
- Android: 11, MIUI

patrickgold:
This feature is not supported by FlorisBoard anymore, thus I am closing this issue as not-planned.

Issue #1216, opened 2021-08-31
Add option to specify which theme attributes are affected by adaptive themes


patrickgold:
### Discussed in https://github.com/florisboard/florisboard/discussions/1211

<div type='discussions-op-text'>

<sup>Originally posted by **imDusht** August 30, 2021</sup>
Hi Dev, right now adapt color to app theme option changes keyboard : background and key : background color, i was wondering if it could be possible to implement that we can choose which parts of keyboard ui adapts to app color, like implement an option which lists keyboard : background, key : foreground (currently it doesn't changes forever white), key : background parts and we can choose one or multiple options between them.</div>

imDusht:
I'm excited for this one, thank you.

AKDub:
Was going to suggest this for just the "Enter" key as a nice subtle theme change. Good idea having it more customisable.

patrickgold:
This feature is not supported by FlorisBoard anymore, thus I am closing this issue as not-planned.

Issue #853, opened 2021-05-11
Color Adaption UI "bug"


FrameXX:
Hi. There is an option inside FlorisBoard that makes adapt colors to app that is in foreground. This option works, however there's one small problem.
As background colors changes on every opened app, letters and other vectors switches between white and black to be visible on background (if background is darker letters are white if...).
But this is not a thing with vectors on emoji screen. No matter what, these vectors stays black (with light theme) all the time and are nearly invisible on some dark backgrounds as keyboard adapts to the app colors. (screenshot attached).

EDIT: This also applyes for clipboard history section.

#### Environment information
- FlorisBoard Version: 0.3.12
- Install Source: GitHub

![Screenshot_20210511-133020~2](https://user-images.githubusercontent.com/56247804/117810907-3f16de80-b260-11eb-80dd-95236092254d.png)

FrameXX:
I am making issue on github second time and I still don't know how to add labels. It needs to be hidden somewhere.

Glitchy-Tozier:
The labels get added automatically depemding on what template you choose  ;)

For example, you (correctly) chose the bug-template and thus (bug) was added. If you had chosen the question-template, the label (question) would've been added.

![Screenshot_20210511-143144_GitHub.jpg](https://user-images.githubusercontent.com/59611881/117815522-dcc0dc80-b265-11eb-9148-eb90ae240c07.jpg)

TheLoving:
I also have the same issue with color adapted theme.

patrickgold:
This feature is not supported by FlorisBoard anymore, thus I am closing this issue as not-planned.

Glitchy-Tozier:
@patrickgold That's unfortunate. Are there any plans to bring it back at some point of time?

patrickgold:
No there are not atm because its a very difficult feature to get right, apps use all sorts of coloring schemes nowadays, we have Material 1, 2 and 3, then we have apps that are written in Jetpack Compose and thus often times leave the xml colors on the default colors, so yeah

PR #2473, opened 2024-05-13
Add addons support

### LINKED ISSUE #2468: Add support for official Addons Store
### Feature idea

This is a direct preparation for the upcoming official Addons Store, which in the future will be the main hub for official and community created themes, layouts, language packs, etc.

As FlorisBoard is a privacy-focused keyboard and we thus cannot add the NETWROK permission to the app, we have to get creative on how to easily install stuff from the Addons Store. Thus the idea is to display and manage the Addons Store as a website inside the user's browser, and when the user wants to install a new addon, they can simply download the extension file and then open it, which will trigger FlorisBoard's file handler defined in the Manifest and redirect this to the extension import screen.

- [x] Add file import handler for `.flex` files
- [x] Add share import handler for `.flex` files
- [x] Add user-readable error messages for common errors
- [x] Add measures to protect the integrity of the app
  - [x] Prevent ZIP bomb attacks
  - [x] Prevent large file size attacks
  - [x] Prevent file/path name max length attacks
  - [x] Prevent relative parent path expansion overwrite attacks (e.g. `$zip/../../files/prefs.json`)
  - [x] ... more? ...
- [x] Add support for update status detection 
- [x] Add UI for managing installed addons easily within the app UI

Important is that the Addons Store never has direct control, it can just suggest to install/update an extension, the final decision is done by the user from within the import handler app UI. Removal of addons MUST be done from within the app UI.

Initial goal is that theme extensions are supported, later on also keyboard extensions. With v0.5 support for language packs will be added.

In parallel to this efforts the Addons Store itself will be publicly beta-tested by @4H1R, who kindly offered his expertise writing the Addons Store for this project too. This issue is intended to stay open for a few PRs and beta releases.

Any feedback/idea proposals for this are highly appreciated, just comment down below!

patrickgold:
Closes #2468

Please only give implementation feedback in this comment thread, for general feedback regarding the addons integration as planned comment in the related issue!

---

Due to the theme screen rework in conjunction with addons it has been decided that the adapt to app theming option will be entirely removed (which wasn't working anyways for like the last 2 stable versions). There are no plans for re-adding this feature again anytime soon. The following issues are closed as not-planned: #274 #399 #853 #1216 #1313 #1338


TPS:
> Due to the theme screen rework in conjunction with addons it has been decided that the adapt to app theming option will be entirely removed (which wasn't working anyways for like the last 2 stable versions).

Possible to do via addon?

patrickgold:
> Possible to do via addon?

No, due to the fact that this feature requires reading the xml colors of the target app and those are a mess nowadays due to Material 1 2 and 3 all being out in the wild, and together with apps using Jetpack Compose those colors often times do not match at all what the apps colors are, thus I removed it.

app/build.gradle.kts
```diff
@@ -64,6 +64,8 @@ android {
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
         buildConfigField("String", "BUILD_COMMIT_HASH", "\"${getGitCommitHash()}\"")
+        buildConfigField("String", "API_VERSION", "\"v~draft2\"")
+        buildConfigField("String", "STORE_URL", "\"fladdonstest.patrickgold.dev\"")
```
patrickgold, line 68: we could give the build config fields a common prefix, e.g. FLADDONS_***

lm41: Good Idea. I will add that.

lm41 (approved):
changes look good

app/src/main/kotlin/dev/patrickgold/florisboard/app/ext/ExtensionHomeScreen.kt
```diff
@@ -0,0 +1,101 @@
...
+            )
+            Row(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(horizontal = 6.dp),
+            ) {
+                FlorisTextButton(
+                    onClick = {
+                        //context.launchUrl("https://beta.addons.florisboard.org")
+                        context.launchUrl("https://fladdonstest.patrickgold.dev")
```
patrickgold, line 59: don't we have a build config variable for this url? instead of hardcoding it

app/build.gradle.kts
```diff
@@ -64,6 +64,8 @@ android {
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
         buildConfigField("String", "BUILD_COMMIT_HASH", "\"${getGitCommitHash()}\"")
+        buildConfigField("String", "FLADDONS_API_VERSION", "\"v~draft2\"")
+        buildConfigField("String", "FLADDONS_STORE_URL", "\"fladdonstest.patrickgold.dev\"")
```
patrickgold, line 68: changing later before release or now to correct values?

lm41: I would say before release when the update mechanism is ready.

patrickgold (commented):
looks good to me now

PR #2496, opened 2024-06-20
remove deprecated Swipeable API and switch to M3 ModalBottomSheet

lm41:
This PR removes the use of the deprecated Swipeable API.


patrickgold (approved):
changes look good, merging in

PR #2501, opened 2024-07-08
Add possible detection for "All keys invisible" bug

patrickgold:
Related to #2362

This PR tries to add a detection for the "All keys invisible" bug and show a warning box.


lm41 (approved):
lgtm

PR #2508, opened 2024-07-15
Add deep link support for UI

patrickgold:
Add the following scheme: `ui://florisboard/<route>` to allow linking into the FlorisBoard UI


lm41 (approved):
LGTM :D

Issue #2526, opened 2024-08-13
Consider switching to sliding screen nav animation instead of default


soupslurpr:
### Feature idea

Hi, I see that florisboard recently switched to Material 3 design. I want to suggest a switch to a sliding + fade animation for navigating between screens instead of the just the fade animation. The fade animation can lead to misclicks as the UI is still active while it is exiting to another screen. Even the Android/AOSP settings app uses a sliding + fade animation from what I can tell.

I recommend taking a look at my implementation in the GrapheneOS Info app which handles it all automatically: https://github.com/GrapheneOS/Info/blob/main/app/src/main/kotlin/app/grapheneos/info/InfoApp.kt

PR #2527, opened 2024-08-13
Switch to sliding navigation animation

lm41:
Closes: #2526 
This PR replaces the animation that is played during settings navigation. Now it is the same as in the android settings.
In addition, predictive back was prepared with the `android:enableOnBackInvokedCallback="true"` manifest flag.


patrickgold (approved):
lgtm

PR #2535, opened 2024-08-25
fix addon store link pointing to incorrect url

Book-reader:
Closes: #2537


lm41 (approved):
Thank you for your fix. That was definitely a mistake.

patrickgold (approved):
thanks a lot for the quick fix!

PR #2561, opened 2024-09-14
Localize debug log screen and add option for preformatted exporting

lm41:
This PR adds a button that copies the debuglog already preformatted for github to the clipboard.
In addition the stringres for the ExportDebugLogScreen were added.


patrickgold (approved):
lgtm

PR #2565, opened 2024-09-18
Rework version name handling in Gradle

patrickgold:
The version suffix is now encoded in `versionName` instead of being in a separate gradle.properties value to ease handling in bash scripts


lm41 (approved):
:+1:

PR #2567, opened 2024-09-18
Add hint to find_program for Rust discovery

patrickgold:
Fix needed for F-Droid build recipe to properly work as somehow the PATH env is not correctly set up in that process


lm41 (approved):
:+1:

PR #2591, opened 2024-09-22
Fix utility key action not correctly applied

### LINKED ISSUE #2587: Utility key action setting applied belatedly
### Short description

When changing the Utility key action setting, the change doesn't happen immediately, it seems a change in the keyboard layout (switch to symbols, caps lock) is needed to trigger the change. It's not only visual (the icon) that stays unaltered, but also the action. This also leads to the problem described in #2586.

### Steps to reproduce

1. Go to "Keyboard" settings
2. Change the "Utility key action" setting
3. Open the keyboard to see if the option has been applied

### FlorisBoard Version

0.4.0

### Install Source

GitHub

### Device

Poco F2 pro

### Android

14, crDroid

patrickgold:
Closes #2586
Closes #2587

This PR addresses both the issue that the utility key action pref was not observed and that subtype list changes did not properly propagate to the evaluator, causing the evaluate**() methods to be always 1 behind.


tsiflimagas:
It does fix the issue. Thanks for the immediate fix!

patrickgold:
Thanks for testing the debug build! Merging in, will be part of the 0.4.1 patch release.

PR #2592, opened 2024-09-23
Update quick action reordering based on row

### LINKED ISSUE #2556: Inability to Append any Actions  to Singular Entry Rows
Dragging any actions to an almost empty row will either fail and instantly snap back or will replace the only element present rather than being appended next to it.

KevinVan9:
Closes #2556 

Quick action item to swap with now includes the last item in the row when in the empty space in the row. To do this, items' y-offset bounds that contain the given offset are marked as being "close" and returned if no items' bounds explicitly contain the given offset


patrickgold (approved):
Thanks a lot for the PR, this fixes the issue with the drag markers. Merging in, will be part of the 0.4.1 patch release.

PR #2611, opened 2024-09-29
Major rework of emoji history

### LINKED ISSUE #1975: Allow switching between most recent and most frequent emoji
The recent emoji panel provides quick access to the emoji that were recently used, and if the number of emoji shown is not too small, it shows all frequently used emoji, preventing me from having to dig through the other emoji panels.

However, with each emoji that I use, its sorting order gets changed around again, making it hard to find a specific emoji on that panel.

Switching the panel's sorting order from "most recent" to "most frequent" would dramatically reduce that problem, since it would result in a much more static order.

### LINKED ISSUE #1990: Allow disabling emoji history
Emoji history was added in #268 which is great for those who use it: but now there's no way to not have it! :smiley: 

I suggest changing the "emoji history max size" setting so that the low end of the spectrum is 0, which would disable the recently used pane and make the smilies pane the default. The "unlimited" setting could be moved to right after the current numerical max (120).

### LINKED ISSUE #2129: One emoji missing from recents emojis
If you set the history limit to, say, 36 and pressed bunch of random emojis it'll fill all the slots correctly. But if you exit from the emoji tab and select an emoji from the history, one of them will disappear the next time you reenter, resulting in a total of 35 emojis being stored and not 36.

### LINKED ISSUE #2273: Pinning Emojis
Ability to pin emojis in the frequently used emoji area would be a welcome feature.

patrickgold:
This PR's main goal is to fully rework the internal behavior of the emoji history and to add a few new fetaures. List of changes:

- Add ability to enable/disable emoji history (closes #1990)
- Add ability to pin emojis in emoji history (closes #2273)
- Add ability to manually reorder emojis in history (closes #1975)
- Fix emoji history update bug if items is max size (closes #2129)
- Rewrite emoji history logic and data class

#### Screenshot of the new emoji history settings:

image

#### Screenshot of the updated emoji history:

image

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/app/AppPrefs.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/app/AppPrefs.kt
@@ -31,7 +31,7 @@ import dev.patrickgold.florisboard.ime.keyboard.IncognitoMode
 import dev.patrickgold.florisboard.ime.keyboard.SpaceBarMode
 import dev.patrickgold.florisboard.ime.landscapeinput.LandscapeInputUiMode
 import dev.patrickgold.florisboard.ime.media.emoji.EmojiHairStyle
-import dev.patrickgold.florisboard.ime.media.emoji.EmojiRecentlyUsedHelper
+import dev.patrickgold.florisboard.ime.media.emoji.EmojiHistory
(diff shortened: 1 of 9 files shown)


lm41 (approved):
:+1:

PR #2617, opened 2024-10-03
Fix emojis with differing meta data not being seen as identical

patrickgold:
This PR fixes a bug introduced in PR #2611, where equal emoji values but with differing emoji meta data (name, keywords) were not seen as the same.

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/media/emoji/Emoji.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/media/emoji/Emoji.kt
@@ -47,7 +47,7 @@ enum class EmojiHairStyle(val id: Int) {
     BALD(0x1F9B3);
 }
 
-data class Emoji(val value: String, val name: String, val keywords: List<String>) : KeyData {
+class Emoji(val value: String, val name: String, val keywords: List<String>) : KeyData {
     override val type = KeyType.CHARACTER
     override val code = KeyCode.UNSPECIFIED
     override val label = value
@@ -77,6 +77,14 @@ data class Emoji(val value: String, val name: String, val keywords: List<String>
 
     override fun toString(): String {
         return "Emoji { value=$value, name=$name, keywords=$keywords }"
+    }
+
+    override fun hashCode(): Int {
+        return value.hashCode()
+    }
+
+    override fun equals(other: Any?): Boolean {
+        return other is Emoji && value == other.value
     }
 
     object ValueOnlySerializer : KSerializer<Emoji> {


itsmartashub:
**It looks like the issue is fixed now!🎉** Just in case, I’ll just add a additional explanation why I cant be 100% sure.

> [!NOTE]
> Since the exact trigger is unknown to me, testing has relied on a backup with duplicated emojis. The bug appears to be resolved in this debug version, but the testing backup will be attached just in case.

The initial trigger for the bug isnt clear to me tho, which complicates testing. However, once an emoji is duplicated, reproducing the duplication for emojis already populated the `Recents`  become easier. It seems that app prevents duplicates when **reusing** an emoji from `Recents`, but not when selecting it again from the full emoji panel or suggestions. Idk if something was messed up with emoji ids, or indexing in previous debug version.
This might also relate to the `Update strategy`, such as adding, reordering, deleting, and readding emojis combinating all the possible scenarios with different sorting strategies. [see the screenshot below where 🪤 emoji is prepend and append]

This is probably wrong, but its like the bug might happen more often when the keyboard hasnt been used for a while.

An attempt was made to recreate the bug on a different device using the previous `rework/emoji-history` version with default settings. After testing the new options and switching between apps (exiting Florisboard, using another app, and returning), the bug was triggered again. I hope that the attached log file might provide more useful insights.


#### Recording: #2617 debug [build](https://github.com/florisboard/florisboard/actions/runs/11170068821): bug is not reproducible
https://github.com/user-attachments/assets/0d517a05-3647-4706-8807-bf67ed475be0


#### Recording: #2611 debug [build](https://github.com/florisboard/florisboard/actions/runs/11092700267): bug is reproducible 
https://github.com/user-attachments/assets/4a9deee7-6a22-4a9c-b436-74f81cc9c926

#### Log file and screenshot: Emoji duplication triggered for the first time with default settings:
- [emoji-duplication-log-prev-debug.txt](https://github.com/user-attachments/files/17256526/emoji-duplication-log-prev-debug.txt)

400x500


#### Testing backup with emoji duplicated:
- [emoji-rework-duplication-bug--backup_dev.patrickgold.florisboard.debug_97_1728031978758.zip](https://github.com/user-attachments/files/17256545/emoji-rework-duplication-bug--backup_dev.patrickgold.florisboard.debug_97_1728031978758.zip)

patrickgold:
Thanks for the extensive testing!

The trigger for the duplicate emojis (as far as I could identify) was that the same emoji existed with different metadata attached. Basically each emoji in FlorisBoard exists (#NumSubtypes + 1) times, so the emoji suggestions provider can find the emoji by name/keywords. This however caused the equals check to fail because it checked ALL OF value, name and keywords, thus you have been able to easily duplicate emojis with the emoji suggestion provider.

Now with this PR's fixes only the value is considered (so the actual hex code) and meta data is ignored, thus duplicates shouldn't be possible anymore. Except when exporting the settings, manually adding duplicates to the emoji history and importing this, but tbh i couldn't care less if someone does this.

The update strategy itself did not affect the incorrect behavior, as regardless of the strategy all emojis are searched through.

patrickgold:
> If I got it right, there is the main version (+1) and all of its subtypes like diff keywords, names, etc

Basically yes

> 1. Functionality - Clear all btn

That's a neat idea, can you make an issue out of this so I can track this for the new emoji panel? In the old one I won't add it though.

> 2. UI: Can subheadings like Recents and Pinned be tagreted in the theme editor?

No, that's an oversight from me. I will make some adjustments to it takes one of the existing colors, in the emoji rework those headings will get a separate style target

> 3. UX: I assume there are reasons why reordering emojis wasn't done with a drag&drop like reordering actions tiles but rather with arrows.

Yes because the existing code is extremely fragile and adding the popup as it exists now was already a challenge and is a lot of boilerplate code, drag and drop makes this even worse. I can try adding adrag-and-drop in the new emoji history but can't promise this as the emojis live in a lazy list.

itsmartashub:
> can you make an issue out of this so I can track this for the new emoji panel? In the old one I won't add it though.

Sure, thank you for even considering it.


> No, that's an oversight from me. I will make some adjustments to it takes one of the existing colors, in the emoji rework those headings will get a separate style target

Checked out the #2620. It now has the inherited fg color from the existing elements (to be precise, from the `emoji-key` or from the `emoji-key[code=-211|-7]` when/if its defined, if I'm not wrong) 🙏


> Yes because the existing code is extremely fragile and adding the popup as it exists now was already a challenge and is a lot of boilerplate code, drag and drop makes this even worse. I can try adding adrag-and-drop in the new emoji history but can't promise this as the emojis live in a lazy list.

Well, if I understand some of the comments on other posts correctly, there are plans to rewrite the emoji layout even further. So maybe this could be considered for another time when it's clearer how the new layout/panel will work in practice. It's possible that the drag-n-drop logic might not fit into that code, or might not even be necessary/useful.

PR #2631, opened 2024-10-13
Clipboard History enhancements

### LINKED ISSUE #2510: Hide contents of sensitive clip data
### Feature idea

If you copy sensitive information such as passwords and verification codes, display these sensitive information as ***.

### LINKED ISSUE #2615: Backspace in clipboard
### Feature idea

IMG_20241002_035403

I'm referring to that little thing at the bottom right (SS from within Samsung keyboard)

lm41:
Closes #2510 
Closes #2615

### DIFF
--- a/app/schemas/dev.patrickgold.florisboard.ime.clipboard.provider.ClipboardHistoryDatabase/3.json
+++ b/app/schemas/dev.patrickgold.florisboard.ime.clipboard.provider.ClipboardHistoryDatabase/3.json
@@ -1,0 +1,92 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 3,
+    "identityHash": "282a1b421e498fd0e21c055b6a4315e0",
+    "entities": [
+      {
+        "tableName": "clipboard_history",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`_id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `type` INTEGER NOT NULL, `text` TEXT, `uri` TEXT, `creationTimestampMs` INTEGER NOT NULL, `isPinned` INTEGER NOT NULL, `mimeTypes` TEXT NOT NULL, `isSensitive` INTEGER NOT NULL, `isRemoteDevice` INTEGER NOT NULL)",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "_id",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "type",
+            "columnName": "type",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "text",
+            "columnName": "text",
+            "affinity": "TEXT",
+            "notNull": false
+          },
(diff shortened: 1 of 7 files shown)


patrickgold (changes requested):
Thanks a lot for the changes! In general the handling seems fine, however I've identified two issues:
1) If the clipboard entry gets suggested and is sensitive it is displayed as clear text, all instances of the clip data text needs to be obfuscated. I suggest instead of doing this at UI level, add a `ClipboardItem.displayText()` method that does the obfuscation for you and update all manual text usages once, so we future-proof this. We can also move the ****** string to Kotlin, IMO no real need for this to be in the strings_dont_translate.xml
2) I am not a particular fan of the solution for the backspace button, as it takes away the space of a whole row for just one button. I'd suggest either something like a "mini-fab" or the placement of it in the header row.

patrickgold (changes requested):
Thanks again for the changes! The backspace button is now in a good position.

For backspace, it isn't possible to hold down the backspace key. Have a look at this code snippet from MediaInputLayout, which has a convenient KeyboardLikeButton widget: https://github.com/florisboard/florisboard/blob/d6f724e518f6933eb16139a85dde0eec1926b1ae/app/src/main/kotlin/dev/patrickgold/florisboard/ime/media/MediaInputLayout.kt#L111-L153

Other than that the changes look good to me!

patrickgold (approved):
lgtm now, merging in!

itsmartashub:
Thank you very much for implementig this feature.

I hope you wont mind me to ask, but is this expected behavior when copying a sensitive data? Specifically, this was triggered by VirusTotal API key copied from BitWarden, and I would say it has to do with numeric characters. When there are several consecutive numbers, followed by letters/chars, and then more numbers, depending on how many numeric sequences there are, you'll get that many obfuscated suggestions + 1 (the main full pass)

When we click the first obf. suggestion, the full pass/api etc will be pasted. If we click the second sugg, it pastes the first number sequence after the first letter (or group of letters). If we click the third sugg, it pastes only the numbers from the next sequence (the second one), and so on. 

I hope this makes sense. Also, it seems the number of digits in the sequence matters for this to be triggered, such as combinations. For example, the minimum number of digits in the first sequence is 5? If needed, I can record a video

Example:

`r00000b11111n222222k3333`

Copying this would trigger 4 suggestions instead one:

1st: `r00000b11111n222222k3333`
2nd: `00000`
3rd: `11111`
4th: `222222`

`3333` would not be suggested because it doesnt have min 5 digits?

lm41:
I've identified the issue.

The numbers are recognized as “phone numbers” and are therefore suggested.
There are two potential fixes for that.
1. Disable this type of suggestion for sensitive clip data:
image


2. Flag these “phone numbers” (or email/urls) as non-sensitive and display them as such, which would result in potentially sensitive data being leaked visibly from the history:
image

I think 1 (not showing these suggestions) is the right way to go, but I am open to feedback.

patrickgold:
I'd say smart detection of what a sensistive clip is should be disabled (just always assume normal clip), not for privacy reasons because its just a regex check but to not crowd the clip suggestions and prevent the issue Marta uncovered

PR #2637, opened 2024-10-16
Disable compose strong skipping mode

### LINKED ISSUE #2629: Long press popup does not appear often
### Short description

Up to the current release I never had an issue regarding long pressing a key and then choosing from that list. I use german umlauts and special chars a lot but since v0.4.1 sometimes simply nothing happens when long pressing a key. It immediately restarts to work when I close and reopen the keyboard or by simply switching to the numbers only view and back. Then it works the whole text long but re-appears next time. 

### Steps to reproduce

1. Long press a key 
2. Sometimes it work always sometimes not

### FlorisBoard Version

0.4.1

### Install Source

F-Droid

### Device

Pixel 7pro

### Android

13

patrickgold:
This PR disables [compose strong skipping mode](https://developer.android.com/develop/ui/compose/performance/stability/strongskipping), which causes all sorts of state bugs in FlorisBoard. Long-term goal is to support strong-skipping, but for now keep it disabled.

Closes #2629

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import org.jetbrains.kotlin.compose.compiler.gradle.ComposeFeatureFlag
 import java.io.ByteArrayOutputStream
 
 plugins {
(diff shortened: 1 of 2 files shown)


lm41 (approved):
Seems like that really was the problem. Looks good again now.

PR #2641, opened 2024-10-18
Fix migration in clipboard database

lm41:
This PR fixes the missing items in the clipboard database after updating to 0.4.2 by adding a migration (version 2 -> 3) to the clipboard database


patrickgold (approved):
Lgtm

PR #2644, opened 2024-10-19
Fix some labels in the IPA symbols layout

lucassanginetto:
In the IPA symbols layout, the bracket "⟪" is incorrectly labeled as "⟩", and the bracket "⟧" is incorrectly labeled as "⟦".


patrickgold (approved):
Thanks a lot for the fix, merging into main!

PR #2647, opened 2024-10-19
Fix state issues in TextKeyboardLayout (popups)

### LINKED ISSUE #2629: Long press popup does not appear often

### LINKED ISSUE #2642: Pop-ups dimensions stuck on portrait mode
### Short description

When turning the device into landscape mode then returning to portrait, the key pop-ups remain at the landscape dimensions. It's being solved by switching to the characters tab.

### Steps to reproduce

1. Turn auto-rotate on
2. Turn the device into landscape mode
3. Open keyboard
4. Go back into portrait mode
5. The key pop-ups are big, as if the screen is in landscape mode

### FlorisBoard Version

0.4.2 (100)

### Install Source

GitHub

### Device

Pixel 6a

### Android

GrapheneOS 14 (AP2A.240905.003.F1)

### LINKED ISSUE #2648: Toggling utility key causes spacebar and utility key misalignment
### Short description

There is a misalignment issue with the spacebar and utility key when toggling the `Show utility key` option. When enabling or disabling this option, the spacebar fails to resize correctly, leading to the partial visibility or hiding of the utility key.

To temporarily resolve the issue:

- `Force stop` the keyboard app and relaunch it (but before it choose the desired option)
- Long-pressing the `,` key and selecting the emoji key from its popup, then going back to main keyboard layout triggers a layout update. Or just press `?123`, or 3-dots settings, or any key that changea the layout. This behavior looks similar to the issues you fixed with #2625

### Recording

https://github.com/user-attachments/assets/b690266c-0994-4ae4-8c93-43b4373bef95

### EDIT
Ah, I should rec the video with the bordered theme. In border themes is more noticable that when the utility icon is hidden/removed, its border remains visible, creating an empty, non-functional inactive key area (even `:pressed` feedback style is not triggered)

fb_toggling_utility_key_bordered

### Steps to reproduce

1. Open the keyboard settings and locate the `Show utility key` option
2. Enable the option and observe the keyboard layout. The spacebar should shrink to accommodate the utility key, but it remains at its original size, causing the utility key to be partially hidden behind it.
3. Now, disable the `Show utility key` option. Notice that the utility key is no longer visible, but the spacebar does not expand to fill the available space.

### FlorisBoard Version

0.4.2 (100)

### Install Source

GitHub

### Device

OnePlus 5T (A14) and POCO F3 (A13)

### Android

A14 and A13

patrickgold:
This PR fixes state issues in TextKeyboardLayout and also removes unused smartbar keyboard support.

Closes #2629 
Closes #2642
Closes #2648

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/popup/PopupUiController.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/popup/PopupUiController.kt
@@ -48,12 +48,13 @@ import dev.patrickgold.florisboard.lib.toIntOffset
 @Composable
 fun rememberPopupUiController(
     key1: Any?,
+    key2: Any?,
     boundsProvider: (key: Key) -> FlorisRect,
     isSuitableForBasicPopup: (key: Key) -> Boolean,
     isSuitableForExtendedPopup: (key: Key) -> Boolean,
 ): PopupUiController {
     val context = LocalContext.current
-    return remember(key1) {
+    return remember(key1, key2) {
         PopupUiController(context, boundsProvider, isSuitableForBasicPopup, isSuitableForExtendedPopup)
     }
 }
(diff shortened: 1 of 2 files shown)


lm41 (approved):
:+1:

PR #2729, opened 2025-01-08
Fix colors in add subtype screen and enhance subtype presets dialog

lm41:
| Before the change | After the change |
|--------|--------|
| signal-2025-01-08-163843 | signal-2025-01-08-163843_004 |
|signal-2025-01-08-164844 | signal-2025-01-08-163843_003|


patrickgold (approved):
Thanks for your changes, tested them and lgtm!

PR #2731, opened 2025-01-11
Add subtypes and installed extensions to debug logs

lm41:
This PR adds two new information groups to the debug log. Installed theme extensions (can be expanded with more extension types in the future) and the currently selected and installed subtypes. These changes were made to create a better and more accurate environment for debugging bugs, and to better understand possible configuration-specific bugs.

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/lib/devtools/Devtools.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/lib/devtools/Devtools.kt
@@ -37,7 +37,7 @@ import java.io.InputStreamReader
 object Devtools {
     fun generateDebugLog(context: Context, prefs: AppPrefs? = null, includeLogcat: Boolean = false): String {
         return buildString {
-            generateDebugLogHeader(context, prefs)
+            append(generateDebugLogHeader(context, prefs))
             if (includeLogcat) {
                 appendLine()
                 append(generateLogcatDump())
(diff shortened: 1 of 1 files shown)


patrickgold (changes requested):
Something seems to cause an issue here, in my tests the debug log header is empty:

image

I can't figure out why your changes cause that behavior though...

lm41:
I can

app/src/main/kotlin/dev/patrickgold/florisboard/lib/devtools/Devtools.kt
```diff
@@ -34,6 +36,16 @@ import java.io.InputStreamReader
 @Suppress("MemberVisibilityCanBePrivate")
 object Devtools {
     fun generateDebugLog(context: Context, prefs: AppPrefs? = null, includeLogcat: Boolean = false): String {
+        return buildString {
+            generateDebugLogHeader(context, prefs)
```
lm41, line 40:
```suggestion
            append(generateDebugLogHeader(context, prefs))
```
I just messed up in the refactoring process.

patrickgold (changes requested):
Omg I didn't see that missing append, thanks for the quick fix!

I had a look now and it is a good addition to the log. Minor improvement regarding the subtype output:
1) could we print the active subtype on a newline, same as the enabled subtypes? the subtype string is already long enough and I think it would improve readability
2) I think "Enabled Subtypes" is confusing, a better fit would be the term "Installed Subtypes" or "Configured Subtypes"

patrickgold (approved):
Lgtm now, thanks!

Issue #2621, opened 2024-10-05
Add option to reset emoji history


itsmartashub:
### Feature idea

The emoji history feature is very useful for quickly accessing recently used emojis, but some users may want the option to reset or clear this history and start over. 

This issue proposes adding a reset functionality to the emoji history, allowing users to start fresh or remove all unwanted recent emojis in once instead manually remove them one by one (the new emoji option introduced in #2611 )

(We already discussed this in #2617)

PR #2732, opened 2025-01-11
Add option to delete EmojiHistory from settings screen

lm41:
Closes: #2621

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/media/MediaScreen.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/media/MediaScreen.kt
@@ -19,10 +19,16 @@ package dev.patrickgold.florisboard.app.settings.media
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.EmojiSymbols
 import androidx.compose.material.icons.outlined.Schedule
+import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import dev.patrickgold.florisboard.R
 import dev.patrickgold.florisboard.app.enumDisplayEntriesOf
+import dev.patrickgold.florisboard.app.florisPreferenceModel
 import dev.patrickgold.florisboard.ime.media.emoji.EmojiHistory
+import dev.patrickgold.florisboard.ime.media.emoji.EmojiHistoryHelper
 import dev.patrickgold.florisboard.ime.media.emoji.EmojiSkinTone
 import dev.patrickgold.florisboard.ime.media.emoji.EmojiSuggestionType
 import dev.patrickgold.florisboard.lib.compose.FlorisScreen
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/media/MediaScreen.kt
```diff
@@ -138,4 +164,49 @@ fun MediaScreen() = FlorisScreen {
...
+
+@Composable
+fun DeleteEmojiHistoryConfirmDialog(
+    shouldDelete: ShouldDelete?,
+    onDismiss: () -> Unit,
+    onConfirm: () -> Unit,
+) {
+    shouldDelete?.let {
+        JetPrefAlertDialog(
+            title = stringRes(R.string.prefs__media__emoji_history_delete_title),
```
patrickgold, line 196: The generic `action__delete_confirm_title` string exists for such delete confirmation titles

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/media/MediaScreen.kt
```diff
@@ -85,6 +99,18 @@ fun MediaScreen() = FlorisScreen {
...
+                onClick = {
+                    shouldDelete.value = ShouldDelete(false)
+                }
+            )
+            Preference(
+                title = stringRes(R.string.prefs__media__emoji_pinned_delete),
+                onClick = {
+                    shouldDelete.value = ShouldDelete(true)
+                }
+            )
```
patrickgold, line 113: These actions miss the `enabledIf = { prefs.emoji.historyEnabled.isTrue() },` condition

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/media/MediaScreen.kt
```diff
@@ -41,6 +50,11 @@ fun MediaScreen() = FlorisScreen {
     previewFieldVisible = true
     iconSpaceReserved = true
 
+    val prefs by florisPreferenceModel()
+
+    val shouldDelete = remember { mutableStateOf<ShouldDelete?>(null) }
```
patrickgold, line 55: COuld be improved by using the Kotlin delegates (will require 2 additional imports but Android Studio suggests them immediately)

```suggestion
    var shouldDelete by remember { mutableStateOf<ShouldDelete?>(null) }
```

patrickgold (changes requested):
Thanks for the changes. I just reviewed them and have some suggestions:
a) for consistency reason should pinned always come before recent (both in UI as well as in code/strings)
b) the string naming is inconsistent and should be aligned with the existing strings
c) I would prefer the term "Reset" over "Delete", as we basically reset the history back to empty

Other than that the functionality itself works well.

app/src/main/kotlin/dev/patrickgold/florisboard/app/settings/media/MediaScreen.kt
patrickgold: Just noticed if we change the term to Reset it would probably be best to add a generic `action__reset_confirm_title`

app/src/main/res/values/strings.xml
```diff
@@ -23,6 +23,8 @@
     <string name="prefs__media__emoji_history_pinned_update_strategy" comment="Preference title">Update strategy (Pinned)</string>
     <string name="prefs__media__emoji_history_recent_update_strategy" comment="Preference title">Update strategy (Recent)</string>
     <string name="prefs__media__emoji_history_max_size">Maximum items to keep</string>
+    <string name="prefs__media__emoji_history_pinned_reset">Reset pinned emoji</string>
+    <string name="prefs__media__emoji_history_reset">Reset emoji history</string>
```
patrickgold, line 27:
```suggestion
    <string name="prefs__media__emoji_history_pinned_reset">Reset pinned emojis</string>
    <string name="prefs__media__emoji_history_recent_reset">Reset recent emojis</string>
```

patrickgold (changes requested):
Thanks for the changes! Some minor stylistic issues persist, other than that the PR is good to go.

PR #2734, opened 2025-01-11
Fix NoSuchMethodError: removeFirst() on android 14 and lower

lm41:
https://developer.android.com/about/versions/15/behavior-changes-15#openjdk-api-changes


patrickgold (approved):
lgtm 👍

PR #2735, opened 2025-01-12
Add better detection for all keys visible bug

patrickgold:
This PR adds better detection for the "All keys invisible" bug, reported in https://github.com/florisboard/florisboard/issues/2362.


lm41 (approved):
lgtm :+1:

PR #2737, opened 2025-01-13
Fix issues with all keys invisible detection

patrickgold:
Fixes some issues introduced with https://github.com/florisboard/florisboard/pull/2735


lm41 (approved):
Works as intended with the change :+1:

PR #2743, opened 2025-01-16
Possible fixes for all keys invisible bug

patrickgold:
This PR tries to finally resolve the all keys invisible bug from #2362


lm41 (approved):
:+1: lgtm

PR #2757, opened 2025-02-06
Fix BottomSheetContent drawing behind system nav bar

lm41:
This commit fixes an issue where the BottomSheetContent was drawn behind the system navigation bar. To fix this issue, I applied the saveDrawingPadding Modifier to the AnimatedVisibility. There was a color problem on the system navigation bar, so I set the background color to SheetOutOfBoundsBgColorInactive. Now the system nav bar is colored the right way.


patrickgold (approved):
lgtm :+1:

PR #2758, opened 2025-02-06
Fix text being rendered behind the top app bar

lm41:
This commit fixes an issue where the text was rendered behind the top app bar in the crash dialog activity by removing the top app bar in the style and adding a top app bar in the layout XML.


patrickgold (commented):
Do you have a way to reproduce this error? i never noticed any issues with the current setup.

patrickgold (approved):

Issue #2684, opened 2024-11-04
Specify keyboard subtype when switching to voice IME


ElishaAz:
### Short description

After switching to a voice IME using the mic button (e.g. Sayboard), it is not possible to switch back to Florisboard using [InputMethodService#switchToPreviousInputMethod())](https://developer.android.com/reference/android/inputmethodservice/InputMethodService#switchToPreviousInputMethod()) in android 14.

However, it is possible if Florisboard specifies the subtype to switch to when calling [InputMethodService#switchInputMethod](https://developer.android.com/reference/android/inputmethodservice/InputMethodService#switchInputMethod(java.lang.String,%20android.view.inputmethod.InputMethodSubtype)).

The line https://github.com/florisboard/florisboard/blob/c470b792c110b3540ea7980987b55d5ac4eaf007/app/src/main/kotlin/dev/patrickgold/florisboard/FlorisImeService.kt#L237
Needs to be replaced with
```kotlin
ims.switchInputMethod(el.id, el.getSubtypeAt(i))
```

This only happens on Android 14 (It worked in android 13).

Related issue: https://github.com/ElishaAz/Sayboard/issues/86

### Steps to reproduce

1. Click on a text field
2. Click on the mic button to switch to a voice IME
3. Try to switch back to Florisboard using a button that calls `InputMethodService#switchToPreviousInputMethod()`
4. The keyboard does not switch back to Florisboard

### FlorisBoard Version

0.4.3 on F-Droid, and master(c470b79)

### Install Source

F-Droid

### Device

Nothing Phone (1)

### Android

14, crDroid

PR #2759, opened 2025-02-07
Fix switching back from voice ime (#2684)

lm41:
This commit fixes an issue where the voice ime was not possible to switch back to FlorisBoard, because the keyboard subtype was not specified.

Many thanks to @ElishaAz for pointing out the issue and providing a fix.
Closes: #2684


patrickgold (approved):
lgtm :+1:

Issue #352, opened 2021-02-14
Select Text button in smartbar


goyalyashpal:
Split from https://github.com/florisboard/florisboard/issues/341 (1)

#### Short description
Ttoggle select mode" can be added to smartbar in my opinion. With toggle select mode, i meant this button in text editor mode. There is just enough space for one more icon to come there.

![zjw3NnJ7ec](https://user-images.githubusercontent.com/19423063/107872959-6ca7f200-6ed4-11eb-89cf-5643db8d680d.png)

---
**Rough mockup:** \
the icon i chose is basically a combination of area icon ([see example here🡭](https://www.iconfinder.com/icons/6602651/area_design_graphic_select_selection_tool_icon)) and text cursor. The circle underneath was used to ensure equal spacing.
![photo_2021-02-12_23-30-39 (2)](https://user-images.githubusercontent.com/19423063/107873454-b2ff5000-6ed8-11eb-96b0-9316fde7f4f2.jpg)

---
**Use case:** 
i use an app called hovetext a lot. And it is VERY cumbersome to select text there as it doesnt show the android hovering text select options, rather it displays its own static toolbar which occupies the limited screen space - which becomes a recurring papercut. So, having quick access to enable select mode will be veryy very handy to me.

![image](https://user-images.githubusercontent.com/19423063/107874887-d3340c80-6ee2-11eb-96f3-1beabd45ec2f.png)

Glitchy-Tozier:
Just a little heads-up: You might also be able to use a long-press or gesture to switch to the selection keyboard.

yashpalgoyal1304:
> I really like the icon you created.

thanks :)

yashpalgoyal1304:
> Just a little heads-up: You might also be able to use a long-press or gesture to switch to the selection keyboard.

i couldnt find that. Where?

Glitchy-Tozier:
> > Just a little heads-up: You might also be able to use a long-press or gesture to switch to the selection keyboard.
> 
> i couldnt find that. Where?

I think i made that up... I vaguely remembered a discussion on the topic, but apparently it never was inplemented.

patrickgold:
> I think i made that up... I vaguely remembered a discussion on the topic, but apparently it never was inplemented.

We talked about different toolbars in the Smartbar - and that you can switch between them via gestures (which is still on the todo list). And we also talked about the fact that the current selection keyboard is very inaccessible. So that's the reason I think you vaguely remember.

[...]

PR #2782, opened 2025-02-27
Add select text button in smartbar

lm41:
Closes: #352


lm41:
See https://github.com/florisboard/florisboard/issues/352#issuecomment-2699223265

Issue #2806, opened 2025-03-10
Florisboard doesn't seem to work with the Android Terminal introduced in Android 15 QPR2


matchboxbananasynergy:
### Short description

Android 15 QPR2 allows you to enable a developer option to enable the Terminal app, which runs a Debian VM on your device.

When using the terminal with Florisboard, it doesn't seem to fully work. Pressing space seems to delete what you've typed, and pressing enter doesn't seem to input the command in the terminal.

### Steps to reproduce

1. Open the terminal app
2. Try to type within the terminal
3. Observe that spaces and the enter button doesn't function as expected.

### FlorisBoard Version

0.4.6

### Install Source

GitHub

### Device

Google Pixel 6

### Android

15 QPR2, GrapheneOS

lm41:
Fully reproducible for me.

lm41:
Input is working if all suggestions are disabled.

patrickgold:
I will have a closer look at that tomorrow evening and possibly make a cherry-picked bug fix release out of that, but as Lars already experimented seems to be an FlorisBoard issue

PR #2807, opened 2025-03-10
Do not ignore flagTextNoSuggestions.

lm41:
Fixes: #2806

@patrickgold would it be possible to add this to a v0.4 bugfix release?


patrickgold (approved):

patrickgold:
> @ patrickgold would it be possible to add this to a v0.4 bugfix release?

we should probably make an hotfix release yes, lets see

Issue #2746, opened 2025-01-18
Navigation Bar color issue.


Ansarmughal99:
@patrickgold @itsmartashub I can tell u I was having the Navigation bar color issue with Heliboard see pic 1 but after the recent update an option is added in appearance as color navigation bar, see pic 2 after this option there is no issue with color in navigation bar see pic 3 . Might this is not a system issue. 
Pic 1:- ![IMG_20250117_145540.jpg](https://github.com/user-attachments/assets/51e647f4-e661-4775-a927-1b8c77a43dea)

Pic 2 :- ![IMG_20250117_145458.jpg](https://github.com/user-attachments/assets/6d6b274b-774b-48e6-96fa-2f65f0f9a555)

Pic 3:- After Navigation bar color is set On ![IMG_20250117_145532.jpg](https://github.com/user-attachments/assets/7339c264-80e7-4a85-9d05-07a97396ebf6)

_Originally posted by @Ansarmughal99 in https://github.com/florisboard/florisboard/issues/2461#issuecomment-2597807186_

Ansarmughal99:
@patrickgold 
@lm41

Ansarmughal99:
@patrickgold 
@lm41 
This issue negatively impacts the user experience when using the keyboard.
I use the rise keyboard feature to give a dull feel as the bottom side of the keyboard is in  other color while uper in other color.

Ansarmughal99:
![IMG_20250214_201101.jpg](https://github.com/user-attachments/assets/6c0738af-eb4e-48d0-b799-15d63e645c9a)

klaurence:
This might have been fixed for the next release by #2757. @lm41, can you confirm?

lm41:
No. This PR specifically fixes the colors and contentPadding of the BottomSheet activated when editing the smartbar actions.

lm41:
@Ansarmughal99 what is 'the recent update' of heliboard. When posting something like this please be more specific in the future. This way it is easier for us to find the right change.

itsmartashub:
I finally managed to reproduce this issue almost every time.

TLDR: **Material You themes miscalculate colors on wallpaper change** causing system nav and keyboard background inconsistencies?

When using material you themes (floristyle), changing the wallpaper often leaves the **system nav unstyled/mistyled** and also causes the **theme to miscalculate** at first. Some elements, like the keyboard bg, retain colors from the previous wallpaper, while others update correctly. (Think I've mentioned this back then in #2461)

[...]

PR #2855, opened 2025-04-23
Theme rework part II (Snygg v2)

### LINKED ISSUE #256: Add image or animation as keyboard background
Make it possible to set an image as a keyboard background. An example of how this is implemented in GBoard: https://www.mobigyaan.com/set-background-image-in-google-keyboard-android
And, if technically possible, add the ability to use animation as the keyboard background (for example, to add an animated gradient).

### LINKED ISSUE #506: Allow selection of font for key labels
I wish there was a way to at least select between Sans-Serif, Monospace and Serif Fonts for the rendering of the Key labels on FlorisKeyboard.

As it stands now, the font used for rendering key labels is hardcoded in [KeyView.kt:101](https://github.com/florisboard/florisboard/blob/1f3221a8865350effc0c54a821243f43acf16cc5/app/src/main/java/dev/patrickgold/florisboard/ime/text/key/KeyView.kt#L101)

I would be happy to see a new setting, for example under Keyboard > Keys > Below the options for font size
allowing one to pick between the Options DEFAULT, MONOSPACE, SANS-SERIF, SERIF as defined in [android.com/reference/android/graphics/Typeface](https://developer.android.com/reference/android/graphics/Typeface)

A nice-to-have would be an option to pick between fonts installed on the system, one could use `Typeface.create("Name of font",Typeface.NORMAL)` to create an instance of that typeface.

Image shows FlorisBoard with a custom theme and typeface manually changed to SERIF using `typeface = Typeface.SERIF`
Screenshot_20210326-230230_Telegram_FOSS

### LINKED ISSUE #1547: Themeable shadows' colours
As the title says, I'm proposing to have the colours of shadows (under the keys and any other elements) to be themeable via the new theme engine.
This has already been discussed on the Matrix group, but I'm posting it here so it can be properly tracked.

### LINKED ISSUE #2015: Uneven key color surface on symbols
#### Short description
I made a borderless material you theme but sadly the space bar, the enter surface color shape size does not match.
Screenshot_20220814_142432_com microsoft emmx

#### Steps to reproduce
1. Go to manage theme
2. Create a borderless theme
3. Change shape of symbols key, space and enter to circle 
4. Put color surface for those keys
5. See error, the three keys have uneven shape

#### Environment information
- FlorisBoard Version: 0.4.0-alpha02 (88)
- Install Source: Github
- Device: HUAWEI NOVA 7SE
- Android:ANDROIS 10, EMUI 12

### LINKED ISSUE #2330: Enhancements for Theme Engine (Snygg)
I am aware that there are more pressing matters here, but let this suggestion stand for consideration, and perhaps it will be addressed at some point.

Namely, the theme editor you've conceived and created is something truly unique, unlocking creativity, and I thank you all guys for enabling such extensive keyboard personalization.

While creating a Gboard-like UI, I encountered a few things that I couldn't resolve, and it might be useful to add in general:

**1.  Add properties for `padding`, `margin`, `gap`**
- Somewhere, I managed to simulate the gap between elements with `border-width` and `border-color` (like in `action key tails`, `sticky action key`, etc.). However, this doesn't always have the desired impact or doesn't work at all, or the element doesn't support those properties. Besides, using borders for spacing is a bit of a silly solution on its own, I admit 🥲
&nbsp;&nbsp;&nbsp;

**2.  Separate the design of `action keys` from `sticky action key`**
- If a sticky key (e.g., mic🎙️) has a specific design, perhaps someone doesn't want their sticky key to be a mic but incognito or something else, and reordering them using the `Reorder Actions` button, the design remains bland for the mic, and it looks inconsistent when the key with the sticky design is moved to the `action row`. Also, the new key that is now in place of the `sticky action key` does not have the desired sticky design. Huh, if you managed to understand anything of what I said.

florisboard--sticky-action-key
&nbsp;&nbsp;&nbsp;

**3.  Enable change value of `font-size` in `action keys tail`**
- The `font-size` property exists there, but adding any value doesn't seem to have an impact, or perhaps I'm doing something wrong?
&nbsp;&nbsp;&nbsp;

**4.  Perhaps implement `font-weight` property?**
&nbsp;&nbsp;&nbsp;

**5.  Consistent `height`/`padding` in the keyboard footer?**
- I'm not sure what this is related to, but when using a borderless theme, the submit button and space have a smaller height (or padding) than keys `?123`/`ABC` (`-202`/`-201`). While in the theme editor, it looks as it should, but when using the keyboard, somehow the height/padding decreases.

florisboard--footer-height
&nbsp;&nbsp;&nbsp;

**6.  Separate design for all elements of the num layout.**
- Separate design from keys when the num layout is only displayed because if we want to change the design for characters `+`, `(`, `)`, `-`, `%`, `=`, etc., they will be altered on every layout, and that seems odd. But I understand if this might be complicated to implement/rewrite.

florisboard--num-layout
florisboard--num-character

&nbsp;&nbsp;&nbsp;

There are some more nuances, but I won't overburden you with them now.

### LINKED ISSUE #2549: Use system default font for hinted symbols, same as HeliBoard
### Feature idea

Screenshot_20240904-214818_FlorisBoard Beta
Screenshot_20240904-214829_FlorisBoard Beta

patrickgold:
# Snygg v2 Changelog
### New general properties (closes #2330)
- Improve `font-size` property => customize the font size in `sp` units
- Add `font-style` property => customize the font style (`normal`, `italic`)
- Add `font-weight` property => customize the font weight (`thin`, `extra-light`, `light`, `normal`, `medium`, `semi-bold`, `bold`, `extra-bold`, `black`)
- Add `letter-spacing` property => customize the letter spacing of texts in `sp` units
- Add `line-height` property => customize the line heights of texts in `sp` units
- Add `margin` property => customize the spacing between the other elements and the border
- Add `padding` property => customize the spacing between the border and the content
- Add `shadow-color` property => customize the shadow color (except for UI buttons) (closes #1547)
- Add `text-align` property => customize the text align (`left`, `right`, `center`, `justify`, `start`, `end`)
- Add `text-decoration-line` property => customize the text underline (`none`, `underline`, `line-through`)
- Add `text-max-lines` property => customize the text max lines
- Add `text-overflow` property => customize the text overflow behavior (`clip`, `ellipsis`, `visible`)
- Add `clip` property => control if the `shape` clips the element. (`yes`, `no`, defaults to `yes`)

### Variable changes
- Variable names must now start with `--`
- Variables do *not* support multi-level resolving currently

### Inherit behavior changes
- Add full support for multi-level property value inheriting from parent element
- There are two modes:
  - Explicit inherit => works for all properties, but needs to be explicitly defined
  - Implicit inherit => works automatically for selected properties: `foreground`, `font-family`, `font-size`, `font-style`, `font-weight`, `text-align`, `text-decoration-line`, `text-max-lines`, `text-overflow`
- This might break existing themes upon migration, so keep this in mind!

### Navigation bar color changes
- Nav bar color is now controlled by the main surface color (so `keyboard` or a bottom sheet background)
- Remove `system-nav-bar` element
- Related #2746 but not fixed

### Custom fonts (closes #506)
- Add `@font` rule for declaring custom fonts
  - Array of object: (only use 1 item if you have a bundled font file, or multiple if you have split font files)
    - Required property `src`: specify URI path to font file within archive
    - Optional property `font-style`: Specify the font style of the font file, used by the system font matcher
    - Optional property `font-weight`: Specify the font weight of the font file, used by the system font matcher
- Add `font-family` property in elements => reference a generic font (`system`, `sans-serif`, `serif`, `monospace`, `cursive`) or a custom font
- Key hinted label is now not forced to monospace anymore => closes #2549

### Custom background image (closes #256)
- Add `background-image` => customize the background image
  - We use the [coil](https://coil-kt.github.io/coil/) library for this, so any file types supported by coil should be supported (including GIFs !!). We do not intend to support additional image types that coil does not support.
  - This property only has an effect if used in an `keyboard` element.
- Add `content-scale` property => customize the background image fitting behavior (`crop`, `fill-bounds`, `fill-height`, `fill-width`, `fit`, `inside`, `none`)

### Borderless themes
- Remove `isBorderless` theme meta attribute
- `margin` property now controls the squashing of the key height together with specific rules (e.g. `key[code=10]`)
- Fixed #2015

### Schema
- New schema (available at https://schemas.florisboard.org/snygg/v2/stylesheet).
- You must annotate your stylesheet json with a `$schema` attribute with exact value  `https://schemas.florisboard.org/snygg/v2/stylesheet`, else the parser will unconditionally reject your stylesheet.


lm41 (approved):
lgtm :+1:

PR #2873, opened 2025-05-09
Fix Smartbar quick actions duplicating in prefs migration

patrickgold:
Originally reported in Matrix group chat.


lm41 (approved):
:+1: Looks good to me

PR #2897, opened 2025-05-13
Fix Snygg and IME UI state bugs regarding theme state updates

### LINKED ISSUE #2884: Theme doesn't load upon first app-start
### Short description

When starting the app after updating it or stopping its execution via the settings, it uses the wrong(/fallback?) theme. Every key you press reverts to the original theme. When pressing shift, the whole keyboard resets itself.
First occurence:
https://github.com/user-attachments/assets/65987ced-6a01-44be-be69-86c11542c989

Manual activation by stopping the app:
https://github.com/user-attachments/assets/51af6bee-363f-4bc1-9811-4b0ee3c720af

### Steps to reproduce

1. Go to FlorisBoard's settings
2. Stop execution
3. Go to any text-field and look at the keyboard

patrickgold:
Also enable Compose strong skipping mode

Fixes #2884

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -15,7 +15,6 @@
  */
 
 import java.io.ByteArrayOutputStream
-import org.jetbrains.kotlin.compose.compiler.gradle.ComposeFeatureFlag
 
 plugins {
     alias(libs.plugins.agp.application)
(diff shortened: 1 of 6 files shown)


lm41 (approved):
:+1:

PR #2899, opened 2025-05-13
Add `mode` attribute to select element styles

patrickgold:
This PR adds the `mode` attribute, which allows to further customize the theme based on the active mode.

The following attribute values are supported:
- `characters`
- `symbols`
- `symbols2`
- `numeric`
- `numeric_advanced`
- `phone`
- `phone2`

It can be applied to the following elements:
- `root`
- `window`
- `key`
- `key-hint`
- `key-popup-box`
- `key-popup-element`
- `key-popup-extended-indicator`


lm41 (approved):
Looks good to me :+1:

PR #2871, opened 2025-05-09
Add Armenian layout to extension.json

MikayelB:
The [armenian_alt_phonetic.json](https://github.com/florisboard/florisboard/commit/a7f8980d350aedaa6af26083e678fbca5b3d7833#diff-2b961f94cbb97e3bf1d5ec590214ff2fc568e200c8e701d73a4a5d075ef236c0) was configured (commit [a7f8980](https://github.com/florisboard/florisboard/commit/a7f8980d350aedaa6af26083e678fbca5b3d7833)) but wasn't properly added to extension.json, that's why it doesn't show up in the layout menu.

Thank you all for your hard work :)

PR #2904, opened 2025-05-14
Fix Amenian phonetic layout addition

MikayelB:
Sorry for the second PR on the same topic, but I previously didn't add the Armenian layout properly (I thought I did)(#2871)

It is now appearing this way (no "enter"/"shift" etc)

Screenshot From 2025-05-14 12-23-49

And here is the fixed version (this PR's changes)

Screenshot From 2025-05-14 12-23-33

I properly checked, the changes are still minor but correct this time.
Sorry for the inconvenience


patrickgold (approved):
Thanks a lot for the fix, tested it out and seems to work fine, merging in!

PR #2918, opened 2025-05-20
Fix ignored paths not applied to `pull_request` in workflow

patrickgold:


patrickgold:
Does not work as expected

PR #2920, opened 2025-05-20
Fix theme editor navbar color bug

### LINKED ISSUE #2909: Navigation bar icons colour not adapting in theme editor
### Short description

If you create a day light theme and open the theme editor, the icons on the navigation bar, which should turn dark, stay white. This though, only happens if the theme editor is opened and only for the light theme. Outside the theme editor, the navigation bar adapts as it should. Also, you don't need to be in FlorisBoard to trigger it, if you leave the editor opened and switch to another app, the issue is still present.

In the theme editor

Image

Outside the theme editor

Image

### Steps to reproduce

1. Go to FlorisBoard settings
2. Create a light theme
3. Open the theme in the theme editor
4. See the icons stay white on there light background

### FlorisBoard Version

0.5.0-alpha03 (107)

### Install Source

GitHub

### Device

Poco F2 pro

### Android

15, crDroid

patrickgold:
Fixes #2909


lm41 (approved):
:+1:

PR #2932, opened 2025-05-30
Implement QWERTY (Ç) layout

### LINKED ISSUE #1939: Add QWERTY(Ç) layout
this layout is very similar to QWERTY, for as presented, it has the character "Ç" located next to the letter "L" in the second row of letters, this design is applicable in PT-BR, I would like to implement it, thanks in advance!

klaurence:
Closes #1939


satreks:
Friend, thank you very much for taking the initiative to implement the code...
I'm not a programmer, so please don't take me too seriously regarding technical aspects of the code — I'm just a user and a bit curious.

That said, here’s my point of view...

I noticed that you removed the Catalan layout and created a new one called QWERTY (Ç).
However, I don't think this is ideal for both the Catalan and Brazilian user communities, because the pt-BR layout without the "Ç" next to the "L" will still exist, and the Catalan layout will be removed.

So what will probably happen is that the Catalan user won't find their layout, and the
Brazilian user will still have the layout without the "Ç" next to the "L".

I know that if both users scroll down a bit, they will find the new QWERTY (Ç) layout,

but since the app is categorizing the layouts by language/region (which is at least implied by the way the layout names are organized),
I think the most appropriate approach in this case would be to keep the Catalan layout in the list and just add the "Ç" next to the "L" in the existing pt-BR layout.

What do you think?

klaurence:
I can't test if the migration mechanism works because I only have access to debug builds, which have different signatures and can't be updated/migrated.

klaurence:
Undrafting this PR. It's theoretically done and implemented, but it just needs confirmation that the migration mechanism works.

klaurence:
I realized that I did have a way of testing it, and the migration doesn't work. Drafting.

satreks:
And now? How do we make the blessed "Ç" appear next to the "L"?

PR #2947, opened 2025-06-09
Add subscripts and missing fractions in number row popups

### LINKED ISSUE #2805: Support inputting subscript numbers
### Feature idea

Inputting subscript numbers should be supported.

Superscript numbers are already supported.

Maybe expand support for subscript and superscript symbols too.

eqilnou:
Added subscripts to the popups in the number row. Closes #2805.
Also added the two missing fractions `⅟` and `↉`.

In the Western Arabic and Thai layouts, the subscript is always directly to the left of the main popup, except for the key `1` since there is no space to the left: in this case the subscript is directly up. (Also, is it intended behaviour that Thai does not have the Western Arabic symbols `1`, `2`, ... in the popup? All other layouts have them as main popup symbol.)
The layouts CJK and neo2 are unchanged, since they did not have superscripts nor fractions (except superscript `1`,`2`, `3` in neo2). Let me know if you want me to change this.
In all other layouts, which have the Western Arabic digit as main popup, the subscript is always to the left and the superscript always to the right of the main popup, except `1`, which has the subscript directly up, and `0`, which has the superscript directly up.

Also let me know if the ordering of the various symbols in the popups is good.

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/numericRow/bengali.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/numericRow/bengali.json
@@ -3,46 +3,52 @@
     { "code": 2535, "label": "১", "type": "numeric", "popup": {
       "main": { "code":   49, "label": "1" },
       "relevant": [
-        { "code": 8537, "label": "⅙" },
+        { "code": 8321, "label": "₁" },
         { "code": 8528, "label": "⅐" },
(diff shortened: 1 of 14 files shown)


eqilnou:
While working in the number row files I also noticed that in the CJK layout there is a typo in the popup for the digit 4 so I fixed it (apparently someone had already attempted this fix in #1403).
Do you prefer a separate pull request for this?

lm41 (approved):
Thank you for the changes. LGTM :+1:

PR #2970, opened 2025-07-11
Added logic to generate word suggestions and spellcheck

RadicalMuffinMan:
# Description

This PR implements a complete native Rust NLP engine for word suggestions and spell checking, it also addresses the memory constraints discussed in #677.

## Key Features
 - Native Rust Implementation
- All dictionary data and NLP processing runs in native memory, avoiding JVM heap limitations
- Trie-based data structure for efficient prefix matching and word lookup
- LZ4 compression for 40% smaller dictionary files with fast decompression (1GB/s)
- Language switching works seamlessly
- Personal dictionary persists across sessions and languages
- Multi-language word prediction based on the user's chosen subtypes, all languages are equal. Word prediction is based on the word's weighted frequency.
- Added n-grams support to help new users who haven't created a personal prediction library have easier use of the app
- Word predictions includes pronouns, acronyms, contractions, and french elisions 

## Dictionary Source
Word frequency lists sourced from [hermitdave/FrequencyWords](https://github.com/hermitdave/FrequencyWords) (MIT License), curated to remove offensive words while preserving dual-purpose terms (e.g., animal names that can be insults).

### Kotlin

- LatinLanguageProvider.kt - Multi-language binary dictionary loading
- NlpBridge.kt - JNI bridge for native calls
- AppPrefs.kt - Added smartbar.autocorrect preference
- EditorInstance.kt - Auto-shift after sentence endings

### Rust

- nlp - Core NLP engine with trie, suggestions, spell check
- dict-compiler - Tool to compile JSON → binary dictionaries
- rust - JNI bindings

# APK testing

# Checklist

- [X] I have read and understood the [contribution guidelines](https://github.com/florisboard/florisboard/blob/main/CONTRIBUTING.md).
- [X] I have read and understood the [AI policy](https://github.com/florisboard/florisboard/blob/main/AI_POLICY.md).


patrickgold:
Hi, before I can take a closer look at your changes, what is the source for the data.json and under which license is it distributed originally?

RadicalMuffinMan:
> Hi, before I can take a closer look at your changes, what is the source for the data.json and under which license is it distributed originally?

Originally I used the file that was already in the project, just reformatted it into proper json, capitalized some things, and it worked. Afterwards I felt like the list was too short so I sought out a bigger one and found it [here](https://github.com/dwyl/english-words?tab=Unlicense-1-ov-file#readme), which has an "unlimited license".

In the future I want to see if we can either separate each language into it's own separate json file or nest it under this main one.

patrickgold (commented):
Thanks for clarifying.

I have had a closer look at your proposed idea. It's major flaw is that it is implemented in Kotlin, which then in runtime runs in the JVM heap. While for normal logic and UI this is perfectly fine, it is a big issue for anything computationally and memory intensive. If you load the dictionary into a Kotlin map, it is stored in the JVM heap, which is quite limited. If you exceed the limit, it results in an `OutOfMemoryError`, which crashes the whole keyboard. Our first predictive text approach (a long time ago) was also written in Kotlin, so we have first-hand experience with these errors. See also the infamous https://github.com/florisboard/florisboard/issues/677 issue.

Our current approach is to implement it natively in Rust. Native has the big advantage that a) data can be stored with less overhead (no garbage collector metadata), and b) we just have a lot more space in native memory available. Also Unicode provides first-class ICU crates for text manipulation, which is a must for future-proofing the algorithm for non-English languages (think accents on letters, differing concepts in scripts, etc).

From your code I can see that you seem to have some experience with predictive text, or at least some idea on how to approach it. I also have some ideas how to approach it, but from a different angle, and with important constraints in mind. Maybe it would make sense if we get together and combine our ideas, what do you think?

RadicalMuffinMan:
@patrickgold it's ready, friend.

patrickgold:
Thanks a lot for the changes!

I am really impressed, my first impression is quite good! We are currently doing an initial internal review of your changes and will come back to you.

Some questions for now:
1. The NLP engine seems to support current word prediction and spell check. Would the implementation be extensible in the future to also support next word prediction? Or would we need a separate NLP engine for that?
2. How difficult would it be to extend the NLP engine to support multi-language prediction in the future? E.g. en+de at once.
3. The personal dictionary/learning is shared between languages, is this intended?
4. You seem to be quite talented in this area of development. Would you be interested to stay around to develop/maintain NLP-related features in FlorisBoard?

RadicalMuffinMan:
1. I was thinking about that today after using it for a while, let me look more into it and see how I can make it work.

2. Not incredibly hard, i did find having to create subtypes to be able to switch languages a little cumbersome so i might need to work around that

3. Yes

4. I'm also working on my own little projects plus regular work but I can try to help here and there.

RadicalMuffinMan:
@patrickgold so I added predictive text support but wanted to go one step further:

How Next-Word Prediction Worked

**Without N-grams:**
- User types "how " --> Keyboard has NO predictions until user teaches it
- User has to type "how are" multiple times before keyboard learns the pattern
- Cold start problem: new users get zero predictions

**With N-grams:**
- User types "how " --> Keyboard immediately suggests "are", "is", "do" from n-gram data
- Works from first use, no learning required
- As user types more and more, learned patterns override/supplement n-grams

## Data Flow

```
User types "how " (space pressed)
            ↓
    predict_next_word(["how"])
            ↓

┌────────────────────────────────────┐
│  1. Check learned associations     │  ← User's personal patterns (10x weight)
│     context_map["how"] → {"are": 5}│
│                                    │
│  2. Check n-gram data              │  ← Pre-built baseline (1x weight)
│     ngrams["how"] → {"are": 100,   │
│                      "is": 90, ...}│
└────────────────────────────────────┘

            ↓
Merge & rank by weighted score
            ↓
Return: ["are", "is", "do", ...]
```
    
## Priority System

| Source | Weight | Purpose |
|--------|--------|---------|
| Learned (context_map) | 10x | User's personal typing patterns |
| N-grams | 1x | Baseline predictions for everyone |

So after typing "how are" 2-3 times, your learned "are" will outrank any n-gram suggestion.

## Files Involved

- **language.ngrams.json** --> Loaded at startup, provides instant predictions
- **context_map.json** --> Persisted to disk, stores your learned patterns
- Both merge in `predict_next_word()` with learned data prioritized

The n-grams are basically like  a "starter pack" so the keyboard works intelligently from day one, then adapts to each user's style over time.

Now I'm trying to fix contractions and pronouns because it's annoying me. Also I added a Latin America Spanish dictionary

RadicalMuffinMan:
Also I think that Suggestions & Actions extended to be the best default UI for the text prediction in the smartbar, but that's just me

RadicalMuffinMan:
All right, pronouns, acronyms, contractions, and french elisions are added

RadicalMuffinMan:
@patrickgold I added multi language text prediction based on the user's chosen subtypes. All languages are weighed the same to keep it fair, just goes off the word's weighted frequency. Did you want me to try and add a json and dict for all languages available in the subtypes?

patrickgold:
Many thanks for your recent changes, I am really impressed! Will have a look at the changes later today.

> Did you want me to try and add a json and dict for all languages available in the subtypes?

I would argue no, except if you have a good data source to create a new dictionary for an additional language. Other languages can always be added as new dictionary extensions.

Some of the languages (e.g. Chinese, ...) require highly specialized engines anyways, thus dictionaries for these languages would make no sense.

RadicalMuffinMan:
I can add as many of the Latin based languages as possible from the locations referenced in the PR comment, I know non-latin laguages are harder to implement

patrickgold:
If the source has good word data for all the language languages then why not.

Please note that the actual dictionary data will not be merged into the core repo anyways, but a separate addons data repo, but the infra is not ready yet, will be in the following days. So you can generate the dict data, but you can wait with actually committing them when our data repo is ready.

The dictionaries generated by you will also be wrapped into dictionary extension packages, but that is also not that ready yet (you caught us off-guard timing wise with your changes a bit :smile:), I will keep you updated with our progress.

RadicalMuffinMan:
Gotta keep you on your toes 😉 all right I'll remove all but english from the PR and create the other one's and have them ready

patrickgold:
Quick question: what is the difference between the `<lang>.json` and `<lang>.dict` files?

RadicalMuffinMan:
The `<lang>.json` files are human-readable word lists with frequency scores ranging from 0-255. These files serve as the source material used to generate the `.dict` binary files and are not shipped in the app. They're only really used during the build and development process. Kept the english one as a template/example.

The `<lang>.dict` files are binary compiled versions of the `.json` files. They use an optimized format that provides smaller file sizes, faster loading times, and includes canonical forms for contractions and acronyms. These are the files that the app actually loads at runtime.

patrickgold:
That explains a lot, thanks. Does the same also hold for the `<lang>.ngram.json` files, or are these shipped as jsons?

RadicalMuffinMan:
The `<lang>.ngrams.json` files contain word association data used for next-word prediction. They map which words commonly follow other words and are loaded separately to enable predictive typing functionality.

At runtime in the app, only 2 files are needed. The `<lang>.dict` file is required for word suggestions and spell checking, while the `<lang>.ngrams.json` file is optional but needed if you want next-word prediction to work.

The ngrams and the user's learned patterns work together throughout. When predicting the next word, the engine queries both simultaneously and combines their scores:

- Context map (user's learned patterns) gets checked and scores are multiplied by 10x weight so user patterns are heavily prioritized

- Language ngrams (from the `<lang>.ngrams.json` file) gets checked with normal weight

Both contribute to the same prediction list, but the context map has 10x stronger influence. So if you often type "I'm going home" after "work", that learned pattern will score much higher than the generic ngram data.

RadicalMuffinMan:
I also made another PR that fixes the app crashing on banking apps due to the numeric input

patrickgold:
Quick question regarding the dictionary/json files: as far as I see, the binary dict files have a magic header for identification and version field for forward compatibility, while the ngrams.json doesn't. Is this intentional? I am asking because we are currently drafting the dictionary decl format within extension formats and need to decide if the extension.json needs to handle schema versions or if the dictionary files themselves handle this.

RadicalMuffinMan:
The `.dict` binary files have a proper versioning scheme with a 4-byte magic header (`FBTD`), a version field (currently v3 for canonical forms support), and a checksum for integrity validation. The `.ngrams.json` files do not have any versioning and are just plain JSON with an implicit schema of `{"word": {"follower": frequency}}`. This was not an intentional design decision, it was just the easier approach at the time of implementation.

For the extension format, I would recommend handling schema versions in `extension.json` rather than adding versioning to the ngrams files themselves. This keeps the ngrams file as pure data which is easier to generate and process with external tools. The extension manifest already serves as the metadata declaration for its contents, so it makes sense to declare dictionary schema versions there as well. This approach also allows a single extension to potentially contain multiple dictionary formats if needed in the future.

The binary `.dict` files can potentially continue using their internal versioning since they require the header anyway for magic bytes and checksum validation. For ngrams and any future plain-data formats, the extension manifest can declare something like `"ngramsSchema": 1` to indicate the expected format version.

patrickgold:
Thanks for the quick reply. Our current draft for the new `keyboard3` extension (which unifies the existing `keyboard` and `languagepack` extensions) would define Latin dictionaries like this:


```jsonc
{
  "$": "ime.extension.keyboard3", // new ID, will replace current keyboard and languagepack formats in the future
  "meta": {
    "id": "org.florisboard.languages.en_US",
    "version": "0.1.0",
    "title": "Language Pack English (US)",
    "description": "yadayada",
    "maintainers": [ "author1 <author1@example.com>" ],
    "license": "MIT"
  },
  "dictionaries": [
    {
      "$": "ime.dictionary.latin",
      "lang": "en_US",
      "dictVersion": 1,
      "dictPath": "dictionaries/en_US.dict",
      "ngramsVersion": 1,
      "ngramsPath": "dictionaries/en_US.ngrams.json"
    },
  ]
}
```

The `"$": "ime.dictionary.latin"` ID allows us to distinguish between dicts targeting the Latin NLP engine and dicts for other NLP engines.

With this approach the versions are declared in the extension manifest, and will then be passed along to the NLP engine.

RadicalMuffinMan:
This design looks great and addresses the versioning issue. Having both dictVersion and ngramsVersion in the manifest keeps ngrams as pure data while maintaining forward compatibility so that's definitely a plus.

A small note:
- Maybe consider making ngramsPath/ngramsVersion optional since ngrams aren't required for basic suggestions (only next-word prediction)

The binary .dict files already do version checking in their header, so the dictVersion here would be more for the Kotlin layer to verify compatibility before passing to the NLP engine. For ngrams, the Rust side currently does no validation, it just tries to parse the JSON and fails if the structure is wrong.

Should I implement version validation in the NLP engine to handle these fields?

patrickgold:
No thanks, we are currently further reviewing the added code.

One thing that came up during further review of the PR is that we need to maintain the code in the future, with our level of knowledge. Can you explain to us how you came up with the implementation that quickly, and did you use existing resources to base your implementation on?

RadicalMuffinMan:
These from what tabs I still got open:

https://medium.com/basecs/trying-to-understand-tries-3ec6bede0014

https://en.wikipedia.org/wiki/Radix_tree

https://github.com/BurntSushi/fst

https://blog.burntsushi.net/transducers/

https://chromium.googlesource.com/chromium/src/+/main/chrome/browser/spellchecker/spellcheck_custom_dictionary.h

https://android.googlesource.com/platform/packages/inputmethods/LatinIME/+/refs/heads/main/native/jni/src/dictionary/structure/v4/

https://crates.io/crates/lz4_flex

https://github.com/facebook/zstd#benchmarks

https://en.wikipedia.org/wiki/N-gram 

https://web.stanford.edu/~jurafsky/slp3/3.pdf

https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/37567.pdf

https://en.wikipedia.org/wiki/Levenshtein_distance

https://developer.android.com/training/articles/perf-jni

https://mozilla.github.io/firefox-browser-architecture/experiments/2017-09-21-rust-on-android.html

https://norvig.com/spell-correct.html

https://arxiv.org/abs/1511.01432 

plus lots of coffee and claude to tell my why I'm stupid and my code didn't work

patrickgold:
Thanks for sharing the resources used for the implementation, and admitting the usage of generative AI (Claude) for parts of the code.

Please note that the use of generative AI is not something we see positively, as stated in the [AI policy](https://github.com/florisboard/florisboard/blob/main/AI_POLICY.md) you agreed to in the original PR comment. It adds additional overhead to us, **human** reviewers, to not only check the code for potential issues and functionality, but also authenticity, potential copyright infringement, and missing attribution caused by AI. We do not know how similar your approach is to e.g. LatinIME or the Chromium spell checker dictionaries you listed, and if that necessitates attribution. You were able to list the sources used, which is a big plus, and we think in general you have the necessary skills and technical knowledge to develop code yourself, however parts of your submitted code and some of the comments in this reply chain are heavily influenced by generative AI, which is a violation of our AI policy.

We have discussed internally how to proceed, and think that the general idea and approach of the word prediction engine is okay and in simple tests it performs well, however given that AI was involved and left its traces the current state of the PR is not merge-able. We thus propose the following:

- You go over the code and ensure that, where similar approaches are taken, proper attribution to the respective open-source project/paper is given
- You ensure that the code is cleaned from "AI artifacts" and fully generated AI files are removed / replaced by your own code
- Further contributed code is not the product of generative AI.

I hope you can understand our standpoint and why we need to enforce the policy.

jwsp1:
@RadicalMuffinMan Would it be possible for you to take a look at this open pr, or alternatively create a fork and release an app with word suggestions and spellcheck for those who don't mind AI-enhanced code?

PR #2987, opened 2025-07-19
Remove enforcement for navigationBarContrast

lm41:
We can remove this enforcement because we don't draw there anyway. See this SO answer https://stackoverflow.com/a/78237081


patrickgold (approved):
Looks good to me, thanks for the fix!

For the best solution new properties allowing direct control of `isAppearanceLightNavigationBars` and `isNavigationBarContrastEnforced` should be added, but that is only possible once https://github.com/florisboard/florisboard/issues/2993 is implemented.

PR #2991, opened 2025-07-21
Move most of lib.compose to its own module

lm41:


lm41:
I've rebased on the jetprefv3 changes

patrickgold (approved):

PR #2992, opened 2025-07-21
Create python-package-conda.yml

joe10833:


patrickgold:
We are not using Python in this project???

joe10833 (commented):
Thanks

PR #3004, opened 2025-08-01
Migrate to type save navigation

lm41:
This PR changes the navigation logic to the type-based navigation introduced in version `2.8.0` of `androidx.navigation:navigation.compose`. The logic of handling deep-links has been adjusted to work with the new type based approach. 

This PR also updates the predictive back animation.


patrickgold (approved):
lasagne tastes good :D 

(on a serious note, lgtm :+1:)

PR #3008, opened 2025-08-02
Allow emoji suggestion on emoji name and keyword

sammypanda:
## solution: filtering "name OR keyword", instead of "name AND keyword" for emoji suggestions

> I would probably look at emojicon.py but i couldn't find where that is. Also this solves it without having to double up the original name as a keyword -- it feels safe to say since i can't imagine you wouldn't want the emoji name itself as a keyword always.


patrickgold (approved):
Thanks for the change. I reviewed the behavior and I think it works fairly well and the results are not as polluted as I first thought.

Merging in, will be released in 0.5.0-beta04.

PR #3010, opened 2025-08-02
Add forward delete behavior

### LINKED ISSUE #450: Add forward delete behavior to delete key
In Samsung keyboard I found a feature to delete text from left to right by pressing on the capslock icon and then pressing the delete icon while holding the capslock.
A feature like this on FlorisBoard would be great.

patrickgold:
Add forward delete behavior to the existing delete key setup. Implementation details:

delete key | shift? | behavior
---|---|---
tap | no | deletes 1 char/word before cursor
tap | yes | deletes 1 char/word after cursor
tap&hold | no | deletes N chars/words fast before cursor
tap&hold | yes | deletes N chars/words fast after cursor
swipe | no | selects N chars/words and deletes on release before cursor
swipe | yes | selects N chars/words and deletes on release after cursor

Closes #450

TODO:
- [ ] Add explicit forward delete key

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/FlorisImeService.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/FlorisImeService.kt
@@ -675,10 +675,13 @@ class FlorisImeService : LifecycleInputMethodService() {
         }
     }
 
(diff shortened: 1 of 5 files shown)


lm41 (approved):
Thank you for the changes! lgtm :+1:

PR #3013, opened 2025-08-04
Inline autofill correct draw behavior

### LINKED ISSUE #1228: Autofill inline chip graphical issues.
#### Short description
The android 11 autofill chip blocks the key press and long press popups, and it doesn't have a shadow

#### Steps to reproduce
1. Select a password entry box
2. Press / long press on a key near the top
3. See the blockage & missing shadow

#### Environment information
- FlorisBoard Version: 0.3.13
- Install Source: F-Droid
- Device: Samsung Galaxy S10+
- Android: 11 oneui 3.1

patrickgold:
This PR fixes the inline autofill behavior, popups are now correctly drawing over inline autofill chips:

inline_autofill_demo

Pros:
- Works with all window background colors
- Works with all **static** window background images
- Works with all **dynamic** window background images (GIFs)

Cons:
- Potentially more error-prone state logic for background

TODO:
- [x] Re-add content scale support
- [x] Re-add GIF support
- [x] Evaluate GIF frame rate, may need to be adaptive

Closes #1228


lm41 (approved):
Thank you for the fix for this nasty bug. lgtm :+1:

PR #3015, opened 2025-08-05
Disable stylesheet overlay in bfu mode

lm41:
This PR has disables the message about the failed loading of the stylesheet in BFU mode, as this is expected there.


patrickgold (approved):
Thanks a lot for the fix!

PR #3019, opened 2025-08-05
Fix not showing the background image in some situations

lm41:


patrickgold (approved):
This should fix the issue with the quick switch render issue, thanks a lot!

PR #3020, opened 2025-08-06
Switch to AndroidExternalSurface for bg rendering

patrickgold:


patrickgold:
This produces more bugs than the current impl, closing

PR #3079, opened 2025-09-03
Adding Korean Phonetic layout 

Shunnuo:
Adding Korean Phonetic layout (hangul), like HeliBoard:
image


patrickgold (approved):
Thanks a lot, PR lgtm now!

PR #3080, opened 2025-09-04
Add the swedish version of the dvorak layout.

iceaway:
Hi,

I love the keyboard but  was missing one essential thing or switching, and that was the swedish dvorak (svorak) layout. I don't really know anything about Android development but looked at the similar files and tried to recognize by patterns what should be added.


patrickgold (approved):
Seems to be the Svorak layout (see https://en.wikipedia.org/wiki/Dvorak_keyboard_layout#Swedish).

After some review and comparison this looks really good to me, and also clean JSON formatting, thanks a lot :)

Will merge this in, and release it with 0.5.1 soon.

PR #3088, opened 2025-09-07
Add Arabic question mark to ~right popup mappings

be-at-ai:
This update adds the Arabic question mark (؟) to the ~right popup key mappings


patrickgold (approved):
Thanks for your contribution, will be released in 0.5.1!

Issue #53, opened 2020-12-01
GIF search


aircave:
Any chance gif searching would ever be added? Not sure of any other keyboards that have it other than Gboard.

D3SOX:
I think there is a Giphy API and an SDK. For privacy reasons I would implement it using the API, similar to Signal (which is also open source) and it should be no problem.

kojid0:
So florisboard has internet access permissions? 🤔

RickyM7:
> So florisboard has internet access permissions? 🤔

@kojid0 No, and I'd really like it to stay that way, but maybe this feature can be added as dictionaries are added on some keyboards, i.e. adding this feature as an extra app. I don't know if this would really solve the necessary internet access problem, but it would prevent the app size from increasing too much and leave this feature as something additional for anyone who really wants to use (I personally wouldn't use this feature).

kojid0:
> > So florisboard has internet access permissions? 🤔
> 
> @kojid0 No, and I'd really like it to stay that way, but maybe this feature can be added as dictionaries are added on some keyboards, i.e. adding this feature as an extra app. I don't know if this would really solve the necessary internet access problem, but it would prevent the app size from increasing too much and leave this feature as something additional for anyone who really wants to use (I personally wouldn't use this feature).

I wouldn't use this feature either.
Yeah, a separate "Florisboard Plus" app with GIF support (maybe free on F-Droid and Paid on Google Play to support the dev?) would be a solution. Or, the other way around, "Florisboard Slim" with only basic functions/permissions.

D3SOX:
I don't see why this would significantly increase the app size. It's basically just a view with a search box that queries the API and displays the images. I think a setting to enable/disable this is appropriate.

RickyM7:
> I wouldn't use this feature either.
Yeah, a separate "Florisboard Plus" app with GIF support (maybe free on F-Droid and Paid on Google Play to support the dev?) would be a solution. Or, the other way around, "Florisboard Slim" with only basic functions/permissions.

@kojid0 I did not mean to refer to a separate version of FlorisBoard itself, but rather just a small separate app with only the Gifs function, however this is also an option, I think.

flide:
Hi All, 
I am the author of 8vim keyboard for android. I also had a similar problem of wanting to integrate emojis, lennys and gifs to my keyboard. I initially tried to cram all into the keyboard, but it did not feel right. The app size kept increasing and I figured that the user will end up missing out on features that I am not even thinking about. 
[...]

PR #3122, opened 2025-11-06
Rehydrate/v0.1 whisper ok

yourowndog:


lm41:
Whats the difference between this one and #3123?

Per this comment https://github.com/florisboard/florisboard/issues/53#issuecomment-736606545 it is very likely that this feature won't get accepted because it uses the `Internet` permission.

PR #3123, opened 2025-11-06
Restore Stable/whisper ok

yourowndog:
Here it goes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,35 +1,56 @@
-
 name: Build Android App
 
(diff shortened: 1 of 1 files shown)


lm41:
Whats the difference between this one and #3122?

Per this comment https://github.com/florisboard/florisboard/issues/53#issuecomment-736606545 it is very likely that this feature won't get accepted because it uses the `Internet` permission.

lm41 (changes requested):
Change request explained in my previous message.

PR #3141, opened 2025-11-20
Add Georgian Standard keyboard layout

necioerrante:
## Description

Georgian Standard keyboard layout

I am not sure though if the characters on the right side of certain keys should be handled as separate pop-up layout, and how to do that

## APK testing

## Checklist


app/src/main/assets/ime/keyboard/org.florisboard.layouts/layouts/characters/georgian.json
```diff
@@ -0,0 +1,41 @@
...
+    { "$": "auto_text_key", "code":  4306 , "label": "გ" },
+    { "$": "auto_text_key", "code":  4328 , "label": "შ" },
+    { "$": "auto_text_key", "code":  4332 , "label": "წ" },
+    { "$": "auto_text_key", "code":  4310 , "label": "ზ" },
+    { "$": "auto_text_key", "code":  4334 , "label": "ხ" },
+    { "$": "auto_text_key", "code": 4330 , "label": "ც" }
+  ],
+  [
+    { "$": "auto_text_key", "code": 4324 , "label": "ფ" },
+    { "$": "auto_text_key", "code": 4331 , "label": ძ" },
```
lm41: There is a missing " in that line, which leads to a syntax error.

lm41 (changes requested):
Hello, thank you for the changes. I've identified 2 to 3 issues with this PR.

The first is directly in the code (see below).
Secondly this Layout will not work because it was not added to the `extension.json`.

The third point is, that we are currently rewriting the layout engine with the k3lp project, so it would probably be better to wait until this is implemented.

necioerrante:
> Hello, thank you for the changes. I've identified 2 to 3 issues with this PR.
> 
> The first is directly in the code (see below). Secondly this Layout will not work because it was not added to the `extension.json`.
> 
> The third point is, that we are currently rewriting the layout engine with the k3lp project, so it would probably be better to wait until this is implemented.

I believe the errors have been fixed.
Just need someone to review format, and pop ups, for correct usage.
N.B. I wanted to update original commit so it can appear as 1 single commit for clarity, but it is not possible using Web-UI, so if someone wants to refresh commits to a single "Add Georgian Standard keyboard layout" commit, that would be great

Unfortunately there is no templates nor usage guide in wiki to find necessary resources to add layouts.
Also, if this PR works why wait for k3lp as we do not know timeline for it to be done.
If everything is okay, at least keyboard will be available now and not delay usage, without any specific timeframe for completion.

It would help to have resource material available, also for k3lp

necioerrante:
Thank you for your help and your time.
I just tested the ci-build and it works perfectly.
Screenshot_20251121-202943_FlorisBoard Debug

Is there a way to show the hints for the pop up characters?
There are six single pop-up characters, the image below florisboard shows what it can look like with pop-ups hinted to the right of main character

Heliboard has option to select hint source and prioritize language:


Which shows hidden popup characters, instead of the symbols

necioerrante:
@lm41 @patrickgold 
Is there anything missing with this PR?

Please let me know if anything needs to be added or fixed, as I would like to add the configuration to my default florisboard keyboard . Instead of switching to another input method for layout access, that is.

patrickgold:
@necioerrante sry for the delay, we have a lot on our plates.

> Is there a way to show the hints for the pop up characters?

Yes, Floris Settings > Keyboard > Hinted symbols

>  Is there anything missing with this PR?

I don't think so, if you can confirm the layout behaves as intended, then we can merge this PR right before the 0.5.2 release

necioerrante:
> , if you can confirm the layout behaves as intended, then we can merge this PR right before the 0.5.2 release

Thank you for the quick reply.
I understand.
That would be great
I will test it more today and if all goes well will confirm tomorrow.
Thank you once again.

> Yes, Floris Settings > Keyboard > Hinted symbols

I cant seem to get characters to show, though.
There is no language option in florisboard, but I tried all 3 settings, yet it always shows symbol and not the desired popup character:
Hint priority,


Accent priority,


Smart priority,


All show the save hints and not the extra popup character which shows in the image in previous comment with heliboard (symbols show, but when there is popup character that-language is prioritized instead).

necioerrante:
@patrickgold 
I have a question regarding the shift key and capitalization.
Georgian does not have capitalization, but shift key aligns characters to same line:
ფძვთაპქ vs ᲤᲫᲕᲗᲐᲞᲥ

The issue is that florisboard tries to auto-capitalize the first letter.
Is there a way to disable this specifically for a language, or do you just disable the shift key as a workaround?
{ "code":  -11, "label": "shift", "type": "modifier" }

--------
I found some documentation regarding key classes, (florisboard seems to lack documentation):

> Key classes: specified with $, usually you can omit them in HeliBoard
> text_key: normal key, default
> auto_text_key: used in FlorisBoard for a key that changes text case when shift is enabled, HeliBoard does that anyway unless disabled with a labelFlag
> multi_text_key: key with an array of code points, e.g. { "$": "multi_text_key", "codePoints": [2509, 2480], "label": "্র" }

So, if I change from "auto_text_key" to "text_key" does that mean that shift will not change formatting? Nor automatic capitalize first letter?

lm41:
I'm not to familiar with the keyboard layout format myself, but I know that HeliBoard adapted FlorisBoards layout format. Thus I would think it would work. (I'm pushing a fix for that in a second.)

The hinted symbols that you are seeing are from the default "symbols" layout but your defined popup characters are prioritized.

necioerrante:
> Thus I would think it would work. (I'm pushing a fix for that in a second.)
>
Yes, I just tested the ci-build with the changes and it works,
Thank you


@patrickgold 
I believe it is all set now.
I checked over the codes for each letter to make sure they match just in case.
I do not think there are any remaining issues.
Thank you @lm41 for your help

> The hinted symbols that you are seeing are from the default "symbols" layout but your defined popup characters are prioritized

What I meant is that in heliboard when you select priority for language, the popup characters appear as hints on the top right hand corner (instead of the symbol for any particular key which has extra characters):


Symbols appear, but when there is a hidden language popup key that replaces the symbol:


Does florisboard have the ability to show they extra keys, instead of the symbols when there is a clash?

patrickgold (approved):
Thank you for the contribution and effort! From my side the changes look good, and also quickly tested the layout on a device.

Merging in, will be part of the 0.5.2 release.

PR #2750, opened 2025-01-22
Implement custom material you colors

lm41:
Closes #2749 

This PR adds the possibility to select custom material you colors based on an accent color. 

- [x] Create snapshot from https://github.com/patrickgold/jetpref/pull/14
- [x] Implement new ColorPickerPreference
- [x] Modify SnyggMaterialYouValue to work with the new colorscheme
- [x] Optional: Adapt AppTheme to also use the user defined colorscheme
- [x] Improve colorpicker logic in https://github.com/patrickgold/jetpref/pull/14
- [x] Localize strings
- [x] Create new snapshot from https://github.com/patrickgold/jetpref/pull/14 and implement material you logic
- [x] Before Merge: Use release version of https://github.com/patrickgold/jetpref/pull/14

PR #3145, opened 2025-11-23
Enable arbitrary material you colors

lm41:
## Description

This commit introduces the [materialkolor](https://github.com/jordond/MaterialKolor) library which enables color scheme generation from any seed color. This replaces our custom predefined color mappings and enables a more personal look and feel of the app and the keyboard.

As this is a "Material You" theme change I'll would like @itsmartashub as a reviewer. I tested the changes on android 16 QPR1 and android 8. 

## APK testing

## Checklist

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/theme/ThemeExtensionComponent.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/theme/ThemeExtensionComponent.kt
@@ -18,6 +18,8 @@ package dev.patrickgold.florisboard.ime.theme
 
 import dev.patrickgold.florisboard.lib.ext.ExtensionComponent
 import dev.patrickgold.florisboard.lib.ext.ExtensionComponentName
+import kotlinx.serialization.EncodeDefault
+import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import org.florisboard.lib.color.MaterialYouFlags
@@ -52,6 +54,7 @@ interface ThemeExtensionComponent : ExtensionComponent {
     fun stylesheetPath(): String = stylesheetPath.takeUnless { it.isNullOrBlank() } ?: defaultStylesheetPath(id)
 }
 
+@OptIn(ExperimentalSerializationApi::class)
 @Serializable
 data class ThemeExtensionComponentImpl(
     override val id: String,
@@ -59,6 +62,7 @@ data class ThemeExtensionComponentImpl(
     override val authors: List<String>,
     @SerialName("isNight")
     override val isNightTheme: Boolean = true,
+    @EncodeDefault
     override val materialYouFlags: MaterialYouFlags = MaterialYouFlags(),
     @SerialName("stylesheet")
     override val stylesheetPath: String? = null,
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/lib/ext/ExtensionManager.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/lib/ext/ExtensionManager.kt
@@ -58,7 +58,7 @@ import org.florisboard.lib.kotlin.throwOnFailure
 @OptIn(ExperimentalSerializationApi::class)
 val ExtensionJsonConfig = Json {
     classDiscriminator = "$"
-    encodeDefaults = true
+    encodeDefaults = false
     ignoreUnknownKeys = true
     isLenient = true
     prettyPrint = true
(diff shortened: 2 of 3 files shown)


itsmartashub:
Thanks for the trust. Im not in a position to properly test these days, but from a quick look some colors seem a little differ from v0.5.1 (or you can also compare new Floris MY themes from [this action version](https://github.com/florisboard/florisboard/actions/runs/19521275603) vs this one here).

For a predefined/fixed material palette, some colors look more accented (e.g. inversePrimary, primary), while in other present palettes the same elements look less accented than before, like they use some secondary color and not e.g. inversePrimary/primary (if need, I can list them all maybe later today). And the grey color stands out the most, almost like a blue-grey mix.

Dynamic MY function is generally bugged on my curr a16 rom, so can't test it properly but mentioned colors seem fine? :(

Only next week I could test in details and on more diff devices/roms :(

PS attached screenshots are probably huge and messy on desktop and sorry for that

<details>
        <summary>Screenshots</summary>





</details>



---

Out of topic: have you "remove" the bg from the spacebar on purpose for Floris Night Borderless (MY)?

itsmartashub:
It seems that there is a color calc bug: when starting from a custom base hue (for example green) and adjusting saturation and lightness, certain saturation/lightness ranges cause the hue to shift unexpectedly, resulting in green jumping to bluish or reddish tones (e.g. pure white ➡️ bluish, pure black ➡️ reddish). On top of that, the mentioned colors like primary/inversePrimary seem that are not being calculated consistently, so it's like their output doesn't quite align with the expected relationship to the base color.

Would try to upload a video when the poor cellular signal allows me to

lm41:
Thank you @itsmartashub for your findings. I have to say, you are right. The colors differ from the version before. This is because I previously used the `Neutral` PaletteStyle. It should be fixed in the next commit.

lm41:
> Out of topic: have you "remove" the bg from the spacebar on purpose for Floris Night Borderless (MY)?

No, that was a mistake. Thank you for finding it. Fixed in #3146

itsmartashub:
`0.5.0-debug+fe0c310 (114)`

Hm, with this latest update it seems that behaviour is the same, but now some other colors seem more neutral actually? Mostly secondary-related and surface-related colors? They look less-saturated now, like before updating the material3 dep from v1.1.0 to v1.2.0 (#2409). Dynamic themes are being affected too?

<details>
        <summary>Screenshots</summary>





</details>

patrickgold:
> Dynamic themes are being affected too?

Yes, Material You and dynamic themes in the context of Snygg themes are the same, just a more generic term.

Lars added the ability to configure the color generation on a per-stylesheet basis, in the same dialog where you can also modify the stylesheet title. These params are fixed, meaning if you find a certain set of params that yields good Material You colors, these will also apply on all other devices using this stylesheet, regardless if the device runs on Android 16 QPR 1, Android 15, or Android 8.

The only adaptive thing is the seed accent color, which is either derived from the system, or what the user chose in FlorisBoard > Themes.

itsmartashub:
Thanks for explaining it further. When I said *dynamic themes*, I meant the accent color that gets generated from the e.g. wallpaper, so my bad using imprecise term. At some PR we disxussed in the past (#2750), the theme would look correct when using the wallpaper as the source (seed), but when predefined colors from the palette were used, some theme colors looks odd until Lars fixed it. That's one of the reasons I asked whether implementing this feature also affects themes whose accent color is taken from the curr wallpaper/system

lm41:
> but when predefined colors from the palette were used, some theme colors looks odd until Lars fixed it.

As mentioned in #2750 this was only applying to the default green custom color scheme, because I just reused the default settings scheme there. This was fixed back then and not back introduced. All color schemes are now generated by only the accent color. (Either a predefined/user chosen one or the system accent color).

itsmartashub:
Just wanted to note that my previous comments (the first three) still apply. I checked this on a borrowed Realme A11 device, and the same issues are occurring [with the merged PR](https://github.com/florisboard/florisboard/actions/runs/19748781889)

Also, it seems that not only themes are affected. Through the entire app, some colors aren't the same as they did in v0.5.1 stable.

Ofc, I'm only "flagging" this in case these changes weren't intentional. If this was the intended design direction, then please disregard

<details><summary>Realme A11 Screenshots</summary>

realme-a11-dark
realme-a11-light

</details>

lm41:
Yeah that's expected. The defaults have been changed and are now calculated. For themes the new defaults can be changed in the themes meta field. For the settings it is what it is.

itsmartashub:
Good to know. In that case, glückwunsch :)

PR #3179, opened 2026-01-11
Correction and update of the French popup mapping

Norde:
## Description

The popup mapping of the French keyboard displayed many characters that do not exist in French provided an order that was not intuitive in terms of character frequency of use.
This PR proposes a correction to this and a small improvement by adding Greek characters used in science:

- Removal of characters not used in French
- Addition of Greek characters used in scientific/technical annotation
- Reordering of characters according to frequency of use

## APK testing

## Checklist

- [X] I have read and understood the [contribution guidelines](https://github.com/florisboard/florisboard/blob/main/CONTRIBUTING.md).
- [X] I have read and understood the [AI policy](https://github.com/florisboard/florisboard/blob/main/AI_POLICY.md).

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/fr.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.localization/popupMappings/fr.json
@@ -4,76 +4,111 @@
       "main": { "$": "auto_text_key", "code":  224, "label": "à" },
       "relevant": [
         { "$": "auto_text_key", "code":  228, "label": "ä" },
-        { "$": "auto_text_key", "code":  227, "label": "ã" },
-        { "$": "auto_text_key", "code":  229, "label": "å" },
-        { "$": "auto_text_key", "code":  257, "label": "ā" },
-        { "$": "auto_text_key", "code":  170, "label": "ª" },
         { "$": "auto_text_key", "code":  226, "label": "â" },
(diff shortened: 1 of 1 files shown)


patrickgold:
Hi!

Thanks for your PR. In my opinion your "Removal of characters not used in French" (which is totally valid) and "Addition of Greek characters" contradicts itself and is a case for custom layouts (which should release later this year).

patrickgold:
We can either reduce the contents of the PR to only removal on non-used characters, or close it as a custom layout.

Norde:
Hi and thanks for your feedback!

You're right, it's a hybrid implementation.
French actually have the [classic Azerty](https://en.wikipedia.org/wiki/AZERTY) and the “[new Azerty](https://github.com/dispoclavier/nouvel-azerty)” which implements some missing characters, Greek letters (used for SI units), and currency symbols.

As the link is only in french, [here are the docs](https://github.com/MicrosoftDocs/globalization/blob/main/globalization/keyboards/AZERTY_English.md) for windows implementation of the new azerty.

Mine is the new azerty, without the currency.


What I could do is:
Two versions:
- Classic azerty
- New azerty (complete)

Or only the new azerty.

Given that most android keyboards offer an hybrid implementation (adding classic azerty's missing characters and 2-3 Greek letters like µ, alpha and beta).


I would prefer only the new azerty (to avoid confusion and frustration from missing caracters) as it does not require any specific user learning with the popup system.

What do wou think?

PR #3205, opened 2026-01-31
AGP 9.0.0, plus a lot of other fixes too long to list in the title...

shadowslasher410:
## Description

- Fixes warnings and errors in generate-screenshots.sh
- Fixes a ton of little spelling errors (also standardizes the spellings to American English, because if it was the other way around, I'd have to rename everything in the project named color, and there's a LOT of things named color...)
- Updates the benchmark module to Java 11
- Updates GitHub Actions and project dependencies to latest versions, including AGP to version 9.0.0
- Fixes this really annoying resource linking error I kept getting when I tried to build the project by creating some of them in a different place (and (I know it's not a Windows thing, tried it on Kubuntu and got the same error)
- Moves the modifier parameter to be first in some of the Snygg Composables to align with the Composable guidelines (fixes all the stuff that would break because of this)
- Attempts to try and fix some of the issues with escaped special character and plurals in non-English strings.xml (there are still more errors, though)
- A lot of little changes... I'm probably forgetting something, but I tried my best to document it all in the commit messages I want to merge... let me know if you have any questions or need any clarification!

## APK testing

## Checklist

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -30,7 +30,7 @@ androidx-test-ext = "1.3.0"
 androidx-test-espresso = "3.7.0"
 androidx-test-uiautomator = "2.3.0"
 kotest = "6.1.2"
-kotlinx-kover = "0.9.4"
+kotlinx-kover = "0.9.5"
 turbine = "1.2.1"
 
 


app/src/main/kotlin/dev/patrickgold/florisboard/ime/editor/EditorInstance.kt
```diff
@@ -554,8 +554,8 @@ class EditorInstance(context: Context) : AbstractEditorInstance(context) {
          if (!(isActive || forceActive) || selection.isNotValid || selection.start <= 0 || text.isEmpty()) return false
          val textBefore = content.getTextBeforeCursor(1)
          val punctuationRule = nlpManager.getActivePunctuationRule()
-         if (!subtypeManager.activeSubtype.primaryLocale.supportsAutoSpace) return false;
-         return textBefore.isNotEmpty() &&
+         if (!subtypeManager.activeSubtype.primaryLocale.supportsAutoSpace) return false
+        return textBefore.isNotEmpty() &&
```
lm41: The indentation of the return doesn't match the rest.

app/src/main/kotlin/dev/patrickgold/florisboard/lib/compose/FlorisScreen.kt
```diff
@@ -111,7 +111,7 @@ private class FlorisScreenScopeImpl : FlorisScreenScope {
         this.bottomBar = bottomBar
     }
 
-    override fun content(content: FlorisScreenContent) {
+    override fun Content(content: FlorisScreenContent) {
```
lm41: `Content` is not a @Composable fun and should not start with a capital letter. Please also change this back in all files you've edited.

app/src/main/res/values-ca/strings.xml
lm41: Please remove all changes to translation files that are not `values/strings.xml`. See the [contribution guidelines#translations ](https://github.com/florisboard/florisboard/blob/main/CONTRIBUTING.md#translations).

fastlane/metadata/android/en-US/changelogs/12.txt
lm41: I don't think that changing the fastlane metadata after a release updates the release. @patrickgold do you know if this is something we should do?

gradle/libs.versions.toml
```diff
@@ -1,26 +1,26 @@
...
 coil = "3.3.0"
 kotlin = "2.3.0"
 kotlinx-coroutines = "1.10.2"
-kotlinx-serialization-json = "1.9.0"
+kotlinx-serialization-json = "1.10.0"
 ksp = "2.3.4"
-material-kolor = "4.0.5"
-mikepenz-aboutlibraries = "13.2.1"
+material-kolor = "4.1.0"
+mikepenz-aboutlibraries = "14.0.0-b02"
```
lm41: We only upgrade to stable versions and `mikepenz-aboutlibraries` is already on the latest stable version `13.2.1`.

lib/kotlin/src/main/kotlin/org/florisboard/lib/kotlin/Library.kt
```diff
@@ -21,7 +21,7 @@ import kotlin.contracts.contract
 
 inline fun <R> tryOrNull(block: () -> R): R? {
     contract {
-        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
+        callsInPlace(block, InvocationKind.AT_MOST_ONCE)
```
lm41: @patrickgold You know the kotlin contracts better than me. Is this `AT_MOST_ONCE` or `EXACTLY_ONCE`.

app/src/androidTest/res/drawable/ic_app_icon_debug_foreground.xml
lm41: Is there a specific reason why you added the app icons to the `androidTest` source set?

lm41 (changes requested):
Thank you very much for the changes. The first ones are always the most exiting. :tada: We appreciate every outside contributor that is interested in bringing the project forward :D

I couldn't really test your changes and check if it still compiles because of an error I got when trying to gradle sync:
```
Could not create task ':lib:snygg:generateJsonSchema'.
It is too late to add actions as the callbacks already executed.
Did you try to call onVariants from the old variant API
'applicationVariants' for instance? You can instead call onVariants
directly from the androidComponents DSL block.
```
And in a the EAP version of the IDE: 
```
The project is using an incompatible version (AGP 9.0.0) of the Android Gradle plugin. Latest supported version is AGP 9.0.0-alpha06
```

So I've just some code reading and reviewed the code in a static way (see my comments below).

Just a note for future PRs. Please split PRs like this in smaller (scope based) PRs, which will make it much easier to review. You've organized your commit messages good and now you only need to open multiple PRs according to the scope of the change :)

lm41:
@patrickgold our validate no translated strings is broken (or was broken by the changes of this PR)

patrickgold (changes requested):
Thanks the PR and congrats on your first PR in FlorisBoard :tada: 

As Lars already noted, this PR is way too large to be reviewable properly, as it does too many things at once.

I propose splitting the PR. I would prioritize only the dependency upgrades as the first split PR, because these can introduce implicit changes and need to be tested separately. Then we can slowly work on the other changes.

lib/kotlin/src/main/kotlin/org/florisboard/lib/kotlin/Library.kt
patrickgold: It is exactly once, as block() is always called, without a condition. It doesn't matter if block() may throw, block is called exactly once.

Also, this is the warning in the library module that Gradle shows since ages and I wanted to fix next time I did some changes in this module

patrickgold:
Btw, as a tip, to prevent commits like these:

image

you can amend changes to a previous commit and force-push to your feature branch.

shadowslasher410:
> @patrickgold our validate no translated strings is broken (or was broken by the changes of this PR)

I didn't change any of the text, the only thing I did was add an escape to an unescaped percent sign, and change  some of the actual tags for the plurals (From other to many for example). I read that you wanted the actual translations done on weblate, but I didn't know if mine was a translation since I didn't actually translate anything, just added an escape and edited some plurals... Also, there are a bunch of errors in android studio regarding quantity strings, although they don't seem to show up in IntelliJ Idea (I got the 1 month trial of the ultimate edition)

patrickgold:
> I didn't change any of the text, the only thing I did was add an escape to an unescaped percent sign, and change some of the actual tags for the plurals (From other to many for example).

The thing is, with the next run of the translation script it will just roll over the translated strings.xml files and void your changes.

This is why all changes for translated files, including whitespace ones, are dependent on the generator. If plural definitions are incorrect, they only need to be fixed in strings.xml, and next translation script run all translated strings.xml will reflect the change.

@lm41 what do you think of the idea to add an "this file is auto-generated please do not touch" header to the translated strings.xml files?

shadowslasher410:
Oh, also, I was wondering if I/we should update the Java version (like in the sourcecompatibility/targetcompatibility/jvm to version 17, because according to https://developer.android.com/build/releases/agp-9-0-0-release-notes the minimum jdk version is 17. That reminds me, I saw you guys updated some of the other tools, but they still aren't at the latest versions, and I was just wondering why? Does it introduce some kind of compatibility issue?

patrickgold:
Hmm, AGP 9.0 is a major update by the looks of it. That alone justifies a separate PR, as it now also implicitly adds Kotlin runtimes, and there are major DSL changes.

> That reminds me, I saw you guys updated some of the other tools, but they still aren't at the latest versions, and I was just wondering why?

Do you have examples? Normally we do batch upgrades by area, and then only select the latest _stable_ version.

patrickgold:
If you mean CMake, we need to ensure the specific CMake version is available in the IDE, in GitHub workflows, and during reproducible builds. That's why we lag behind here a bit, but tbh it doesn't really matter that much.

shadowslasher410:
> Hmm, AGP 9.0 is a major update by the looks of it. That alone justifies a separate PR, as it now also implicitly adds Kotlin runtimes, and there are major DSL changes.
> 
> > That reminds me, I saw you guys updated some of the other tools, but they still aren't at the latest versions, and I was just wondering why?
> 
> Do you have examples? Normally we do batch upgrades by area, and then only select the latest _stable_ version.

Well, according to my SDK manager, the latest stable version of the build tools is 36.1.0, also gradle can support up to JDK 25 (although that would probably cause some issues initially with compatibility)

lm41:
> @lm41 what do you think of the idea to add an "this file is auto-generated please do not touch" header to the translated strings.xml files?

Yes I think this would be a good idea. (Is there an option to specify that in crowdin, because otherwise our header would be overridden :upside_down_face:)

lm41:
AGP 9.x is currently unsupported.

> Android Studio supports AGP 9.0.0 starting with Otter 3 Feature Drop 2025.2.3. IntelliJ IDEA support for AGP 9.0.0 is expected in Q1 2026.
https://kotlinlang.org/docs/multiplatform/multiplatform-project-agp-9-migration.html

Only if both AS and intellij have AGP 9.x support in a stable release we can start to migrate. 

JDK 17 is minimum and default for AGP 8.13 (https://developer.android.com/build/releases/agp-8-13-0-release-notes)

shadowslasher410:
Okay, so do I just make a new pull request with the latest commit or do I just merge it into this one somehow?

patrickgold:
Best way to do split PRs is to just create a new feature branch branching from latest main, cherry-pick the changes you want in the split PR and then make a PR.

You can start with just the dependency upgrades if you want, without the AGP 9.0 upgrade.

lm41:
Could you please _split_ your changes across multiple feature branches (as requested several times) before adding new commits to this one? This PR is too large to review properly and simply contains too many unrelated changes. Please use my first review as a basis and extract meaningful patches for the use in the feature branches.

patrickgold:
As both Lars and I have stated multiple times now, this PR is way too big to be reviewable/mergeable, and new commits don't help with that fact. As such we will close this PR.

You are open to create new PRs that only contain isolated changes (e.g. only dependency upgrades, only typo fixes, etc.), and we will then review those changes.

shadowslasher410:
Whoops I didn't realize that it updated the pull request every time I updated my repo! I apologize! I am working on putting everything into manageable chunks so that I can more easily make smaller PRs

PR #3215, opened 2026-02-05
Version updates (+ benchmark API update, rename some const vals to uppercase)

shadowslasher410:
## Description

Mostly version updates, although I do also update the benchmark module to the latest APIs, and renamed some const vals to their expanded uppercase versions.

## APK testing

## Checklist


shadowslasher410:
I was going to ask about this earlier, but I forgot until now. I can't find anything explicitly prohibiting it, but the comments in the benchmark section I got from https://developer.android.com/topic/performance/baselineprofiles/measure-baselineprofile and I'm not sure how to cite that (or if I should?)...

patrickgold (changes requested):
@lm41 did perform an successful upgrade to AGP 9.0 in #3224  and also updated this PR. This PR requires a rebase now so it only contains the other changes before it can be merged.

patrickgold:
> I was going to ask about this earlier, but I forgot until now. I can't find anything explicitly prohibiting it, but the comments in the benchmark section I got from https://developer.android.com/topic/performance/baselineprofiles/measure-baselineprofile and I'm not sure how to cite that (or if I should?)...

Regarding this, you can just add an `See: <url>` text at the end of the docstrings, this mostly suffices in this project

PR #3219, opened 2026-02-07
Added Azerbaijani Alt Layout [QƏERTY] As An Alternative To Azerbaijani

os-guy-original:
## Description

Some people in Azerbaijan uses the QƏERTY keyboard layout.
It's also known as the "Azərbaycan dili" in the GBoard, which is the first option.

image

It appears on the Samsung Keyboard when user chooses the Azerbaijani language with the QWERTY layout.

image

The current Azerbaijani layout is mostly suitable for PC keyboards, not mobile.

This layout is more usable in Azerbaijan since the letter "ə" is used in a lot of words in Azerbaijani and is more accessible in this layout.

## Checklist

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
@@ -45,6 +45,12 @@
         "direction": "ltr"
       },
       {
+        "id": "azerbaijani_alt",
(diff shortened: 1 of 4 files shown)


khazar-os-linux:
As someone from Azerbaijan, I use this keyboard layout as my default. I'm one of the testers of this layout, and it has a very positive effect on typing speed because it's both more minimalist and more comfortable to write.

PR #3226, opened 2026-02-13
AGP 9.0.0 Fixes

shadowslasher410:
Updated source and target compatibility and JVMTarget to version 17 Fixed KSP and About Libraries configuration in app/build.gradle.kts Updated androidx-compose-bom to 2026.02.00; kotest to 6.1.3; kotlinx-kover to 0.9.7 in libs.versions.toml Bumped buildTools to 36.1.0 in tools.versions.toml Removed the workaround in android/build.gradle.kts - no longer needed Fixed the generateJsonSchema task in snygg/build.gradle.kts so that it doesn't use depreciated code.

## Description
Did a successful build with the following changes, so if you still can't get it to build, then... I don't know, maybe it's a Linux thing? (built on Windows)

## APK testing

## Checklist


lm41 (changes requested):
Thank you for your attempt on completing the AGP upgrade. 
If I gradle sync the project in IntelliJ I get the following error:

```
Suppressed sync exceptions

Could not create task ':lib:snygg:generateJsonSchema'.
org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreationException: Could not create task ':lib:snygg:generateJsonSchema'.
```

Running `:lib:snygg:generateJsonSchema` fails with:
```
A problem occurred configuring project ':lib:snygg'.
> Could not create task ':lib:snygg:generateJsonSchema'.
   > It is too late to add actions as the callbacks already executed.
     Did you try to call onVariants from the old variant API
     'applicationVariants' for instance? You can instead call onVariants
     directly from the androidComponents DSL block.
```

Updating kotest causes our tests to fail, which is why I didn't do it in my PR as this requires further investigation.
image

shadowslasher410:
I'm pretty sure this is an issue with IDEA, since it works just fine in Android Studio... Anyway, I can't find a way that doesn't use depreciated methods, so I guess I'll just close the PR.

PR #3240, opened 2026-02-21
Add Kaitag (`xdq`) layout

alkaitagi:
## Description

This PR adds Kaitag (ISO-639 `xdq`) layout along with long-press keys. The spec can be found [in this repo](https://github.com/urssivar/script).

There's an important issue with displaying the language name though. Because it's a small minority language unknown to the OS, naturally both the language list entry and the layout's whitespace key fallback to `xdq`. We need some sort of name overwriting mechanism; it will be needed for other small languages as well.

## APK testing

## Checklist

### DIFF
--- a/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
+++ b/app/src/main/assets/ime/keyboard/org.florisboard.layouts/extension.json
@@ -275,6 +275,12 @@
         "modifier": "org.florisboard.layouts:jis"
       },
       {
+        "id": "kaitag_jcuken",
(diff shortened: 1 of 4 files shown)


patrickgold:
Hi, thanks for the PR!

This PR introduces a new json layout. Json layouts will be deprecated in favor of the XML-based layouts that are introduced with [k3lp](https://codeberg.org/k3lp/k3lp). I'd suggest we wait with this Pr until k3lp is merged as a preview into FlorisBoard and then convert this layout into the new XMl format.

alkaitagi:
Thanks for the heads up, sure!

PR #3255, opened 2026-03-19
Add enthium layout

ArikRahman:
## APK testing

## Checklist

- [X ] I have read and understood the [contribution guidelines](https://github.com/florisboard/florisboard/blob/main/CONTRIBUTING.md).
- [X ] I have read and understood the [AI policy](https://github.com/florisboard/florisboard/blob/main/AI_POLICY.md).


lm41 (changes requested):
This layout is not valid JSON and won't work in the app. The layout is also not registered with the layout `extension.json`.

I would suggest that you wait and resubmit the layout once we have [k3lp](https://codeberg.org/k3lp/k3lp) [keyboard3](https://unicode.org/reports/tr35/tr35-keyboards.html) layouts implemented (should come in an upcoming alpha version) because JSON layouts will be deprecated shortly after.

PR #3283, opened 2026-05-09
feat: support icon hints for system_gui keys

caco3:
I extended florisboard to add support for icons in the hints (see keys `a`, `s`, `,`, `.` and `ENTER`):
image

## Summary

This PR enables keyboard layout authors to assign **icon hints** to character
keys by placing `system_gui` action keys in the symbols layout row that aligns
with the character row.

Previously, hints were always rendered as text labels. Special action keys
(clipboard, smartbar toggle, settings, etc.) have icons via `computeImageVector`
but no meaningful text label — so they either showed an internal label string or
nothing at all in the hint position.

## Ceavat:
I made it with AI which is clearly disallowed according to the guidelines.
My Kotlin skills are very limited. I reviewed the code and it looks ok for me, but I did not write it myself.
I understand your reasons regarding AI, but I propose on rethinking the strictness of the AI guidelines!

### `Key.kt`
Added `hintedImageVector: ImageVector?` field alongside the existing `hintedLabel`.
This stores the icon to render in the hint position when the hint key is a
system action rather than a character.

### `TextKey.kt`
In `computeLabelsAndDrawables`:
- When the resolved hint (`symbolHint` / `numberHint`) has a **negative key
  code** (all system_gui action keys use negative codes), `computeImageVector`
  is called and the result stored in `hintedImageVector` instead of setting
  `hintedLabel`.
- When no symbol/number hint exists, the existing fallback to `computedPopups.main`
  is extended: if `main` has a negative code and `computeImageVector` returns a
  non-null icon, that icon is used as the hint (e.g. a key whose only popup is a
  smartbar-toggle action will show the keyboard icon as its hint).
- `hintedImageVector` is cleared alongside `hintedLabel` on each recompute.

### `TextKeyboardLayout.kt`
Added rendering of `hintedImageVector` as a `SnyggIcon` in the hint position
(`BiasAlignment(0.85f, -0.7f)`, 11dp) when present, using the `KeyHint` style
element — consistent with how the text hint is styled.

Added missing imports: `androidx.compose.foundation.layout.size` and
`androidx.compose.ui.unit.dp`.

### `LayoutManager.kt`
`addRowHints` previously skipped any hint key whose `type != hintType`.
For symbol hints (`hintType = KeyType.CHARACTER`), `KeyType.SYSTEM_GUI` keys
are now also accepted. This allows a symbols layout to include system action
keys at specific column positions, causing those columns' character keys to
display the action's icon as a hint.

## Use Case Example

A symbols layout row can place `{ "code": -241, "type": "system_gui" }` (smartbar
toggle) at a specific column. The character key in the same column will then
show the smartbar-toggle icon as a hint and execute the action on long-press.

Similarly, clipboard paste (`-33`), settings (`-301`), or any other action with
a registered icon in `computeImageVector` can be used as a hint.

## Backwards Compatibility

- Keys with positive-code hints continue to render text labels exactly as before.
- Keys with no applicable hint are unaffected.
- No JSON schema changes required; existing layouts are unaffected.

## Testing

- Verified icon hints render correctly at the hint position for smartbar toggle,
  clipboard paste, and settings keys.
- Verified text hints on all existing character keys are unaffected.
- Verified no hint is shown when `computeImageVector` returns null for a
  negative-code hint key.
- Manually Tested on my device (Android 16)

## APK testing

## Checklist

### DIFF
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/Key.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/Key.kt
@@ -95,6 +95,11 @@ abstract class Key(open val data: AbstractKeyData) {
     open var hintedLabel: String? = null
 
     /**
+     * The computed UI hint ImageVector of this key. Used when the hint has an icon instead of a text label.
+     */
+    open var hintedImageVector: ImageVector? = null
+
+    /**
      * The computed ImageVector of this key. This value is used by the keyboard view to temporarily save the
      * ImageVector for UI rendering and should not be set manually.
      */
--- a/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/LayoutManager.kt
+++ b/app/src/main/kotlin/dev/patrickgold/florisboard/ime/keyboard/LayoutManager.kt
@@ -301,7 +301,9 @@ class LayoutManager(context: Context) {
     private fun addRowHints(main: Array<TextKey>, hint: Array<TextKey>, hintType: KeyType) {
         for ((k,key) in main.withIndex()) {
             val hintKey = hint.getOrNull(k)?.data?.compute(DefaultComputingEvaluator)
-            if (hintKey?.type != hintType) {
+            val typeMatch = hintKey?.type == hintType ||
+                (hintType == KeyType.CHARACTER && hintKey?.type == KeyType.SYSTEM_GUI)
+            if (!typeMatch) {
                 continue
             }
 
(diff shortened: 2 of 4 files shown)


patrickgold:
Hello,

thanks for the PR. As you probably could have already guessed, this PR is in violation of the AI policy you agreed to. I could just close the PR at this point and move on, but I want to give you some thoughts on why we have this policy in place and why we stand behind it.

Firstly, the changelog. Why would you use AI for this? Have you even read through the changelog? "Added missing imports: `androidx.compose.foundation.layout.size` and `androidx.compose.ui.unit.dp`." Ok?? Do you know where I see that? In the git diff of the PR. Also why is the changelog significantly longer than the code changes? There's so much filler text that is completely unnecessary. It would have completely sufficed to write a few sentences describing the feature and presenting the use case for the user. Maybe a screenshot, that's fine. But not that abomination of changelog that serves no other purpose than to waste my time. Same goes for your commit message btw.

Second, if I want to talk to an AI, I can just type `claude.ai` in my URL bar and do that myself. I really hate it when humans don't even bother to use their own words anymore, but delegate the most simplest of tasks to AI. Open-source is more than simply pumping out as much code as possible. It is about building a project and collaborating together with other humans.

Sorry for the rant and harsh words, I just want to give an insight as to why we stand behind our AI policy and why we are sick of the repeated violations that occurred already.

PR #3286, opened 2026-05-24
fix(telex): uppercase shift-state leak in composer + improved telex rules

ngocthanhgl:
## Fix

`WithRules.getActions()` in `Composer.kt` used `.uppercase()` on the entire output string when any uppercase input was detected. For example typing `Qwas` produced `QUÁ` instead of `Quá`.

The fix checks the uppercase scope:
- Whole match uppercase → all-caps output
- Only first key uppercase → capitalize first char only
- Neither → lowercase output

## Rules

Replaced the built-in Telex rules with telex2 rules modified by Me (original from https://github.com/daanturo/viet-im-rules) for improved Vietnamese input support.

Fixes the uppercase composition bug discussed in PR #2267.


ngocthanhgl:
## English

This PR fixes a bug where typing uppercase accented Vietnamese in Telex mode produces wrong output. For example, typing `Quas` gave `QUÁ` instead of `Quá`. The issue was in `WithRules.getActions()` which applied `.uppercase()` to the entire output string regardless of where the shift was pressed.

The fix correctly handles three cases:
- Whole input uppercase → all-caps output (e.g. `QUAS` → `QUÁ`)
- Only first letter uppercase → capitalize first character only (e.g. `Quas` → `Quá`)
- All lowercase → normal lowercase output (`quas` → `quá`)

